PR #1, opened 2020-07-20
Update github actions

johnschoeman:
Why:
We would like for the ci to run on every PR and not every commit.


mattThousand:
:shipit: 💯

PR #2, opened 2020-07-20
iOS Guard Against Server Error When Submitting Empty Key Payload

mattThousand:
### Why
Users downloading the app to test for the first time may submit keys before their device has generated any. This will cause an error when submitting a positive diagnosis to the server, and present confusing information to the tester. We get around this by checking to see if there are any local keys present before submitting to the server, and if there are no keys present, we show an actionable message to the user. 

### This Commit
This commit checks if the keys fetched from the local device are an empty collection, and if so presents the user with an actionable localized message

Simulator Screen Shot - iPhone 11 - 2020-07-19 at 21 38 27


johnschoeman:
@mattThousand , looks like the ci got in a funky place, you might need to push a dummy commit to get it to restart.

PR #3, opened 2020-07-20
Consolidate helpers folder into utils

johnschoeman:
Why:
We currently have two somewhat similar folders of `helpers` and `utils`,
we would like to have a single place for utility functions.


src/locales/languages.ts
```diff
@@ -5,7 +5,7 @@ import i18next from "i18next"
 import { initReactI18next, useTranslation } from "react-i18next"
 import { NativeModules, Platform } from "react-native"
 
-import { GetStoreData, SetStoreData } from "../helpers/General"
+import { GetStoreData, SetStoreData } from "../utils/General"
```
aledustet, line 8: Should we be importing things from the index file or is not that important?

johnschoeman: I think it is. The goal if this pr was to get everything in the same place.  I think a great next pr would be to clean up the utils in general. I think we have a few that arent being used for example and the name `General` could be more specific for example.

PR #5, opened 2020-07-20
Convert self assessment end screen snapshot tests

aledustet:
Why:
----
We want to have tests that reflect better the behavior and feel of the
system, snapshot tests offers little information in that regard and are
hard to make sure we are updating the right thing.

Reviewers:
----
This is an ongoing effort that we will try to split in multiple commits.

This PR:
----
- Switch AssessmentComplete spec from snapshot to text verification
- Update Caregiver spec to match text and not snapshot
- Convert Distancing spec from snapshot to text matching
- Convert Emergency screen test from snappshot to text matching
- Convert Isolate screen tests from snapshot to text matching
- Convert Shre screen snapshot test to text matching


__mocks__/react-native-config.js
```diff
@@ -1,4 +1,5 @@
 export default {
-  FLAG_FOO_BAR: 'true',
-  TRACING_STRATEGY: 'bt',
-};
+  FLAG_FOO_BAR: "true",
```
johnschoeman, line 2: can we remove this one?

aledustet: My bad rebase 😅

PR #7, opened 2020-07-20
iOS Pull Region Codes from .env

mattThousand:
### Why
The GAEN server requires individual region codes for each HA.

### This Commit
This commit pulls region codes from the .env file.


ios/BT/API/Model/Region.swift
```diff
@@ -1,3 +1,6 @@
 enum Region: String {
-  case US
+  case US,
```
johnschoeman, line 2: Should we put more/allish of the regions in? https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes

mattThousand: good point, the `Region` enum wasn't doing much for us. Replaced it with a `String` typealias so that any string passed in from the .env file will be allowed.

PR #8, opened 2020-07-20
Hide Problematic Debug Menu Items 

mattThousand:
### Why
We'd like to prevent testers from manipulating data and producing confusing behavior from the app

### This Commit
This commit hides debug menu items that manipulate local data or make calls to the GAEN server. It leaves in place only the most necessary buttons that are useful in keeping track of local vs server state.

Screenshot 2020-07-20 at 12 06 55 PM

### DIFF
--- a/src/More/ENDebugMenu.tsx
+++ b/src/More/ENDebugMenu.tsx
@@ -103,71 +103,67 @@ const ENDebugMenu = ({ navigation }: ENDebugMenuProps): JSX.Element => {
       ) : (
         <ScrollView>
           <View style={styles.section}>
-            {/* <DebugMenuListItem
-              label="Reset Exposures"
-              style={styles.lastListItem}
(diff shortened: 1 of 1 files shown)


src/More/ENDebugMenu.tsx
```diff
@@ -119,55 +103,55 @@ const ENDebugMenu = ({ navigation }: ENDebugMenuProps): JSX.Element => {
       ) : (
         <ScrollView>
           <View style={styles.section}>
-            <DebugMenuListItem
+            {/* <DebugMenuListItem
```
aledustet, line 106: Are these menu items going to be useful in any scenario? If not how do we feel about deleting them and probably the native module functions they invoke?

johnschoeman: plus one, we should remove code if its not being used.  are these something that we want to keep internally for devs? or remove altogether?

PR #13, opened 2020-07-20
Fetch info.plist in fetch_ha_env script

johnschoeman:
Why:
We would like to fetch the ha specific google service info.plist in the
fetch ha script to automate the build process for various HAs.

### DIFF
--- a/bin/fetch_ha_env_github.sh
+++ b/bin/fetch_ha_env_github.sh
@@ -9,11 +9,21 @@ ACCESS_TOKEN = ARGV[0]
 
 def fetch_env
   source = ".env.bt"
+  google_service_source = "GoogleService-Info.plist"
+
   env_url =
   "https://#{ACCESS_TOKEN}@raw.githubusercontent.com/Path-Check/pathcheck-mobile-resources/master/environment/github/.env.bt"
 
-  open(source, 'w') do |f|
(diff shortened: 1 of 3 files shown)


bin/fetch_ha_env_github.sh
```diff
@@ -9,11 +9,21 @@ ACCESS_TOKEN = ARGV[0]
 
 def fetch_env
```
aledustet, line 10: It might an overkill here but have we considered having one script and call it with an argument for release and development? Seems like we will likely be updating both frequently in the future.

johnschoeman: yeah that would be great, i gave that a pass and struggled a bit to get the github action ci to play nicely with the script args.  They are different enough.

locally we want to pass in the HA_LABEL and fetch the token from the env:
```
bin/fetch_ha_env.sh pr
```

on ci we want to pass in the access token and default to the ha label of github

```
bin/fetch_ha_env.sh $ACCESS_TOKEN
```

i suppose we could pass in both as optional args and find a way to switch, but that struct me as more complicated than simply having two files.  open to any suggestions!

johnschoeman: also, dotenv is no on the box image by default, so we would also need to standup all of the plumbing to get that working on ci, which is odd as we dont actually need dotenv on ci, just locally.

PR #14, opened 2020-07-20
Display yellow badge when user has a new exposure

amandabeiner:
#### Description:
As a user, I want to be alerted in the app when I have a new exposure so
that I know where to go to see the details of the exposure.

This commit:
Adds the ability to set a user's exposure status to the. When a new
exposure is detected, the exposure status is set to `true`, and when a
user visits the calendar tab, the exposure status will be set to
`false`. In both cases, the value will be persisted to AsyncStorage, so
that if they force quit the app, the badge will still be visible when
you re-launch.

#### Linked issues:

Fixes [I want to see a bad on the calendar icon](https://trello.com/c/cfe33Gbb/194-as-a-user-i-want-to-see-a-badge-on-the-calendar-icon-when-i-have-an-en-so-that-i-notice-it-as-soon-as-possible)

#### Screenshots:

badge

#### How to test:

- Simulate an exposure
**Expected:** a yellow badge appears on the calendar icon
- Quit the app and reopen it
**Expected:** the yellow badge notification persists
- Navigate to the exposure history screen
**Expected:** the yellow badge disappears


src/ExposureHistoryContext.tsx
```diff
@@ -17,23 +17,23 @@ import {
   ExposureCalendarOptions,
   ExposureInfo,
 } from "./exposureHistory"
+import { GetStoreData, SetStoreData } from "./utils/General"
```
johnschoeman, line 20: this looks a bit out of date, we updated the utils folder a bit.

src/ExposureHistoryContext.tsx
```diff
@@ -87,6 +87,18 @@ const ExposureHistoryProvider: FunctionComponent<ExposureHistoryProps> = ({
     })
   }, [])
 
+  const getExposureStatus = useCallback(() => {
+    GetStoreData(NEW_EXPOSURES).then((exposureStatus) => {
```
johnschoeman, line 91: is there a way we can do this without using async storage?

amandabeiner: @johnschoeman This is based on a conversation in the old repo: https://github.com/Path-Check/covid-safe-paths/pull/1286#issuecomment-660332673

The desired behavior is that the new exposure badge should be persisted if a user force quits the app without clearing the badge notification. Is there somewhere else you would prefer to persist this state?

src/ExposureHistoryContext.tsx
amandabeiner: Just rebased, it is up to date now.

src/ExposureHistoryContext.tsx
johnschoeman: nope, makes sense to me was just wondering.

johnschoeman: although I was chatting with @tparesi about this, are we clear on what the 'last updated at' should be? is it the last time we received an actual exposure? or is it the last time we checked for an exposure?

I think if its the last time we checked we need the async storage as thats where the data is saved for the batch file processing.

If its the last time we received a possible exposure notification, could we update the lastDetectionDate when we consume the emitted exposure event from the native layer?  That would allow us to not need do an extra ping to the native side.

ios/Gemfile
```diff
@@ -2,7 +2,7 @@ ruby '~> 2.6.3'
 source "https://rubygems.org"
 
 gem "bundler", "=2.1.4"
-gem "cocoapods", "1.9.1"
+gem "cocoapods", "1.9.3"
```
johnschoeman, line 5: 👍

src/ExposureHistoryContext.tsx
```diff
@@ -89,6 +88,16 @@ const ExposureHistoryProvider: FunctionComponent<ExposureHistoryProps> = ({
     })
   }, [])
 
+  const getExposureStatus = useCallback(async () => {
+    const exposureStatus = await StorageUtils.getNewExposuresData()
```
johnschoeman, line 92:
```suggestion
    const hasNewExposure = await StorageUtils.getHasNewExposure()
```
A little confused on the naming. It looks, like it wants to be a boolean, but is named as a string or object?

src/ExposureHistoryContext.tsx
```diff
@@ -89,6 +88,16 @@ const ExposureHistoryProvider: FunctionComponent<ExposureHistoryProps> = ({
...
+    setUserHasNewExposure(exposureStatus)
+  }, [])
+
+  const setExposureStatus = useCallback(async (exposureStatus: boolean) => {
```
johnschoeman, line 96:
```suggestion
  const setHasExposure = useCallback(async (hasExposure: boolean) => {
```

johnschoeman: why are we setting the exposure status from the JS layer? isnt this data that is determined by the native layer?

src/utils/storage.ts
```diff
@@ -35,3 +35,28 @@ export async function setStoreData(
     console.log(error.message)
   }
 }
+
+const NEW_EXPOSURES = "NEW_EXPOSURES"
+export async function getNewExposuresData(): Promise<boolean> {
```
johnschoeman, line 40: I may be a bit confused here, but why do we need this new entry in the async store?  Dont we already have the exposure info data? that is we either have a new exposure or we dont? and saving this boolean separately introduces multiple sources of truth?

johnschoeman (approved):
Code looks good! A little confused on some of the functionality, might be work doing some context sharing IRL.

mattThousand (approved):
Awesome, thank you!

ios/Gemfile
mattThousand: 👍

src/ExposureHistoryContext.tsx
amandabeiner: @johnschoeman This might be worth some IRL context sharing as you mentioned, but if I understand you correctly, it seems like we are slightly conflating the concepts of "last updated at" and "new exposure badge". IIRC, "last updated at" represents the last time that PathCheck communicated with the GAEN server to see if there was a new exposure. Whereas the "new exposure badge" represents that the PathCheck check for exposures process completed, **and** that a new exposure was found, **and** that that happened since the last time that a user visited the ExposureHistory screen.

This leads me to believe that these two concepts are separate, and that it needs to have access to the JavaScript layer, because it is dependent on the user interacting with the ExposureHistory tab.

src/ExposureHistoryContext.tsx
amandabeiner: Slightly related to the above, but this is because this isn't determining whether there was an exposure, but rather whether the user has seen that exposure. I think your suggestion about changing the naming of the variable could make this more clear.

src/utils/storage.ts
amandabeiner: This is the first part of the work we introduced in #15 in order to have more type safety around what data we are saving to the device. Since the time that the exposure was detected and the time that the user saw that they have a new exposure are different, we have to track the status of whether or not a user has seen their most recent exposure separately. Matt had suggested that this should persist if a user force quits without acknowledging the notification, which is why we are keeping it in the async store.

octohub:
@amandabeiner There was a change made to the build checks so you'll have to merge the latest from `develop` and push to this branch so the `build_debug_android` passes.  Sorry for the hassle!

johnschoeman (commented):
@amandabeiner , is there anything in this one we want to merge in? or should we close the pr?

PR #16, opened 2020-07-20
Fix navigation error when selecting a language

amandabeiner:
Why:
As a user just downloading the app, I want to be able to select my
preferred language.

This commit:
Fixes a bug where the LanguageSelection screen was removed from the
stack navigator in the OnboardingStack.

#### Screenshots:

bienvenido

#### How to test:
- Download a fresh version of the app
- You should be able to switch languages from the welcome screen.

### DIFF
--- a/src/navigation/MoreStack.tsx
+++ b/src/navigation/MoreStack.tsx
@@ -13,28 +13,21 @@ import ENLocalDiagnosisKeyScreen from "../More/ENLocalDiagnosisKeyScreen"
 import ExposureListDebugScreen from "../More/ExposureListDebugScreen"
 import LanguageSelection from "../More/LanguageSelection"
 
-import { Screens, Stacks } from "./index"
+import { MoreStackScreens, MoreStackScreen } from "./index"
 
-const Stack = createStackNavigator()
+type MoreStackScreenParams = {
+  [key in MoreStackScreen]: undefined
+}
+const Stack = createStackNavigator<MoreStackScreenParams>()
 
 const SCREEN_OPTIONS = {
   headerShown: false,
 }
 
-type MoreStackRouteName =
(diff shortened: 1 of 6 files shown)


amandabeiner:
@johnschoeman Getting reliable tests for navigation components will be difficult with our mocking patterns, but we implemented some changes to the Screen and Stack types to be more explicit about which screens need to exist to avoid regressions in the future!

johnschoeman:
> @johnschoeman Getting reliable tests for navigation components will be difficult with our mocking patterns, but we implemented some changes to the Screen and Stack types to be more explicit about which screens need to exist to avoid regressions in the future!

makes sense to me! Yeah anything with react navigation or screen transitions in general would require a detox spec, imo. The mocking route seems like a lot of work for not a lot of benefit.

src/navigation/MoreStack.tsx
```diff
@@ -13,36 +13,29 @@ import ENLocalDiagnosisKeyScreen from "../More/ENLocalDiagnosisKeyScreen"
...
-  | "Settings"
-  | "About"
-  | "Licenses"
-  | "ENDebugMenu"
-  | "LanguageSelection"
-  | "AffectedUserFlow"
-  | "ExposureListDebugScreen"
-  | "ENLocalDiagnosisKey"
-
 interface MoreStackRouteState {
```
devinjameson, line 27: @amandabeiner Some of this code is being removed in https://github.com/Path-Check/gaen-mobile/pull/20 so it may make sense to rebase onto that once it's merged? Not sure what is best here. Lmk what you think!

devinjameson: Actually scratch that, I'll rebase onto this!

PR #20, opened 2020-07-20
Refactor nav bar on settings screens

devinjameson:
Why: prior to this commit, the screens within the "settings" stack were
using a navigation bar wrapper component instead of the header built in
to React Navigation.

Note: we encountered some issues with getting the screen transitions to
perform correctly when transitioning to or from a screen with a hidden
tab bar. We had to transition back to including modal views within the
top-level navigator. Turns out this is the only way to make a screen up
over the tab bar without buggy transition animations.

See https://github.com/react-navigation/react-navigation/issues/6841 for
more information on this issue.

Co-Authored-By: Jeff Stolz <jstolz123@gmail.com>

## Before
before

## After
after


.tool-versions
```diff
@@ -1 +1 @@
-ruby 2.7.1
```
johnschoeman, line 1: looks like you need to update your local ruby? `asdf local ruby 2.7.1` ?

ios/Podfile.lock
```diff
@@ -506,4 +506,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: deac9465ea5c76151798974e97010c4bc4c95a37
 
-COCOAPODS: 1.9.3
```
johnschoeman, line 509: same with cocoapods: `brew upgrade cocoapods`

src/AffectedUserFlow/index.tsx
```diff
@@ -1,46 +1,17 @@
...
 import Complete from "./Complete"
 import PublishConsent from "./PublishConsent/PublishConsentScreen"
 
-import { Icons } from "../assets"
-import { Colors, Spacing } from "../styles"
 import { Screens } from "../navigation"
 
 const Stack = createStackNavigator()
 
-const ExportStack = (): JSX.Element => {
```
johnschoeman, line 19: such a nice diff 💯

src/More/About.tsx
```diff
@@ -11,66 +10,55 @@ import {
...
 
-  const backToMain = () => {
-    navigation.goBack()
-  }
-
   return (
-    <NavigationBarWrapper
-      title={t("screen_titles.about")}
-      onBackPress={backToMain}
+    <ScrollView
```
johnschoeman, line 21: 💯

src/More/Menu.tsx
```diff
@@ -0,0 +1,194 @@
...
+        style={[styles.icon, { marginRight: Spacing.small }]}
+      />
+      <RTLEnabledText style={{ ...Typography.mainContent }}>
+        {label}
+      </RTLEnabledText>
+    </View>
+  </TouchableHighlight>
+)
+
+const SettingsScreen = ({ navigation }: SettingsScreenProps): JSX.Element => {
```
johnschoeman, line 60:
```suggestion
const MenuScreen: FunctionComponent<MenuScreenProps> = ({ navigation }) => {
```

src/More/index.tsx
```diff
@@ -1,204 +1,90 @@
...
+  | "ENDebugMenu"
+  | "LanguageSelection"
+  | "AffectedUserFlow"
+  | "ExposureListDebugScreen"
+  | "ENLocalDiagnosisKey"
 
-  const navigateTo = (screen: string) => {
-    return () => navigation.navigate(screen)
-  }
+interface MoreStackRouteState {
```
johnschoeman, line 44: for context, this typing would ideally be removed and replaced with something directly from the library, but I wasnt able to get it to line up from what is provided from their declarations so we needed to recreate it.

src/More/Menu.tsx
devinjameson: Good catch!

src/More/index.tsx
devinjameson: Got it makes sense!

devinjameson:
> I dig the refactor! what are your reasons for moving the react-navigation stack from the navigation folder to the view layer folder? It looks good to me, but curious to your reasoning.

@johnschoeman Good question. I kept expecting the stack navigator for the more menu to be at `More/index.tsx` and was confused when it was actually the menu screen. Makes more sense to me for the stack to live at `index.tsx.`

.tool-versions
devinjameson: Thanks!

ios/Podfile.lock
devinjameson: Thanks!

src/More/index.tsx
devinjameson: We actually don't need this any more right? Given that we're not doing the tab bar visibility thing.

src/navigation/MainNavigator.tsx
```diff
@@ -23,6 +27,11 @@ const MainNavigator: FunctionComponent = () => {
         ) : (
           <Stack.Screen name={Stacks.Onboarding} component={OnboardingStack} />
         )}
+        <Stack.Screen
```
devinjameson, line 30: @johnschoeman I hate that we have to do this for any "modal" screens, but it's the only way to get it to work properly. Is there a better way to organize this? I don't like how this is going to scale.

src/More/index.tsx
devinjameson: > > I dig the refactor! what are your reasons for moving the react-navigation stack from the navigation folder to the view layer folder? It looks good to me, but curious to your reasoning.
> 
> @johnschoeman Good question. I kept expecting the stack navigator for the more screens to be at `More/index.tsx` and was confused when it was actually the menu screen. Makes more sense to me for the stack to live at `index.tsx` and the menu screen to live at `More/Menu.tsx`.

@johnschoeman Do you think this should live in `navigation/MoreStack.tsx` instead?

johnschoeman: I could see it going both ways tbh.  on the one hand being consistent and having all the navigators in one spot is nice. On the other hand grouping things by feature rather than implementation detail usually leads to a code base more aligned with business goals.  So long as we are consistent, I dont think it ultimately matters too much.  That said, I do see most projects keep all their navigators colocated so that is somewhat more of the idiomatic approach.

totally up to you on what you think is best, I think you have the most context on the navigation solution at this point.

src/navigation/MainNavigator.tsx
johnschoeman: Yup I think we'll want a `Modal` stack of some sort: https://reactnavigation.org/docs/modal/

which not 100% sure, but likely wants to live in the 'MainNavigator' bit. so something like:

```
<NavigationContainer>
{ hasCompletedOnboarding ?
  <Navigator.Stack mode={Modal}>
      <MainTabNavigator />
      <Modal 1 />
      <Modal 2/>
  </Navigator.Stack>
  : <OnboardingScreen>
}
</NavigationContainer>
```

tparesi:
@devinjameson On the About and Legal screens, the color is off under the text with these changes.

Besides that, LGTM.

devinjameson:
@johnschoeman Could you help with the failing unit-test-ios tomorrow?

PR #21, opened 2020-07-20
Remove unused notification service

johnschoeman:
Why:
This class and its dependencies are no longer being used.


jest/setupFile.js
```diff
@@ -1,41 +1,37 @@
...
-jest.mock('react-native-linear-gradient');
+jest.mock("react-native-pulse")
+jest.mock("react-native-linear-gradient")
 
 // Silence YellowBox useNativeDriver warning
-jest.mock('react-native/Libraries/Animated/src/NativeAnimatedHelper');
+jest.mock("react-native/Libraries/Animated/src/NativeAnimatedHelper")
 
-jest.mock('@react-native-community/async-storage', () => mockAsyncStorage);
-jest.mock(
```
johnschoeman, line 29: this is the only mock that was removed.

aledustet:
Is it missing a pod install?

johnschoeman:
> Is it missing a pod install?

turns out that the native layer is using `#import <RNCPushNotificationIOS.h>` which is the native module for the `react-native-push-notification` library. so while we dont need it in the js layer, it is a used dependency in the native layer.  @mattThousand, @aledustet , yall have thoughts on best practice here? would it make more sense to not use the RN dependency and replace it with a native only solution?

aledustet:
> > Is it missing a pod install?
> 
> turns out that the native layer is using `#import <RNCPushNotificationIOS.h>` which is the native module for the `react-native-push-notification` library. so while we dont need it in the js layer, it is a used dependency in the native layer. @mattThousand, @aledustet , yall have thoughts on best practice here? would it make more sense to not use the RN dependency and replace it with a native only solution?

I believe it is ultimately a matter of what type of codebase we are running, even if we don't have a usage for this library on the javascript layer it feels like it is a linked dependency from both the of the codebase sides. I am not opposed on removing them in favor of just an import on the iOS side since we have dependencies that are exclusive of each one but we would have to include something like that on the android side also, thus splitting in 3 the complexity unless we bring the `react-native-notifications` for the android side as a node module also and we manage those dependencies from the react-native project.

PR #23, opened 2020-07-21
Remove emotion dependency

johnschoeman:
Why:
We where only using the emotion dependency in one component and we would
like to remove themed components in favor of the style system provided
by react-native directly.


src/components/IconButton.tsx
```diff
@@ -1,6 +1,7 @@
-import styled from "@emotion/native"
 import React from "react"
+import { StyleSheet, TouchableOpacity } from "react-native"
```
aledustet, line 2: I guess this answers my question https://github.com/Path-Check/gaen-mobile/pull/26#discussion_r458072265 😅  I'll update the tests once you merge this one

PR #26, opened 2020-07-21
Convert last batch of snapshot tests

aledustet:
Why:
----
Snapshot tests offer just one layer of testing, that the component is semantically correct but it is hard to understand if the change actually granted an update of the snapshot, making it not very valuable in confidence terms.

This Commit:
----
- Convert IconButton specs from snapshots
- Remove unused Switch component
- Remove home index snapshot spec in favor of home component spec
  * This one was removed because this component is acting as a proxy for the presentational one, that is tested.

How to test:
----
- The test suite is passing and the app builds


src/components/IconButton.tsx
```diff
@@ -21,6 +21,7 @@ export const IconButton = ({
   return (
     <Container accessibilityLabel={accessibilityLabel} {...otherProps}>
       <SvgXml
+        testID="icon-button-svg"
```
aledustet, line 24: this felt weird but eventually made some sense since this is a presentational component and it's logic is based on proxying the props down, happy to remove this.

src/Home/index.spec.tsx
```diff
@@ -1,16 +0,0 @@
...
-import { render } from "@testing-library/react-native"
-import { useFocusEffect } from "@react-navigation/native"
-
-import HomeScreen from "./index"
-
-jest.mock("@react-navigation/native")
-;(useFocusEffect as jest.Mock).mockReturnValue(jest.fn())
-
-describe("HomeScreen", () => {
-  it("renders", () => {
```
aledustet, line 11: This is a matter of how we organize these components, and how we test them afterward. 
- Should we be testing this one with all the setup? Context and all but just the happy path and then the presentational one with all the bells and whistles and all the branches of execution? 
- Should we not be testing this one at all?

src/components/IconButton.tsx
aledustet: Also, this file is using some form of styled component that I'm not sure we need, thoughts?

johnschoeman: yup, removed in another commit!  Also worth noting that this component is only used in one spot. What do you think about inlining it?

src/components/Switch.js
```diff
@@ -1,24 +0,0 @@
-import React from "react"
```
johnschoeman, line 1: 💯 , i assume this hasnt been used anywhere for a while now?

aledustet: It sure looks like it was part of an old set of UI requirements

src/Home/index.spec.tsx
johnschoeman: I think this one makes sense to be able to test the branches as they are quite important.

We will want some better coverage around, showing the possible states with ctas of 
```
// ios
1. en are authorized and app is enabled
2. en are authorized and app is not enabled
3. en is not authorized

// android
1. en are enabled
2. en are not enabled
```

johnschoeman: so i would tend to taking a pass at some TDD on this and see where the component wants to go as far as split.

johnschoeman: theres also tests we want to have around the listening for updates to the auth status via events

aledustet: I'm not sure I'm following here. This means that we should not test the presentational component in isolation? Like we do on the `Home.spec.tsx`? So should I move that test one level up and just test this entry file with all the branches and callback logic and then remove/rename the component test?

aledustet: I believe my concern is more about how we are organizing some of these components that receive some data from the context on an index file and then render and make decisions on the data on the more visually packed component. Is this in an effort to make testing easier by not having to have a context provider to test those or is it to separate logic from view?

src/components/IconButton.tsx
aledustet: This is only used on the `Eula` component, I think that adding more into that file might have to be reverted later, happy to do it if we feel is better

src/Home/index.spec.tsx
johnschoeman: > Is this in an effort to make testing easier by not having to have a context provider to test those or is it to separate logic from view?

Imo, these are both positives gained by having them separate. In general i like things that tend towards functional core / imperative shell.

As far as testing the index/thing that connects to the provider/does IO: its hard to test IO stuff in general so keeping IO co located and at the boundary of the system helps. We may need a fake or some pretty heavy mocking to make a useful spec suite.

That all said, i definitely defer to you/whoever is writing the specs to what makes sense in this case in particular.  If it turns out better to consolidate the components we should do that.

src/components/IconButton.tsx
aledustet: @johnschoeman I moved the components into the Eula Modal and transformed it to Typescript, you mentioned something about letting the GPS team know, should I point them back to this commit?

johnschoeman: Yes, or if you are feeling particularly generous you could open a similar commit to the GPS repo.

PR #31, opened 2020-07-21
Add loading indicator to EULA webview

devinjameson:
after

### DIFF
--- a/src/Onboarding/EulaModal.tsx
+++ b/src/Onboarding/EulaModal.tsx
@@ -8,6 +8,7 @@ import {
   SafeAreaView,
   StatusBar,
   Image,
+  ActivityIndicator,
 } from "react-native"
 import { useTranslation } from "react-i18next"
 import loadLocalResource from "react-native-local-resource"
@@ -89,16 +90,17 @@ const EULA_FILES: Record<AvailableLocale, string> = {
 
 type EulaModalProps = {
   selectedLocale: string
-  continueFunction: () => void
+  onPressModalContinue: () => void
 }
 
 const EulaModal: FunctionComponent<EulaModalProps> = ({
   selectedLocale,
-  continueFunction,
+  onPressModalContinue,
 }) => {
   const [modalVisible, setModalVisibility] = useState(false)
   const [boxChecked, toggleCheckbox] = useState(false)
   const [html, setHtml] = useState<string | undefined>(undefined)
+  const [isLoading, setIsLoading] = useState(true)
   const { t } = useTranslation()
 
   // Pull the EULA in the correct language, with en as fallback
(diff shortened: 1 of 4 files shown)


ios/Podfile.lock
```diff
@@ -247,7 +247,7 @@ PODS:
     - React
   - RNCMaskedView (0.1.5):
     - React
-  - RNCPushNotificationIOS (1.4.0):
+  - RNCPushNotificationIOS (1.2.0):
```
devinjameson, line 250: Why is this happening?

.tool-versions
```diff
@@ -1 +1 @@
-ruby 2.7.1
```
johnschoeman, line 1: we probably wanna stay on 2.7.1 here.

ios/Podfile.lock
```diff
@@ -488,4 +488,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: deac9465ea5c76151798974e97010c4bc4c95a37
 
-COCOAPODS: 1.9.3
```
johnschoeman, line 491: same with this guy.

src/Onboarding/EulaModal.tsx
```diff
@@ -128,6 +130,16 @@ const EulaModal: FunctionComponent<EulaModalProps> = ({
 
   const canContinue = boxChecked
 
+  const handleOnPressContinue = () => {
+    setModalVisibility(false)
+    continueFunction()
```
johnschoeman, line 135: I know this wanst introduced by you here. but `continueFunction` is a pretty unhelpful name. what do you think about renaming it here to something like: `onPressModalContinue` or similiar?

.tool-versions
devinjameson: Can we pair on this later? I keep running into this issue.

johnschoeman: sure! ill have some time around 4:30 today if that works for you

ios/Podfile.lock
johnschoeman: did you pod install?

aledustet: @devinjameson  I think this might be because the global cocoa pods version on your system could be lower than the one bundled? Is the output from `gem which cocoapods` showing the `1.9.1` version or the `1.9.3` ?

ios/Podfile.lock
```diff
@@ -12,6 +12,37 @@ PODS:
     - React-Core (= 0.61.5)
     - React-jsi (= 0.61.5)
     - ReactCommon/turbomodule/core (= 0.61.5)
+  - Firebase/CoreOnly (6.25.0):
```
devinjameson, line 15: @aledustet Is this ok to merge? Can't figure out how to get rid of all this stuff.

PR #32, opened 2020-07-21
Add script to update display names before release

aledustet:
Why:
----
Each HA will have a unique display name they want shown on the application that gets released to the final user. We need a way to update this properties before cutting a production build

This Commit:
----
- Add a script that uses the libraries `PlistBuddy` and `xmllint` to edit the `Info.plist` for the iOS application and the `strings.xml` for the android application

How to test:
----
- After executing the `fetch_ha_env` script run `./set_display_name.sh` and verify that the names have been updated for both applications.

### DIFF
--- a/bin/set_display_names.sh
+++ b/bin/set_display_names.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env ruby
+#
+# Update the applications display name based on the release env file
+#
+# Usage
+#
+#   bin/set_display_names <?env>
+#
+# Example
+#
+#   bin/fetch_env
+#
+# Requirements
+#
+# 1. An .env.bt.release file in the root folder or an existing environment file
+# in the route of the argument
+
+require 'dotenv'
+require 'open3'
+
+# Constants
+ENV_FILE = ARGV[0] || ".env.bt.release"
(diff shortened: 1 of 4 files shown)


bin/set_display_names.sh
```diff
@@ -0,0 +1,95 @@
...
+  else
+    puts "Could not read ios current name from file #{PLIST_PATH}, error: #{error}"
+    exit 1
+  end
+end
+
+def update_ios_display_name(new_name)
+  puts "Updating ios display name from #{get_current_ios_name} to #{new_name}"
+  output, error, status = Open3.capture3(
+    "/usr/libexec/PlistBuddy -c \"Set :CFBundleDisplayName #{new_name}\" \"#{PLIST_PATH}\""
```
aledustet, line 43: Opted for using the local libraries for `macos` happy to add ruby specific libraries, for that we should add a `Gemfile` in the root of the project. This felt lightweight  and the right amount of effort for the task.

johnschoeman: This makes sense to me.  I dont think we should avoid adding gemfile persay, but keeping the build process as lean as possible is a good goal, imo.

PR #34, opened 2020-07-22
Fetch EULA from the web

zhaccc:
#### Description:

This PR fetches EULA from a mock web server. Locales are also supported.
This <ins>**SHOULD NOT**</ins> be merged before we have a real webserver to fetch EULA from.

### DIFF
--- a/src/Onboarding/EulaModal.tsx
+++ b/src/Onboarding/EulaModal.tsx
@@ -18,7 +18,7 @@ import { Button, RTLEnabledText } from "../components"
 
 import { Icons, Images } from "../assets"
 import { Spacing, Buttons, Colors, Typography, Forms } from "../styles"
-import { EULA_URL } from '../constants/eula';
+import { EULA_URL } from "../constants/eula"
 
 type CloseModalIconProps = {
   closeModal: () => void
@@ -78,20 +78,20 @@ const Checkbox: FunctionComponent<CheckboxProps> = ({
 const Spinner = () => {
   return (
     <ActivityIndicator
-      size={'large'}
+      size={"large"}
       color={Colors.darkGray}
       style={styles.activityIndicator}
-      testID={'loading-indicator'}
+      testID={"loading-indicator"}
     />
-  );
-};
+  )
+}
 
 const eulaPicker = (selectedLocale: string): string => {
   if (selectedLocale) {
-    return `${EULA_URL}/${selectedLocale}`;
+    return `${EULA_URL}/${selectedLocale}`
   }
-  return `${EULA_URL}/en`;
-};
+  return `${EULA_URL}/en`
+}
 
 type EulaModalProps = {
   selectedLocale: string
@@ -146,9 +146,7 @@ const EulaModal: FunctionComponent<EulaModalProps> = ({
                 source={{ uri: eulaPath }}
                 startInLoadingState
                 renderLoading={Spinner}
-                onShouldStartLoadWithRequest={
-                  shouldStartLoadWithRequestHandler
-                }
+                onShouldStartLoadWithRequest={shouldStartLoadWithRequestHandler}
               />
             </View>
           </SafeAreaView>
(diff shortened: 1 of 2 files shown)


src/constants/eula.ts
```diff
@@ -0,0 +1 @@
+export const EULA_URL = "https://covid-safe-paths-eula.herokuapp.com"
```
aledustet, line 1: Could this be a configuration item? In other words, will this be something that could potentially change from HA to HA or even from one country build to the other? Could we benefit from having the constant as part of the environment variables? If we don't and it is a static value, we might also move this constant to the `Eula` component since it is only used there and this file seems too specific in a too general folder. Thoughts?

src/Onboarding/EulaModal.tsx
```diff
@@ -225,6 +219,13 @@ const styles = StyleSheet.create({
   checkboxText: {
     ...Forms.checkboxText,
   },
+  activityIndicator: {
```
aledustet, line 222: I believe there's a PR #31 that works on this it might be worth merging that design with this?

src/Onboarding/EulaModal.tsx
```diff
@@ -145,15 +141,13 @@ const EulaModal: FunctionComponent<EulaModalProps> = ({
...
-                <WebView
-                  style={{ flex: 1 }}
-                  source={{ html }}
-                  onShouldStartLoadWithRequest={
-                    shouldStartLoadWithRequestHandler
-                  }
-                />
-              )}
+              <WebView
+                style={{ flex: 1 }}
```
aledustet, line 145: How do we feel about moving this style to the styles group? It feels like overkill to have but one property but it might help with consistency.

src/Onboarding/EulaModal.tsx
```diff
@@ -77,14 +75,22 @@ const Checkbox: FunctionComponent<CheckboxProps> = ({
...
 
-const EULA_FILES: Record<AvailableLocale, string> = {
-  ["en"]: en,
-  ["es_PR"]: es_PR,
-  ["ht"]: ht,
+const eulaPicker = (selectedLocale: string): string => {
+  if (selectedLocale) {
+    return `${EULA_URL}/${selectedLocale}`
+  }
+  return `${EULA_URL}/en`
```
aledustet, line 93:
```suggestion
const eulaPicker = (selectedLocale? = "en"): string => {
    return `${EULA_URL}/${selectedLocale}`
}
```

Could this work? Maybe simplifies the method?

src/Onboarding/EulaModal.tsx
```diff
@@ -1,26 +1,24 @@
...
 } from "react-native"
 import { useTranslation } from "react-i18next"
-import loadLocalResource from "react-native-local-resource"
 import WebView, { WebViewNavigation } from "react-native-webview"
 import { SvgXml } from "react-native-svg"
 
 import { Button, RTLEnabledText } from "../components"
-import en from "../locales/eula/en.html"
-import es_PR from "../locales/eula/es_PR.html"
-import ht from "../locales/eula/ht.html"
```
aledustet, line 20: Should we also remove these files since now the EULA is being fetched from the web?

src/Onboarding/EulaModal.tsx
```diff
@@ -1,26 +1,24 @@
...
   Modal,
+  ActivityIndicator,
   StyleSheet,
   View,
   SafeAreaView,
   StatusBar,
   Image,
 } from "react-native"
 import { useTranslation } from "react-i18next"
-import loadLocalResource from "react-native-local-resource"
```
aledustet, line 13: Oooo maybe we can delete one dependency here 🤞  Do we know if it's used anywhere else?

aledustet (commented):
Awesome! Excited about this change. 
I think we could benefit from taking a pass at `Eula.spec.tsx` since we won't need to mock any of the HTML files, perhaps this will greatly simplify the test? We could also get rid of the `html.d.ts` on the root folder if we won't be importing HTML files anymore. And it might be worth checking the uses of the `react-native-local-resource` since it could simplify the dependencies. Happy to talk more about this.

octohub:
@zhaccc There was a change made to the build checks so you'll have to merge the latest from `develop` and push to this branch so the `build_debug_android` passes.  Sorry for the hassle!

PR #35, opened 2020-07-22
Display self assessment according to jurisdiction requirements

zhaccc:
#### Description:

This PR hides the self-assessment feature according to jurisdiction requirements. 
Also, we hide the next `Next Steps` in case we don't have HA advice URL or if we have self-assessment hidden.

#### Screenshots:

**Hidden:**

**Displayed:**


src/utils/index.ts
```diff
@@ -11,4 +14,20 @@ const isPlatformAndroid = (): boolean => {
   return Platform.OS === "android"
 }
 
-export { DateTimeUtils, StorageUtils, isPlatformiOS, isPlatformAndroid }
+const displaySelfAssessment = (): boolean => {
```
johnschoeman, line 17: What do you think about inlining these to the files inwhich they are used? They arent really utilities and it would be easier to understand the business logic if it was colocated.

src/utils/index.ts
```diff
@@ -11,4 +14,20 @@ const isPlatformAndroid = (): boolean => {
...
+  return DISPLAY_SELF_ASSESSMENT === "true"
+}
+
+const displayNextSteps = (): boolean => {
+  // converting url string to boolean to make this more explicit and to please typescript
```
johnschoeman, line 22: What do you think about not having this comment?  It is extraneous as this is clear from the type signatures.

src/ExposureHistory/ExposureDatumDetail.tsx
```diff
@@ -104,6 +103,25 @@ interface NoDataExposureDetailProps {
   datum: NoData
 }
 
+interface NextStepsButtonProps {
+  buttonText: string
+  onPress: () => void
+}
+
+const NextStepsButton = ({ onPress, buttonText }: NextStepsButtonProps) => {
```
johnschoeman, line 111:
```suggestion
const NextStepsButton: FunctionComponent<NextStepsButtonProps> = ({ onPress, buttonText }) => {
```

example.env.bt
```diff
@@ -1,5 +1,6 @@
 TRACING_STRATEGY=bt
 DOWNLOAD_PATH=pc
+DISPLAY_SELF_ASSESSMENT=true
```
johnschoeman, line 3: Should we also add these to the mobile-resources-repo?

src/utils/index.ts
zhaccc: I am okay with that, but those files are going to be used in a few places, are we ok repeating the same constant a few times?

johnschoeman: yes totally.

johnschoeman: well the constant is defined in the env file. so its fine to repeat the `env.CONFIG_VAR` a few times, if it improves readability and maintainability.

example.env.bt
tparesi: @zhaccc https://github.com/Path-Check/pathcheck-mobile-resources

PR #37, opened 2020-07-22
Add error boundaries

zhaccc:
#### Description:

This PR adds error boundaries and default error screen to the application.
**NOTE:** This PR still waits for the design for the default error screen.


johnschoeman:
@devinjameson , this might a great one for you to pair on with @zhaccc

src/ErrorBoundaries/ErrorBoundaries.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ErrorFallbackComponent } from "./ErrorFallbackComponent"
+
+type Props = {
+  children: JSX.Element
+  onError?: () => void
+}
+
+type State = { error: Error | string; hasError: boolean }
+
+export class ErrorBoundary extends React.Component<Props, State> {
```
johnschoeman, line 12: What do you think about using a functional component here?

johnschoeman: or do errorboundaries still need to be extended from React.Component?

src/ErrorBoundaries/ErrorFallbackComponent.tsx
```diff
@@ -0,0 +1,68 @@
+import React from "react"
+import { SafeAreaView, View, StyleSheet } from "react-native"
+import { useTranslation } from "react-i18next"
+import { Button } from "../components/Button"
+import { RTLEnabledText } from "../components/RTLEnabledText"
+
+import { Colors, Spacing, Typography } from "../styles"
+
+type Props = { error: Error | string; resetError: () => void }
```
johnschoeman, line 9: an interface might be a better fit here, or a least thats the convention thats been established in other components.

src/ErrorBoundaries/ErrorFallbackComponent.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const ErrorFallbackComponent = ({
```
johnschoeman, line 11:
```suggestion
export const ErrorScreen: FunctionComponent<ErrorScreenProps> = ({
```

src/locales/en.json
```diff
@@ -377,5 +377,10 @@
     "more": "More",
     "exposure_history": "Exposure History",
     "more_info": "More Info"
+  },
+  "errors": {
+    "title": "Whoops!",
```
johnschoeman, line 382:
```suggestion
    "title": "Unknown Error",
```

I understand that this still needs to go through design and stuff, but its possible to see this during a demo or something before that happens and given this is an app for health related issues a more formal tone is more appropriate here.

src/ErrorBoundaries/ErrorBoundaries.tsx
zhaccc: Yeah, unfortunately, we still don't have any hooks to support error boundaries.

PR #39, opened 2020-07-22
Clean up MainNavigator

devinjameson:


src/navigation/MainNavigator.tsx
```diff
@@ -21,29 +21,33 @@ const SCREEN_OPTIONS = {
 
 const MainNavigator: FunctionComponent = () => {
   const { isComplete } = useOnboardingContext()
+  const isOnboardingComplete = isComplete
```
johnschoeman, line 24: what do you think about updating this in the OnboardingContext its self?

so we could say something like:
```
const { isOnboardingComplete } = useOnboardingContext()
```

devinjameson: @johnschoeman Thoughts on https://github.com/Path-Check/gaen-mobile/pull/39/commits/41c9d907718ce575373c7cc41e231b69989450f2?

johnschoeman: looks solid!

PR #41, opened 2020-07-22
Remove Unnecessary GPS Files

octohub:
#### Description:

Remove GPS files that are no longer needed since the repo split.


octohub:
@johnschoeman I tried making the change to `gaen-mobile/android/app/src/res/` but it caused issues with `yarn run-android`. Didn't want to spend too much time on it as it's not a big deal. For sure something to come back to later. Just wanted to get a solid first pass done. As for the resources, I did a quick scan and I think we are all good to go.

@mattThousand Great catch, just made the change brother.

johnschoeman:
> @johnschoeman I tried making the change to `gaen-mobile/android/app/src/res/` but it caused issues with `yarn run-android`. Didn't want to spend too much time on it as it's not a big deal. For sure something to come back to later. Just wanted to get a solid first pass done. As for the resources, I did a quick scan and I think we are all good to go.

Tots ma goats.  lets cut a card to capture this for future work: https://trello.com/c/c7kHN5hs/221-as-a-dev-team-we-dont-have-a-single-bt-flavor

PR #42, opened 2020-07-22
Ignore iOS tests if only Android files changed.

octohub:
#### Description:
As documented [here](https://github.community/t/need-help-with-conditional-workflows-using-paths/16650) this will prevent PRs that contain only changes to the Android folder from running iOS tests. This will significantly reduce check time for Android specific PRs.

To confirm this works I think we need to merge it, then I need to make a change in the Android directory and submit a PR.


johnschoeman (approved):
🐐

mattThousand (approved):
💯

octohub:
Leaving this PR open while we see if we have any luck with this Github support post:
https://github.community/t/force-succeed-required-workflow-when-skipped/124055

aledustet (approved):
💪 :shipit: 

One question, aren't these unit tests for the iOS codebase? Shouldn't we also ignore the src folder?

PR #43, opened 2020-07-22
Add script to configure application builds

aledustet:
Why:
----
When we are getting ready to release a new build of the applications we need to update certain characteristics of them. Like the display name and the bundle identifier. Since this depends of each particular HA we would like this process to be automated based on the environment.

This Commit:
----
- Refactor plist reading and writing
- Refactor read and write from xml
- Add replace string in file helper
- Reset android application name to default BT
- Update applicationId on build.gradle
- Update bundleid on `COVIDSafePaths.xcodeproj/project.pbxproj`

How to test:
----
- Run the `./bin/set_ha.sh {HA_ABBREVIATION}` command and verify that the display names and the bundle identifiers and application id have been updated accordingly.

Screen Shot 2020-07-22 at 4 06 15 PMScreen Shot 2020-07-22 at 4 06 25 PM


android/app/src/bt/res/values/strings.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources>
   <!-- Shown on app info, app switcher, etc -->
-  <string name="app_name" translatable="false">PathCheck MN</string>
+  <string name="app_name" translatable="false">PathCheck BT</string>
```
aledustet, line 4: This was updated in a previous commit and we think by default should just be BT right?

johnschoeman: yeah, it likely should be read from a config file at somepoint to avoid this issue.

bin/configure_builds.sh
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env ruby
+#
+# Update the applications display name based on the release env file
+#
+# Usage
+#
+#   bin/set_display_names <?env>
```
johnschoeman, line 7:
```suggestion
#   bin/configure_builds [env]
```

bin/configure_builds.sh
```diff
@@ -0,0 +1,183 @@
...
+#
+# Example
+#
+#   bin/fetch_env
```
johnschoeman, line 11:
```suggestion
#   bin/configure_builds pc
```

bin/configure_builds.sh
```diff
@@ -0,0 +1,183 @@
...
+
+# Constants
+ENV_FILE = ARGV[0] || ".env.bt.release"
+PLIST_PATH = "./ios/BT/Info.plist"
+ANDROID_STRINGS_PATH="./android/app/src/bt/res/values/strings.xml"
+SEPARATOR = "##################################################################"
+
+def failure_message(message:)
+  puts SEPARATOR
+  puts "🚫 #{message}"
```
johnschoeman, line 29: emoji!

bin/set_ha.sh
```diff
@@ -22,4 +22,4 @@ HA_LABEL = ARGV[0]
```
johnschoeman, line 22: should we update the description in this file?

android/app/src/bt/res/values/strings.xml
johnschoeman: thats what this pr does!

octohub:
Just tested, great work!

PR #45, opened 2020-07-22
Use tracingStrategy to access native module

amandabeiner:
#### Description:

We currently have a TracingStrategyContext that is responsible for
determining whether the tracing strategy is gps or GAEN via the
environment. This is no longer necessary now that this repo is
exclusively uses the GAEN strategy.

This commit:
- Removes the unnecessary TracingStrategyContext
- Updates strategy dependent contexts (like Exposure and Permissions
contexts) to import their strategy directly instead of relying on props.
- Updates direct references to the `nativeModule` to be called via the
tracingStrategy instead.

#### Linked issues:

Fixes: [use btnativemodule via tracing strategy](https://trello.com/c/lVJhV0aX/220-only-use-btnativemodule-via-tracingstrategy)

#### How to test:
- The app should build
- Tests should pass
- Affected user flow code input should work as expected
- Affected user flow submit diagnosis should work as expected


devinjameson:
Nice!

src/AffectedUserFlow/CodeInput.spec.tsx
```diff
@@ -9,10 +9,22 @@ import * as API from "./verificationAPI"
 import * as NativeModule from "../gaen/nativeModule"
 import * as Hmac from "./hmac"
 import { Screens } from "../navigation"
+import { ExposureProvider } from "../ExposureContext"
```
aledustet, line 12: Sweet moving the dependency up 🥳  I was reading through the test body wich is a bit long but not dry at all (there's a pretty cool blogpost about it 😄  https://thoughtbot.com/blog/the-case-for-wet-tests) And we are actually mocking the `NativeModule` there 😨  How do we feel about taking this up a notch and exporting the raw context from here and create a provider around the tests that need it and mock the methods necessary to that context? We might add a factory like we had the tracing provider with a few mocked methods like the ones right here in this file and we can add the ones we need. Something like:

```
const exposureContext = exposureContextFactory.build({ getExposureKeys: jest.fn().mockResolvedValueOnce([])});
...
      const { getByTestId, getByLabelText } = render(
        <ExposureContext.Provider
          value={exposureContext}
        >
          <AffectedUserProvider>
            <CodeInputScreen />
          </AffectedUserProvider>
        </ExposureContext.Provider>,
      )
...
```
That way we can get our mocking closer to the examples and I feel this would push this great effort in moving this to a context to a whole new level. Sorry for the rant I'm not sure I'm making sense but happy to 🍐  or just talk more about this after 😄

src/AffectedUserFlow/PublishConsent/PublishConsentForm.spec.tsx
```diff
@@ -4,12 +4,25 @@ import "@testing-library/jest-native/extend-expect"
 import { useNavigation } from "@react-navigation/native"
 
 import PublishConsentForm from "./PublishConsentForm"
-import * as NativeModule from "../../gaen/nativeModule"
 import { Screens } from "../../navigation"
+import { ExposureProvider } from "../../ExposureContext"
 
 afterEach(cleanup)
 
+let mockSubmitDiagnosis: jest.Mock | undefined
```
aledustet, line 12: I think ... think that if we expose the raw context and use the factory strategy for this we could avoid having to mock things in advance and we can put these closer to where they are used. Curious to try that experiment.

src/AffectedUserFlow/CodeInput.spec.tsx
amandabeiner: @aledustet it turns out that with the new mock implementation, the mocking the native module was totally unused, so I deleted it and the tests still passed 🎉 !

However, I think your suggestion is way better anyway, so I think I implemented what you suggested in [4a5c25b](https://github.com/Path-Check/gaen-mobile/pull/45/commits/4a6c25bd2e0185245b768c3d11a698200ccf2bf9). Would you mind taking another look at that commit?

aledustet: You killed it!!! 🎉

PR #49, opened 2020-07-23
Pull URL Info fro ENV Variables and WIP Key Uploader Logic

octohub:
#### Description:
This is a WIP progress. It uses the ENV variables to build the key URLs. It also starts the implementation of the key uploader logic from the Google EN sample project.


android/app/src/bt/java/org/pathcheck/covidsafepaths/MainActivity.java
```diff
@@ -139,4 +172,97 @@ private void handleExposureStateChanged(boolean enabled) {
...
+//                                    getApplication().getString(R.string.generic_error_message));
+//                            postInflight(false);
+                  return;
+                }
+                ApiException apiException = (ApiException) exception;
+                if (apiException.getStatusCode()
+                        == ExposureNotificationStatusCodes.RESOLUTION_REQUIRED) {
+                  showPermissionShareKeys(apiException);
+                } else {
+//                            Log.w(TAG, "No RESOLUTION_REQUIRED in result", apiException);
```
aledustet, line 206: Is this code meant to be commented out? Could we just remove it instead and perhaps rid of the else branch?

android/app/src/bt/java/covidsafepaths/bt/exposurenotifications/DebugMenuModule.java
```diff
@@ -67,7 +85,7 @@ public void onSuccess(Boolean result) {
 
             @Override
             public void onFailure(Throwable t) {
-                callback.invoke(t.getMessage(), null);
+                //callback.invoke(t.getMessage(), null);
```
aledustet, line 88: Should we just delete the code instead? We could always revisit with the magic 🧙‍♀️ of git later

bin/fetch_ha_env.sh
```diff
@@ -61,7 +61,10 @@ end
 
 def fetch_and_write_file(filename, remote_url)
   open(filename, 'w') do |f|
-    Open3.popen2e("curl", "-s", remote_url) do |_, stdout_and_err, wait_thr|
+    Open3.popen2e("curl", "-f", remote_url) do |_, stdout_and_err, wait_thr|
```
aledustet, line 64: 👏

octohub:
@aledustet You're right about the comments. As this is just a WIP PR, I want to keep those comments there for now. This way I know I need to address them in a PR I will be working on next.

PR #54, opened 2020-07-23
Stop use of component for translations

aledustet:
Why:
----
The main reason is that we were missing the labels on the buttons for the Self-Assessment flow, but then more reasons appeared.
We have a script that examines the translation keys used throughout the application and updates the ones that are new or unused for uploading and completing the rest of such translations. We would like to use this script to automate the task of finding unused translation keys but we are seeing erroneous output when executed.
The reason for it is that the parser is encountering a `Trans` component that comes directly from the library creators but that the parsers do not recognize. In our particular case we are seeing that a key like:

```
<0></0><1></1>: <0></0><1></1>
```

Is being added when executing `yarn i18next`. There are [issues] documenting extraneous behavior with this component. Since the usage we were giving it was pretty small, we might benefit from not using it and avoid fighting the issue while gaining confidence in our `yarn i18next` script.

[issues]: https://github.com/i18next/i18next-parser/issues

This Commit:
----
- Removes the usage of the `Trans` component from i18next
- Re-arrange action button text for EmergencyAssessment screen to include the styled texts while not making use of the i18next syntax for the translation component

Reviewers:
----
I worked a bit to make the parser agree with the use of the `Trans` component but hit a brick wall and couldn't advance further. It appeared that the use of this component didn't grant all the effort so I opted for removing it. Happy to continue tinkering further if the benefit grants it. I tried to leave the component I changed as unchanged as possible since I'm planning on doing an overhaul on that folder soon. This one was just to correct an error we currently have on the self-assessment flow.

Screen Shot 2020-07-23 at 2 11 41 PM

### DIFF
--- a/src/SelfAssessment/EmergencyAssessment.js
+++ b/src/SelfAssessment/EmergencyAssessment.js
@@ -1,13 +1,12 @@
 import React from "react"
 import { View, StyleSheet } from "react-native"
-import { Trans, useTranslation } from "react-i18next"
+import { useTranslation } from "react-i18next"
 
 import {
   QUESTION_KEY_AGREE,
   SCREEN_TYPE_RADIO,
   SCREEN_TYPE_EMERGENCY,
 } from "./constants"
-import survey_en from "./survey.en.json"
 import { Info } from "./Info"
 import { InfoText } from "./InfoText"
 import { Button } from "./Button"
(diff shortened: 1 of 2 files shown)


src/SelfAssessment/EmergencyAssessment.js
```diff
@@ -42,17 +36,21 @@ export const EmergencyAssessment = ({ navigation }) => {
         <ChoiceButtons
           agreePress={handleAgreePress}
           agreeTitle={
-            <TranslationButtonText
-              translator={t}
-              text={"assessment.agree_option_agree"}
-            />
+            <RTLEnabledText style={styles.boldText}>
+              {t("assessment.i_am") + " "}
```
aledustet, line 40: This is because there's a rule about not rendering text outside a `Text` component but since this is not syntactically a `Text` component `eslint` is not happy about it. I guess I could put this in a variable outside, happy to do so.

johnschoeman: ive been using consts to get around this, but maybe we want to remove the eslint rule?

PR #56, opened 2020-07-23
Fix behavior leaving a string instead of a boolean

aledustet:
Why:
----
When we render the navigators, the logic gate is being run against a variable that can be assigned as a string if both environment variables are present. We need to treat it as a boolean check to avoid rendering errors on the Navigator component.

This Commit:
----
- Casts environment variables check to determine whether to display the self assessment as boolean

### DIFF
--- a/src/ExposureHistory/ExposureDatumDetail.tsx
+++ b/src/ExposureHistory/ExposureDatumDetail.tsx
@@ -42,8 +42,9 @@ const PossibleExposureDetail = ({
   const exposureDurationText = DateTimeUtils.durationToString(duration)
   const navigation = useNavigation()
   const { t } = useTranslation()
-  const displayNextSteps =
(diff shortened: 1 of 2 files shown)


src/ExposureHistory/ExposureDatumDetail.tsx
```diff
@@ -42,8 +42,9 @@ const PossibleExposureDetail = ({
   const exposureDurationText = DateTimeUtils.durationToString(duration)
   const navigation = useNavigation()
   const { t } = useTranslation()
-  const displayNextSteps =
-    env.DISPLAY_SELF_ASSESSMENT === "true" || env.AUTHORITY_ADVICE_URL
+  const displayNextSteps = Boolean(
+    env.DISPLAY_SELF_ASSESSMENT === "true" || env.AUTHORITY_ADVICE_URL,
+  )
```
aledustet, line 47: Maybe this wants to leave in a configuration context, I have a feeling that this won't be the single "exception" to the flow we will have to handle

PR #60, opened 2020-07-24
 Remove Button component from SelfAssessment

devinjameson:
#### Description:

This commit removes the `Button` component from the SelfAssessment in favor of
using the global `Button component`.


src/SelfAssessment/EmergencyAssessment.js
```diff
@@ -51,41 +50,20 @@ export const EmergencyAssessment = ({ navigation }) => {
...
-            <RTLEnabledText style={styles.boldText}>
-              {iAmNotLabel}
-              <RTLEnabledText style={styles.regularText}>
-                {t("assessment.experiencing_any_symptoms")}
-              </RTLEnabledText>
-            </RTLEnabledText>
-          }
-          backgroundColor={Colors.white}
-          textColor={Colors.black}
+          label={t("assessment.not_experiencing_any_symptoms")}
```
aledustet, line 64: Not sure if there was a need for having the label broken apart here but visually it had the `I am` and the `I am not` in bold, maybe worth checking that this change is safe to make with product

src/locales/en.json
```diff
@@ -23,8 +23,8 @@
     "emergency_cta": "Call 911",
     "emergency_description": " Based on your reported symptoms, you should seek care immediately.",
     "emergency_title": "Call Emergency Services",
-    "experiencing_any_symptoms": "experiencing any of these symptoms",
-    "experiencing_symptoms": "experiencing at least one of these symptoms",
+    "not_experiencing_any_symptoms": "I am not experiencing any of these symptoms",
+    "experiencing_symptoms": "I am experiencing some of these symptoms",
     "i_am": "I am",
     "i_am_not": "I am not",
```
aledustet, line 29: If it is ok to join the two labels these two are not used anymore so we could delete it if you run `yarn i18n:extract` this process is made automatically

src/SelfAssessment/EmergencyAssessment.js
devinjameson: Great catch can you check out this commit reverting this change to use bold text?

a0f4858dcbf51fdffd37e84bbe2b505c691f88ea

PR #61, opened 2020-07-24
[Android] Toggle Exposure Notifications

juancruzgs:
#### Description:

Add method to toggle exposure notifications on Android
Functionality already implemented on React Native & iOS

#### How to test:

Go to the Dev Menu and tap on `Toggle Exposure Notifications` button

### DIFF
--- a/android/app/src/bt/java/covidsafepaths/bt/exposurenotifications/DebugMenuModule.java
+++ b/android/app/src/bt/java/covidsafepaths/bt/exposurenotifications/DebugMenuModule.java
@@ -1,10 +1,12 @@
 package covidsafepaths.bt.exposurenotifications;
 
+import android.app.Activity;
 import android.util.Log;
 
 import com.facebook.react.bridge.Callback;
 import com.facebook.react.bridge.Promise;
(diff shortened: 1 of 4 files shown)


juancruzgs:
Created as draft until my Google Account is whitelisted to test the app

octohub:
Hey brother, great start. Couple things:
1. Use `handleExposureStateChanged` in `MainActivity` which bubbles up to the JS layer that the state has changed.
2. Look at `showPermission` in `MainActivity`. To enable EN, we need to bubble up a permission dialog from the EN SDK. And once the user accepts it, we then need to call `start()` again. This will be hard to test on your end because your account is not whitelisted yet. But you can wire things up.

juancruzgs:
@octohub I made a refactor to move some code from `MainActivity` to `ExposureNotificationClientWrapper`.
In that way we won't be duplicating code and each time `start()` or `stop()` is called it will automatically emit an event to the JS layer and show permission dialog if the call fails.

Does that make sense?
I haven't tested this yet

juancruzgs:
I've been able to test it
screenshot

PR #62, opened 2020-07-24
Convert AssessmentStart and InfoText to Typescript

devinjameson:
#### Description:

This commit converts the `SelfAssessment/AssessmentStart` and
`SelfAssessment/InfoText` files to Typescript.

#### How to test:

- Tests should pass

### DIFF
--- a/src/SelfAssessment/AssessmentStart.spec.tsx
+++ b/src/SelfAssessment/AssessmentStart.spec.tsx
@@ -1,9 +1,13 @@
-import { render, fireEvent } from "@testing-library/react-native"
 import React from "react"
+
+import { render, fireEvent } from "@testing-library/react-native"
 import { I18nextProvider } from "react-i18next"
+import { useNavigation } from "@react-navigation/native"
 
 import i18n from "../locales/languages"
-import { AssessmentStart } from "./AssessmentStart"
+import AssessmentStart from "./AssessmentStart"
+
+jest.mock("@react-navigation/native")
 
 describe("AssessmentStart", () => {
   it("displays the info for the assesment start", () => {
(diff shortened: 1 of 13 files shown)


src/SelfAssessment/AssessmentStart.spec.tsx
```diff
@@ -1,9 +1,15 @@
...
+
+import { render, fireEvent } from "@testing-library/react-native"
 import { I18nextProvider } from "react-i18next"
+import { useNavigation } from "@react-navigation/native"
 
 import i18n from "../locales/languages"
-import { AssessmentStart } from "./AssessmentStart"
+import AssessmentStart from "./AssessmentStart"
+
+const navigationSpy = jest.fn()
```
aledustet, line 10: How do you feel about moving this value closer to where it is used? I feel that it will be easier for readers to gain context on what is being tested if it happens closer to where is exercised.

src/SelfAssessment/AssessmentStart.spec.tsx
```diff
@@ -1,9 +1,15 @@
...
+jest.mock("@react-navigation/native")
```
aledustet, line 11: You will most likely need to leave this line up here because even if we don't use the exact value for the mock in each test, the hook is bein called.

src/SelfAssessment/AssessmentStart.tsx
```diff
@@ -1,19 +1,22 @@
...
 
 import { Colors } from "../styles"
 import { Icons, Images } from "../assets"
 
-/** @type {React.FunctionComponent<{}>} */
-export const AssessmentStart = ({ navigation }) => {
-  let { t } = useTranslation()
+const AssessmentStart: FunctionComponent = () => {
+  const { t } = useTranslation()
+  const navigation = useNavigation()
```
aledustet, line 15: perfect

src/SelfAssessment/InfoText.tsx
```diff
@@ -1,11 +1,16 @@
...
 
 import { Typography } from "../styles"
 
-export const InfoText = ({ title, description }) => {
+interface InfoTextProps {
+  title: string
+  description: string
+}
+
+const InfoText: FunctionComponent<InfoTextProps> = ({ title, description }) => {
```
aledustet, line 13: 💪

src/SelfAssessment/AssessmentStart.spec.tsx
devinjameson: Ah I see thank you!

src/SelfAssessment/AssessmentStart.spec.tsx
devinjameson: Yes good call!

PR #67, opened 2020-07-27
Fix the 911 dial for android and ios

zhaccc:
#### Description:

Fixed linking from app to dialer.


src/SelfAssessment/endScreens/Emergency.js
```diff
@@ -14,7 +14,7 @@ export const Emergency = () => {
   let { t } = useTranslation()
 
   // TODO: This would need to be localized per country
```
aledustet, line 16: I'm creating a ticket on trello with a task for this TODO How do you feel about removing it in this PR?

zhaccc: Sounds good!

PR #70, opened 2020-07-27
Display application name, version on about screen

aledustet:
Why:
----
When a user navigates to the `About` screen, they should see the correct application name based on the build as-well-as the information of the version and the number of the build.

This Commit:
----
- Install react-native-device-info library
- Display the application name and the version and build from the device information
- Add tests for the about screen

Screen Shot 2020-07-27 at 10 20 52 AM

### DIFF
--- a/src/More/About.tsx
+++ b/src/More/About.tsx
@@ -8,14 +8,20 @@ import {
   Text,
   View,
 } from "react-native"
+import {
+  getApplicationName,
+  getBuildNumber,
+  getVersion,
+} from "react-native-device-info"
 
-import packageJson from "../../package.json"
 import { RTLEnabledText } from "../components/RTLEnabledText"
 
 import { Colors, Spacing, Typography } from "../styles"
 
 export const AboutScreen: FunctionComponent = () => {
   const { t } = useTranslation()
+
+  const versionInfo = `${getVersion()} (${getBuildNumber()})`
 
   return (
     <ScrollView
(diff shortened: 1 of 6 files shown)


src/More/About.tsx
```diff
@@ -8,23 +8,29 @@ import {
...
 
-import packageJson from "../../package.json"
 import { RTLEnabledText } from "../components/RTLEnabledText"
 
 import { Colors, Spacing, Typography } from "../styles"
 
 export const AboutScreen: FunctionComponent = () => {
   const { t } = useTranslation()
 
+  const versionInfo = `${getVersion()} ${getBuildNumber()}`
```
aledustet, line 24: Very arbitrary decision here, not sure if we want to add a `B` in the middle.

johnschoeman: maybe `version: 0.0.1 (18)`. as this is the same format that apple uses?

src/locales/en.json
```diff
@@ -115,7 +115,6 @@
     }
   },
   "label": {
-    "about_header_bluetooth": "PathCheck BT",
```
johnschoeman, line 118: 💯

src/More/About.tsx
johnschoeman: yeah actually, we likely want to do something like this as `0.0.1 1` could easily be mistaken for `0.0.11`

PR #75, opened 2020-07-27
Update pull and push scripts for lokalise

aledustet:
Why:
----
The push and pull actions for lokalise are not working as expected

This Commit:
----
- Remove location permission strings for iOS
- Extract to `src` not `app`
- Do not search under the `gps` flavor of android to upload files
- Update README portion of scripts with new env variables names
- Change some translations that mention location
- Add option to pull script to include just a set of languages
- Removed unused strings on iOS
- Upload translations from the src folder
- Ran push and updated master translations file on lokalise

Caveat:
----
This Commit does not include the result of running and commiting the `pull` script since it brings several files into the `iOS` and `android` projects that might not be necessary. There are more arguments like the languages to include in the pull we might add to the script.

### DIFF
--- a/ios/en.lproj/InfoPlist.strings
+++ b/ios/en.lproj/InfoPlist.strings
@@ -1,5 +1,1 @@
 "NSHumanReadableCopyright" = "Copyright © 2020 Path Check Inc. All rights reserved.";
-"NSLocationAlwaysAndWhenInUseUsageDescription" = "Your location history will be saved on your device, and only shared with others if you choose to do so.";
-"NSLocationAlwaysUsageDescription" = "Your location history will be saved on your device, and only shared with others if you choose to do so.";
-"NSLocationWhenInUseUsageDescription" = "Your location history will be saved on your device, and only shared with others if you choose to do so.";
-"NSMotionUsageDescription" = "We use your accelerometer for accurate location history.";
--- a/ios/en.lproj/Localizable.strings
+++ b/ios/en.lproj/Localizable.strings
@@ -1,4 +1,0 @@
-/* Body text of notification when app is closed on iOS */
-"ios.app_closed_alert_text" = "PathCheck is securely storing your GPS coordinates once every five minutes on this device.";
-/* Title of notification when app is closed on iOS */
-"ios.app_closed_alert_title" = "PathCheck was closed";
--- a/src/locales/pull.sh
+++ b/src/locales/pull.sh
@@ -4,7 +4,10 @@ set -e
 # Usage:
 #   - Create a READ ONLY a token at https://app.lokalise.com/profile
 #   - Run the command from the root of the project with:
-#     LOKALISE_TOKEN=<token> yarn i18n:pull
+#     LOKALISE_READ_TOKEN=<token> yarn i18n:pull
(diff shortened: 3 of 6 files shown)


ios/en.lproj/InfoPlist.strings
```diff
@@ -1,5 +1 @@
 "NSHumanReadableCopyright" = "Copyright © 2020 Path Check Inc. All rights reserved.";
```
aledustet, line 1: There are also a couple of notification strings for when the app is closed on the `Localizable.strings` file we might want to remove, more details on [trello](https://trello.com/c/Imm1CVLo)

src/locales/pull.sh
```diff
@@ -4,7 +4,10 @@ set -e
 # Usage:
 #   - Create a READ ONLY a token at https://app.lokalise.com/profile
 #   - Run the command from the root of the project with:
-#     LOKALISE_TOKEN=<token> yarn i18n:pull
+#     LOKALISE_READ_TOKEN=<token> yarn i18n:pull
+#   - To specify which languages do pull add the LANGUAGES with comma separated
+#     locale strings:
+#     LANGUAGES=en,es_PR,el yarn i18n:pull
```
aledustet, line 10: This is in order to filter the languages we support on the build only, otherwise we will add translation files to each one of the languages available on lokalise for android and iOs

ios/en.lproj/Localizable.strings
```diff
@@ -1,4 +0,0 @@
-/* Body text of notification when app is closed on iOS */
-"ios.app_closed_alert_text" = "PathCheck is securely storing your GPS coordinates once every five minutes on this device.";
```
aledustet, line 2: This was not used anywhere

src/locales/pull.sh
johnschoeman: Makes sense to me. But to make sure i fully understand the trade off, would there be any consequence to including all of the language files in lokalise, other than unused code?

And if the languages files are not being used in the app, should we stop translating them in lokalise? Seems like unnecessary work for the translation team? And this would also allow us to avoid needing the extra flags in the script.

aledustet: I see your point, I just don't know what the intent is on having all the translations available. It might be an organizational decision of having all the translations available for when they are used, this just increases the bundle size a little bit I don't think is such a big deal, I just added it to make sure we can update the ones that are ready. Since I believe that if a translation file is present on iOS and the user phone is using that language it will load it, and I'm not sure what the implications are for having the wrong translation vs defaulting to English.

aledustet: I made it so the extra flags are not required but I felt we might need them

alpita-masurkar: This makes sense to me too. If the translations for languages not yet included in any release are adding to the bundle size, there is no need to have the languages folder. If it helps and makes code less complicated, I can exclude the languages not needed before creating the pull request.
We can continue to have translations for such languages in Lokalise as there's no cost associated with it and it helps to have translations done regularly when we have the translators (example, we already had Russian and Spanish ready and updated for a while before they were actually needed for a release). But we can flag them or exclude them in the PR.

PR #76, opened 2020-07-27
(iOS) Debug Menu Updates

mattThousand:
### Why
We'd like to remove debug menu buttons that will cease to work when the gaen server verifies key submission requests

### This Commit
This commit removes the existing key submission button from the debug menu, and also updates the underlying method for forcing a crash on iOS

### DIFF
--- a/ios/BT/bridge/DebugMenuModule.m
+++ b/ios/BT/bridge/DebugMenuModule.m
@@ -18,7 +18,7 @@ RCT_REMAP_METHOD(fetchDiagnosisKeys,
 
 RCT_EXPORT_METHOD(forceAppCrash)
 {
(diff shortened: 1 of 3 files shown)


ios/BT/bridge/DebugMenuModule.m
```diff
@@ -18,7 +18,7 @@ @implementation DebugMenuModule
 
 RCT_EXPORT_METHOD(forceAppCrash)
 {
-  NSAssert(NO, @"Forced Crash (Debug)");
+  @throw [NSException exceptionWithName:NSGenericException reason:@"Forced Crash (Debug)" userInfo:nil];
```
aledustet, line 21: Should this one be part of the string extensions? Probably not since it's a debug one 😄

PR #77, opened 2020-07-27
Only allow code submission when en are enabled

johnschoeman:
Why:
Currently, if a user has not enabled exposure notifications for the app
it is possible to try to submit the verification code in the affected
user flow, which will necessarily error as the exposure notification api
must be accessible in order to submit exposure keys.

We would like to make it such that a user cannot submit a verification
code unless the en api is enabled. Additionally, if the user starts the
affected user flow without enabled exposure notifications, they should
be prompted to enable exposure notifications.

This commit:
Adds a conditional to the CodeInputScreen to conditionally render either
the CodeInputForm or an EnableExposureNotifications screen that has a
link to the settings for en notifications.

Co-Authored-By johnschoeman <johnschoeman1617@gmail.com>

---

### GIF

require-en-permissions

### DIFF
--- a/src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
+++ b/src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
@@ -1,6 +1,4 @@
-import React, { useState } from "react"
-import { useNavigation } from "@react-navigation/native"
-import { useTranslation } from "react-i18next"
+import React, { FunctionComponent, useState } from "react"
 import {
   ActivityIndicator,
   Alert,
@@ -12,14 +10,16 @@ import {
   View,
   Keyboard,
 } from "react-native"
-import { SafeAreaView } from "react-native-safe-area-context"
+import { useNavigation } from "@react-navigation/native"
+import { useTranslation } from "react-i18next"
 
-import { RTLEnabledText } from "../components/RTLEnabledText"
-import { useAffectedUserContext } from "./AffectedUserContext"
-import * as API from "./verificationAPI"
-import { calculateHmac } from "./hmac"
+import { RTLEnabledText } from "../../components/RTLEnabledText"
+import { useAffectedUserContext } from "../AffectedUserContext"
+import * as API from "../verificationAPI"
+import { calculateHmac } from "../hmac"
+import { useExposureContext } from "../../ExposureContext"
 
-import { Screens, Stacks } from "../navigation"
+import { Screens, Stacks } from "../../navigation"
 import {
   Spacing,
   Buttons,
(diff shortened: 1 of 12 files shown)


src/AffectedUserFlow/CodeInput/CodeInputScreen.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import CodeInputScreen from "./CodeInputScreen"
+import { AffectedUserProvider } from "../AffectedUserContext"
+import {
+  PermissionsContext,
+  ENPermissionStatus,
+} from "../../PermissionsContext"
+import { PermissionStatus } from "../../permissionStatus"
+
+afterEach(cleanup)
```
aledustet, line 13: I would hope 🤞  that this gets run automatically if it's not the case we might have something going on on our setup

src/AffectedUserFlow/CodeInput/CodeInputScreen.tsx
```diff
@@ -0,0 +1,38 @@
...
+import EnableExposureNotifications from "./EnableExposureNotifications"
+
+import { Colors } from "../../styles"
+
+const CodeInputScreen: FunctionComponent = () => {
+  const { exposureNotifications } = usePermissionsContext()
+
+  const hasExposureNotificationsEnabled = (): boolean => {
+    const enabledState: ENEnablement = "ENABLED"
+    return exposureNotifications.status[1] === enabledState
```
aledustet, line 16: This feels like it needs to know too much about the exposure notification object to ask a question to a provider we are already creating. Would it make sense(probably not on this PR) to add a way to ask this directly in that context?

src/PermissionsContext.tsx
```diff
@@ -78,6 +79,7 @@ const PermissionsProvider: FunctionComponent = ({ children }) => {
   useEffect(() => {
     const subscription = permissionStrategy.statusSubscription(
       (status: ENPermissionStatus) => {
+        console.log("got permission event:", status)
```
aledustet, line 82: Should we be logging this here?

src/AffectedUserFlow/CodeInput/CodeInputScreen.tsx
devinjameson: That's a nice idea. So we would move this logic to the context and just get something like `userHasExposureNotificationsEnabled` directly from the context?

src/PermissionsContext.tsx
devinjameson: Nope :)

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -214,18 +200,13 @@ const LoadingIndicator = () => {
...
-    width: "100%",
-    height: "100%",
-    resizeMode: "cover",
-    backgroundColor: Colors.faintGray,
-  },
   container: {
     height: "100%",
     justifyContent: "space-between",
     paddingHorizontal: Spacing.medium,
     paddingTop: Layout.oneTenthHeight,
```
devinjameson, line 207: Should we just be using a `Spacing` var here? IMO we should only use `Layout` vars if necessary for some reason.

src/AffectedUserFlow/CodeInput/CodeInputScreen.tsx
aledustet: Maybe, it could also make sense to parse it into an object, that might hold all the info, like having props for when it is available and not enabled or not available at all. It depends on the use case. I would shy away from having the method be too specific but maybe, in this case, specific is exactly what we want.

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
johnschoeman: We could do, but there is no spacing var that is around 1/10th the height of the screens at this point and I dont want to get too into the weeds fiddling with layout on this pr.

PR #79, opened 2020-07-28
No exposures screen

johnschoeman:
no-exposures

Still needs some styling and things, but adds the start of the no exposures screen.


src/ExposureHistory/History.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+        date: twoDaysAgo,
+      })
+
+      const exposures = [datum1]
+
+      const { getByTestId } = render(
+        <History exposures={exposures} lastDetectionDate={null} />,
+      )
+
+      expect(getByTestId("exposure-list")).not.toBeNull()
```
aledustet, line 41: I noticed we are asking for presence by negating the null equality, would the positive read better? `toBeDefined()`

src/ExposureHistory/History.tsx
```diff
@@ -64,7 +72,11 @@ const History: FunctionComponent<HistoryProps> = ({ lastDetectionDate }) => {
           </View>
         </View>
         <View style={styles.listContainer}>
```
aledustet, line 74: I think that this container and it's style might be better of inside the `NoExposures` and the `ExposuresList`, I feel like those will have very different style needs. Is probably fine but we should keep an eye for overriding this style inside those components, might be a sign that we don't need this style.

johnschoeman:
merged this on in directly.

PR #80, opened 2020-07-28
Implement no connectivity screen on next steps

zhaccc:
#### Description:

This PR implements connectivity check and displays a warning if there is no connectivity on the next steps page.
This needs design rework which can be easily implemented since we have all checks we need in place!


src/ExposureHistory/NextSteps.tsx
```diff
@@ -34,8 +35,21 @@ const CloseButton = () => {
...
+    null,
+  )
   const { t } = useTranslation()
   useStatusBarEffect("dark-content")
+
+  useEffect(() => {
+    const unsubscribe = NetInfo.addEventListener((state) => {
+      setConnectivity(state.isInternetReachable)
+    })
+    return () => {
```
devinjameson, line 48: Do you need this return or can you just have `unsubscribe()`?

zhaccc: When we return a function in `useEffect` React will run it when component unmounts, so we are using this to clean up this listener on unmount.

devinjameson: Oh interesting, didn't know that. Thanks for the explanation!

tparesi:
With the new Exposure History flow, you are probably going to have to rework this a bit. Sorry!

octohub:
@zhaccc There was a change made to the build checks so you'll have to merge the latest from `develop` and push to this branch so the `build_debug_android` passes.  Sorry for the hassle!

devinjameson:
@zhaccc I think you might want this on the ExposureDetails screen, but now it's on the NoExposures screen…

src/ExposureHistory/ExposureDetail.tsx
```diff
@@ -22,10 +23,26 @@ const ExposureDetail: FunctionComponent = () => {
...
+  )
+
   const {
     GAEN_AUTHORITY_NAME: healthAuthorityName,
     AUTHORITY_ADVICE_URL: healthAuthorityLink,
   } = env
+
+  useEffect(() => {
+    const unsubscribe = NetInfo.addEventListener((state) => {
+      // netInfo state comes as null while unresolved so to avoid flicker we only set component state
```
devinjameson, line 37: Thoughts on putting this in the commit comment instead to keep the code clean?

src/ExposureHistory/ExposureDetail.tsx
```diff
@@ -22,10 +23,26 @@ const ExposureDetail: FunctionComponent = () => {
   useStatusBarEffect("light-content")
   const { t } = useTranslation()
 
+  const [connectivity, setConnectivity] = useState<boolean | null | undefined>(
```
devinjameson, line 26:
```suggestion
  const [isConnected, setIsConnected] = useState<boolean | null | undefined>(
```

Thoughts on this naming scheme to stay consistent with the rest of the codebase?

src/ExposureHistory/ExposureDetail.tsx
```diff
@@ -22,10 +23,26 @@ const ExposureDetail: FunctionComponent = () => {
...
   const {
     GAEN_AUTHORITY_NAME: healthAuthorityName,
     AUTHORITY_ADVICE_URL: healthAuthorityLink,
   } = env
+
+  useEffect(() => {
+    const unsubscribe = NetInfo.addEventListener((state) => {
+      // netInfo state comes as null while unresolved so to avoid flicker we only set component state
+      // if the netInfo state is resolved to boolean
+      if (state.isInternetReachable !== null) {
```
devinjameson, line 39: Not sure but could this use `state.isInternetReachable && setConnectivity(state.isInternetReachable)`?

src/ExposureHistory/ExposureDetail.tsx
```diff
@@ -203,6 +226,10 @@ const style = StyleSheet.create({
   buttonContainer: {
     alignSelf: "flex-start",
   },
+  connectivityWarningText: {
+    color: Colors.danger100,
```
devinjameson, line 230:
```suggestion
    ...Typography.error,
```

src/ExposureHistory/ExposureDetail.tsx
zhaccc: We actually need to specifically check for null here because that is this method uninitialized state. Also its a bit more readable this way so I think it should be fine.

src/ExposureHistory/ExposureDetail.tsx
zhaccc: Good catch!

src/ExposureHistory/ExposureDetail.tsx
zhaccc: Hmm, it is a good point, but I think we should rather keep this in to tell the future developers why are we checking for `null` here as it is kinda confusing.

PR #90, opened 2020-07-28
Use "style" instead of "styles"

devinjameson:


devinjameson:
> omg, this was so fast.

SPEEDY BOI 🏃‍♂️

PR #92, opened 2020-07-29
Check en permission on app state change

johnschoeman:
Check en permission on app state change …
Why:
Currently if the user updates the permission for the app while the app
is backgrounded the application will not receive the onStatusUpdated
event from the native layer, which leads to a situation where the app
shows the incorrect permission status.

This commit:
Adds a subscription to the PermissionsContext using react-natives
AppState api and will check the permission status on every change of the
app state.

Note that we are still finding cases where the checkENPermission()
method is returning ["Authorized", "Enabled"] when we would expect
["Authorized", "Disabled"], which seems to be another bug unrelated to
the backgrounding behavior.

Co-Authored-By: mattthousand <matt@nicethings.io>
Co-Authored-By: amandabeiner <aebeiner@gmail.com>

### DIFF
--- a/src/PermissionsContext.tsx
+++ b/src/PermissionsContext.tsx
@@ -6,12 +6,12 @@ import React, {
   useContext,
   useCallback,
 } from "react"
-
+import { Platform, AppState } from "react-native"
 import {
   checkNotifications,
   requestNotifications,
 } from "react-native-permissions"
-import { Platform } from "react-native"
+
 import { PermissionStatus, statusToEnum } from "./permissionStatus"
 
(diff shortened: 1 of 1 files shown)


src/PermissionsContext.tsx
```diff
@@ -98,6 +100,19 @@ const PermissionsProvider: FunctionComponent = ({ children }) => {
     }
   }, [checkENPermission])
 
+  const handleAppStateChange = () => {
+    console.log("app state changed, checking permission")
```
aledustet, line 104: Do we need this logging to stay in?

johnschoeman: nope!

PR #94, opened 2020-07-29
Update permissions en status and request on home

johnschoeman:
Why:
There are some inconsistencies in how we are receiving the devices en
permission status.

This commit:
Changes the check of the devices exposureNotificationStatus to check
against the status directly vs. via the exposureNotificationsEnabled
method. We also directly request permissions in the app when the app is
unauthorized rather than making the user open the os settings and
manually updating the permissions.

Co-Authored-By: mattthousand <matt@nicethings.io>

### DIFF
--- a/ios/BT/ExposureManager.swift
+++ b/ios/BT/ExposureManager.swift
@@ -24,7 +24,7 @@ final class ExposureManager: NSObject {
   let manager = ENManager()
   
   var enabledState: EnabledState {
-    return manager.exposureNotificationEnabled ? .enabled : .disabled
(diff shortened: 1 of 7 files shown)


johnschoeman:
running locally, looks like if en are not authorized the request to the native layer requests authorizations intermittently.  cc: @mattThousand, no rush but looks like well need to revisit this and add back a conditional approach.

octohub:
@johnschoeman  There was a change made to the build checks so you'll have to merge the latest from `develop` and push to this branch so the `build_debug_android` passes.  Sorry for the hassle!

PR #96, opened 2020-07-29
Add promise to submiteExposureKeys

juancruzgs:
#### Description:

We were not reporting errors to the UI, so the user could never know if the upload succeeded or not.
This PR uses the Promise that is sent as a parameter and calls `resolve` or `reject` accordingly.


android/app/src/bt/java/org/pathcheck/covidsafepaths/MainActivity.java
```diff
@@ -148,7 +147,7 @@ public void onResolutionComplete(int requestCode, int resultCode) {
       }
     } else if(requestCode == RequestCodes.REQUEST_CODE_UPLOAD_KEYS){
       if (resultCode == RESULT_OK) {
-        share();
+        share(postDiagnosisKeysPromise);
       } else {
         // Don't share.
```
aledustet, line 152: Is the promise rejected on this branch?

juancruzgs: It's not, I'll add it
Thanks!

PR #103, opened 2020-07-30
Update exposure configuration (iOS)

mattThousand:
### Why
We'd like update the `attenuationDurationThresholds` on our exposure configuration. This change is temporary, as we'll ultimately be getting these values from a server

### This Commit
This commit updates our working exposure configuration to match the `attenuationDurationThresholds` values used by the Swiss PT – DP3T App:
https://github.com/DP-3T/dp3t-config-backend-ch/blob/55fbe7493775402804493eced4c8128ac15db9f8/dpppt-config-backend/src/main/java/org/dpppt/switzerland/backend/sdk/config/ws/model/GAENSDKConfig.java#L15

### DIFF
--- a/ios/BT/API/Model/ExposureConfiguration.swift
+++ b/ios/BT/API/Model/ExposureConfiguration.swift
@@ -16,7 +16,7 @@ extension ExposureConfiguration {
 
   static var placeholder: ExposureConfiguration = {
     ExposureConfiguration(minimumRiskScore: 0,
-                          attenuationDurationThresholds: [50, 70],
+                          attenuationDurationThresholds: [53, 60],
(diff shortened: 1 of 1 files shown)


ios/BT/API/Model/ExposureConfiguration.swift
```diff
@@ -16,7 +16,7 @@ extension ExposureConfiguration {
 
   static var placeholder: ExposureConfiguration = {
     ExposureConfiguration(minimumRiskScore: 0,
-                          attenuationDurationThresholds: [50, 70],
+                          attenuationDurationThresholds: [53, 60],
```
aledustet, line 19: these are very specific 😅  could we put them in named constants for "documentation" purposes?

PR #91, opened 2020-07-29
Show notification when exposure is detected

juancruzgs:
#### Description:

This PR will show a native notification when a new exposure is detected.
Changes:
- Updated the code to match latest updates in https://github.com/google/exposure-notifications-android
- Save `ExposureEntity` in Realm database to avoid showing the same notification multiple times
- Use API v1.5

#### Screenshots:

device-2020-07-29-105348

#### How to test:

Tap on the button "Detect exposures now" to trigger the download process right away (it will run after 24 hours otherwise)

PR #105, opened 2020-07-30
Fix lokalise pull script for android

aledustet:
Why:
----
The android file structure resulting from executing the script for the localization files was of the form:
`android/app/src/bt/res/values-{LANG_ISO}/android/app/src/bt/res/values-{LANG_ISO}/strings.xml`

This is incorrect, the files should be stored in the structure
`android/app/src/bt/res/values-{LANG-ISO}/strings.xml`

This Commit:
----
- Update arguments for pulling files for the android project and unzipping them
- Execute pull script for android new file structure

### DIFF
--- a/android/app/src/bt/res/values/strings.xml
+++ b/android/app/src/bt/res/values/strings.xml
@@ -4,11 +4,10 @@
   <string name="app_name" translatable="false">PathCheck BT</string>
   <!-- Shown below icon on launcher only -->
   <string name="app_name_short" translatable="false">PathCheck BT</string>
+  <string name="exposure_notification_channel_description">Exposure notification alerts.</string>
+  <string name="exposure_notification_channel_name">Exposure Notification</string>
+  <string name="exposure_notification_message">Someone you were near has tested positive for COVID-19. Tap for more info.</string>
+  <string name="exposure_notification_title">Possible COVID-19 exposure</string>
   <!-- Channel label for notifications shown in App > Notifications > Path Check BT -->
   <string name="notification_channel_description">PathCheck BT notifications</string>
-
-  <string name="exposure_notification_title">Possible COVID-19 exposure</string>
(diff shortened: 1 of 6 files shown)


android/app/src/bt/res/values/strings.xml
```diff
@@ -6,9 +6,4 @@
   <string name="app_name_short" translatable="false">PathCheck BT</string>
   <!-- Channel label for notifications shown in App > Notifications > Path Check BT -->
   <string name="notification_channel_description">PathCheck BT notifications</string>
-
-  <string name="exposure_notification_title">Possible COVID-19 exposure</string>
-  <string name="exposure_notification_message">Someone you were near has tested positive for COVID-19. Tap for more info.</string>
-  <string name="exposure_notification_channel_name">Exposure Notification</string>
-  <string name="exposure_notification_channel_description">Exposure notification alerts.</string>
```
juancruzgs, line 13: These strings are required, we still need to translate them

aledustet: If those are on master those should have been updated the last time we pushed, do we know when these were added?

juancruzgs: They were added here https://github.com/Path-Check/gaen-mobile/pull/91
I'm not sure how's the process to get them translated, should we add them to lokalise?

aledustet: It should be happening automatically when we merge to develop, let me double-check that

aledustet: The routes on the GitHub workflow file are wrong 🤦  so changes to the strings file on android are not being picked up. Nice catch! thanks, @juancruzgs will update this

johnschoeman: this may be related to this card: https://trello.com/c/6AAUYzag/200-lokalize-string-bugs-in-android, @aledustet , @juancruzgs

PR #109, opened 2020-07-30
Accept configuration for available locales

aledustet:
Why:
----
The language selection screen has a static group of available locales. Those are the same locales available for the users of the applications. This selection needs to be dynamic for each one of the builds of the applications since different geographic zones will have specific translation needs.

How:
----
We need to add an environment variable that can be set per HA to control which translations will be made available for each. With this change, the initialization of the translation service can not be made by default. It needs to happen as the application starts.

This Commit:
----
- Initialize i18Next on demand when the application starts not when the `languages` file loads
- Set the available locales from a configuration variable
- Fall-back to `en` if the provided locales are not valid or not available

Reviewers:
----
The way we are testing the application by checking the real translation value forces us to initialize the translation service with the test runner. This might seem like an overreach for the purpose of testing the units, but this was the way it was done before and I felt it was out of the scope of this PR. We can address it later and decide if it is correct for us to test the translation values vs a mock of the service that could print the key it receives.

Screen Shot 2020-07-30 at 2 06 17 PM


jest/setupFile.js
```diff
@@ -30,3 +32,5 @@ jest.mock(
   () => "push-notification-ios",
 )
 jest.mock("rn-fetch-blob", () => "Blob")
+
+initializei18next(["en"])
```
aledustet, line 36: This needs to happen for all of our test matches to work, also the `RTLEnabledText` accesses the direction of the text, this gets initialized with `i18Next`

src/locales/languages.ts
```diff
@@ -188,18 +150,55 @@ export function supportedDeviceLanguageOrEnglish(): Locale {
...
+
+export const initializei18next = async (
+  withLocales: Locale[],
+): Promise<void> => {
+  const config = {
+    interpolation: {
+      // React already does escaping
+      escapeValue: false,
+    },
+    lng: "en", // this is for initialization purposes only
```
aledustet, line 187: This gets overridden right away by `initializeLocale`

src/locales/languages.ts
```diff
@@ -188,18 +150,55 @@ export function supportedDeviceLanguageOrEnglish(): Locale {
...
 const getUserLocale = async (): Promise<string | null> => {
   return await StorageUtils.getUserLocaleOverride()
 }
-// detect user override
-getUserLocale().then((locale) => {
-  if (locale) {
-    setLocale(locale)
+
+const initializeLocale = (): void => {
+  // detect user override
```
johnschoeman, line 158: thoughts about removing this comment and extracting a named function instead?

example.env.bt
```diff
@@ -21,3 +21,4 @@ AUTHORITY_ADVICE_URL=https://pathcheck.org
 DEBUG_API_URL=https://example.debug.com/
 DEBUG_API_TOKEN=DEBUG_API_TOKEN
 
+LOCALES=en,es_PR
```
johnschoeman, line 24:
```suggestion
SUPPORTED_LOCALES=en,es_PR
```

aledustet:
Still can't get the CI to run 😢  anyone has any ideas?

PR #112, opened 2020-07-30
Remove Staging Variant

octohub:
#### Description:
Remove staging variant as it is no longer used.


juancruzgs:
You should also update `fastlane/Fastfile` to use release or debug instead of staging

PR #113, opened 2020-07-30
Save last processed file name

juancruzgs:
#### Description:

- Save in Realm the last file name we processed and expose a function to get it from React Native (for debugging/testing purposes)
- Do not download the file again if it was already processed
- Remove 20 batches per day limit. According to documentation the limit is per call, and we only call the method once a day.

#### How to test:

Open the debug menu and tap on "Show Last Processed File Path"

### DIFF
--- a/android/app/src/bt/java/covidsafepaths/bt/exposurenotifications/network/Uris.java
+++ b/android/app/src/bt/java/covidsafepaths/bt/exposurenotifications/network/Uris.java
@@ -37,13 +37,11 @@ import java.util.List;
 
 import covidsafepaths.bt.exposurenotifications.common.AppExecutors;
 import covidsafepaths.bt.exposurenotifications.storage.ExposureNotificationSharedPreferences;
-import covidsafepaths.bt.exposurenotifications.storage.RealmSecureStorageBte;
 
 /**
  * Encapsulates logic for resolving URIs for uploading and downloading Diagnosis Keys.
  */
 public class Uris {
-    private static final int MAX_KEY_BATCHES_PER_DAY = 20;
     private static final String TAG = "Uris";
     private static final Splitter WHITESPACE_SPLITTER =
             Splitter.onPattern("\\s+").trimResults().omitEmptyStrings();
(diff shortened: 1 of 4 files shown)


android/app/src/bt/java/covidsafepaths/bt/exposurenotifications/network/Uris.java
```diff
@@ -92,16 +90,10 @@ public Uris(Context context) {
...
                             // "exposureKeyExport-US/1589490000-00002.zip"
-                            final String zipFileName =  RealmSecureStorageBte.INSTANCE.getLastDownloadedKeyZipFileName();
-                            final int startIndex = getStartIndex(indexEntries, zipFileName);
-                            for (int i = startIndex; i < indexEntries.size(); i++) {
+                            // In case that we want to limit the number of downloaded files we can add some logic to check the latest file name here
+                            for (int i = 0; i < indexEntries.size(); i++) {
                                 final String indexEntry = indexEntries.get(i);
                                 uriList.add(baseDownloadUri.buildUpon().appendEncodedPath(indexEntry).build());
-
-                                if(uriList.size() == MAX_KEY_BATCHES_PER_DAY) {
```
juancruzgs, line 101: This is not needed anymore
According to the documentation there is only a limit of 6 calls to `provideDiagnosisKeys()` per day, but there is no limit on the number files we provide.
It won't have any impact since we only call the method once a day.
https://developers.google.com/android/exposure-notifications/exposure-notifications-api#methods

octohub:
@juancruzgs There was a change made to the build checks so you'll have to merge the latest from `develop` and push to this branch so the `build_debug_android` passes.  Sorry for the hassle!

juancruzgs:
@octohub We were not doing that logic before, we were only saving the last zip file if there were more than 20 files. It downloaded everything again if we have less than 20 files.

I can implement it here, should be straightforward.
The zip file won't change its content after it's added to the list, right?

octohub:
Tested and works as expected

PR #117, opened 2020-07-31
Zf/improve chronic condition selection

zhaccc:
#### Description:

This PR removes `None` selection from the chronic conditions question and life-threatening symptoms question.
`None` is implicitly selected if `Next` button is pressed without any selections.

This still needs a design for skip button.


src/SelfAssessment/AssessmentQuestion.js
```diff
@@ -116,6 +117,16 @@ export const AssessmentQuestion = ({ onNext, onChange, option, question }) => {
     onChange(selectedValues)
   }, [selectedValues, onChange])
 
+  const handleOnNextPress = () => {
+    // if nothing is selected on multi question select none value from the survey json
```
johnschoeman, line 121: what do you think about removing this comment?

src/SelfAssessment/AssessmentQuestion.js
```diff
@@ -63,28 +63,29 @@ export const AssessmentQuestion = ({ onNext, onChange, option, question }) => {
...
-        answer={selectedValues.find((v) => v.index === index)}
-        index={index}
-        key={option.value}
-        onSelect={(value) => onSelectHandler(value, index)}
-        option={option}
-        isSelected={selectedValues.some((v) => v.index === index)}
-        optionType={question.screen_type}
-      />
-    ))
+    option.values.reduce((result, option, index) => {
```
aledustet, line 66: How do you feel about extracting this bit into a method, perhaps in this same file but outside the view logic? Reading it inside this context feels a bit more complex than what it is IMO.

src/SelfAssessment/AssessmentQuestion.js
aledustet: Perhaps if it is very necessary, maybe we can extract the body of the clause into a function and we can name it something similar to the comment?

johnschoeman:
Closing due to lack of activity

zhaccc:
> Closing due to lack of activity

I'm back from vacation so I am going to reopen and address this PR now. :)

zhaccc:
> Code looks good! A bit confused on the highlevel flow though, if we want to select none implicitly, why do we have this logic that removes it from the options list at run time? could we not have it in the selection list to begin with? and if the user doesnt have a selection on press next, so `null`, then set the value on the survey to none?
> 
> Perhaps thats a bigger refactor though so makes sense to ship this and then come back and clean it up?

I thought about this and that would have been for sure an easier solution to this. But the current flow in `survey.en.json` has condition based on the answer selected so I thought it would be a good idea to do it like this for now. What do you think about that?

src/SelfAssessment/AssessmentQuestion.js
zhaccc: I think since this is a small block of code there is no need to extract it into a different function. I will remove the comment, it's unnecessary.

zhaccc:
> I'm lacking some context on why the update but it looks like it's doing what the description of the PR says. Do we have a ticket for the skip button on Trello? Should we create one?

It seems there already is a ticket on trello for this:
https://trello.com/c/W0zYRFmm/80-as-a-user-i-want-to-be-able-to-skip-questions-during-self-assessment-that-dont-apply-to-me-or-i-am-not-comfortable-answering

src/SelfAssessment/AssessmentQuestion.js
```diff
@@ -129,14 +114,26 @@ export const AssessmentQuestion = ({ onNext, onChange, option, question }) => {
...
       <View style={style.footer}>
         <Button
-          disabled={!selectedValues.length}
-          onPress={onNext}
+          onPress={handleOnNextPress}
           label={t("assessment.next")}
+          disabled={
+            !(
+              question.question_type === QUESTION_TYPE_MULTI ||
+              selectedValues.length
```
aledustet, line 134: this `!selectedValues.length` seems to be repeated a couple of times across, nothing wrong with it, I was just wondering if there might be a concept behind it that could be better expressed as a variable?

src/SelfAssessment/AssessmentQuestion.js
```diff
@@ -145,6 +142,33 @@ export const AssessmentQuestion = ({ onNext, onChange, option, question }) => {
...
+          option={option}
+          isSelected={selectedValues.some((v) => v.index === index)}
+          optionType={question.screen_type}
+        />,
+      )
+    }
+    return result
+  }, [])
+
+  return displayAsOption && mapAssessmentOptions
```
aledustet, line 169: How do you feel about returning an explicit null here? I feel like is a bit easier to read that we won't render the assessment option if it shouldn't be rendered as an option. 
Does it work if we short-circuit the method to return null as soon as we know that we shouldn't be rendering the options at all?

src/SelfAssessment/EmergencyAssessment.js
```diff
@@ -64,9 +64,9 @@ const ChoiceButtons = ({ agreePress, disagreePress }) => {
         accessibilityRole="button"
         style={style.button}
       >
-        <GlobalText style={{ ...style.buttonText, ...style.boldText }}>
+        <GlobalText style={[style.buttonText, style.boldText]}>
```
aledustet, line 67: Just curious if this change was due a pattern we are using or is it better somehow?

src/SelfAssessment/AssessmentQuestion.js
```diff
@@ -50,41 +50,17 @@ export const AssessmentQuestion = ({ onNext, onChange, option, question }) => {
...
   const onSelectHandler = (value, index) => {
     if (question.question_type === QUESTION_TYPE_MULTI) {
       return setSelectedValues((values) => {
         // TODO: Better way to filter single value questions?
-        const singleValueQuestions = [
-          "Choose not to answer",
-          "None of the above",
-          "None",
-        ]
+        const singleValueQuestions = ["Choose not to answer", "None"]
```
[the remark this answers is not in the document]
zhaccc: This constant is only used for this particular function. It is used to figure out single value questions in multi-value question type.

src/SelfAssessment/AssessmentQuestion.js
zhaccc: Hmm... We can make this more explicit but I think there is no need. This is a known conditional rendering React pattern so I think it is safe to stick with it.

src/SelfAssessment/AssessmentQuestion.js
zhaccc: Good point, but since we repeat this only two times in the whole component lets just leave it as is for now.

PR #118, opened 2020-07-31
Update readme

octohub:
#### Description:
Recommend `bin/set_ha.sh` instead of `bin/fetch_ha_env` because `bin/set_ha.sh` already handles fetching the environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Populate the following 2 `.env` files with the relevant urls for your GAEN serve
 .env.bt # variables used in building binaries
 ```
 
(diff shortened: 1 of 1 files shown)


juancruzgs:
`set_ha.sh`  will also change the appId in `build.gradle`.
How do we make sure that the change it's not committed to the repository?
I'd only use `set_ha` to distribute builds, I'm not sure if it's needed for debugging purposes

aledustet:
Is there something extra to set up the repo that needs to happen from the `fetch_ha_env` script?

johnschoeman:
> `set_ha.sh` will also change the appId in `build.gradle`.
> How do we make sure that the change it's not committed to the repository?

We run into the same issue when building ios. its a bit annoying, but we just remember to not commit those changes.  And if they were to be committed by accident, it wouldnt be a huge deal. So not sure we need to make the extra effort to make sure this never happens. 

And once we get fast lane setup properly, ideally this will become a non-issue as we wont be building locally anymore.

PR #119, opened 2020-07-31
Remove bt flavor

juancruzgs:
#### Description:

We don't need flavors anymore since we moved GPS to a different repository.
This PR removes the `bt` flavor and its folders. Files from that flavor have been merged to `main`

#### Linked issues:

https://trello.com/c/c7kHN5hs/221-as-a-dev-team-we-dont-have-a-single-bt-flavor

#### Notes:
I changed the application ID from `org.pathcheck.covidsafepathsBt` to `org.pathcheck.covidsafepaths.bt` because some plugins don't support uppercase letters (Firebase for example).
I will only impact debug builds (release builds take the appId from .env files) and you will have to uninstall the previous app. It might delete your previous keys.


juancruzgs:
Build was failing because Gradle 4.0.0 doesn’t allow local aar dependencies.
I have updated `react-native-background-fetch` dependency to fix that https://github.com/transistorsoft/react-native-background-fetch/issues/276.

juancruzgs:
You are right, thanks @aledustet !

aledustet:
Solid work BTW this was MUCH needed 💪

PR #127, opened 2020-08-03
Fetch Exposure Configuration from Server (iOS)

mattThousand:
### Why
We'd like to be able to set the exposure configuration at runtime instead of hardcoding the value

### This Commit
This commit updates `APIClient` to accept a request to fetch the exposure configuration from the server, and executes that request on `init` (`ExposureManager`) and `willEnterForeground`

### DIFF
--- a/ios/AppDelegate.m
+++ b/ios/AppDelegate.m
@@ -48,7 +48,6 @@
   // Schedule background task
   [[ExposureManager shared] scheduleBackgroundTaskIfNeeded];
 
-  // Broadcast EN Status and fetch exposure configuration
   [[ExposureManager shared] setup];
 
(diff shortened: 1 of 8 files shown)


ios/AppDelegate.m
```diff
@@ -48,16 +48,17 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   // Schedule background task
   [[ExposureManager shared] scheduleBackgroundTaskIfNeeded];
 
-  // Broadcase EN Status
-  [[ExposureManager shared] broadcastCurrentEnabledStatus];
+  // Broadcast EN Status and fetch exposure configuration
```
emilianojuan, line 51: maybe this comment will be better in the method signature

PR #129, opened 2020-08-03
Get privacy policy from environment

devinjameson:


src/More/Licenses.tsx
```diff
@@ -22,7 +23,7 @@ const Licenses: FunctionComponent = () => {
   const infoEmailLink = "mailto:info@pathcheck.org"
   const pathCheckWebAddress = "pathcheck.org"
   const pathCheckUrl = "https://pathcheck.org/"
-  const privacyPolicyUrl = "https://pathcheck.org/privacy-policy/"
+  const privacyPolicyUrl = env.PRIVACY_POLICY_URL
```
aledustet, line 26: Should we have a fallback value for it? Or perhaps not show the link if no privacy policy URL is present? It sucks to do but the data allows for it to happen 😨

devinjameson: Yeah good call.

PR #130, opened 2020-08-03
(Android) Distribute MN builds via Bitrise & Fastlane

juancruzgs:
#### Description:

Add a new workflow to automatically distribute Android builds on each commit to develop.

**Releases:**
https://appcenter.ms/orgs/PathCheck/apps/packcheck-bt-mn/distribute/releases/

**Passing builds:**
https://github.com/Path-Check/gaen-mobile/runs/941610921?check_suite_focus=true

### DIFF
--- a/.github/workflows/distribute.yml
+++ b/.github/workflows/distribute.yml
@@ -1,0 +1,111 @@
+name: Distribute
+
+on:
+  push:
+    branches: [develop]
+
+jobs:
+  distribute_android:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Cache node_modules/
+        uses: actions/cache@v1
+        id: node_cache
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-yarn-cache-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.OS }}-yarn-cache-
+
+      - name: Install Node Dependencies
+        if: steps.node_cache.outputs.cache-hit != 'true'
+        run: yarn --frozen-lockfile
+
+      - name: Setup kernel for react native, increase watchers for release versions.
+        run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
+      - name: Cache gradle
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Set up Ruby 2.7
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.7
+
+      - name: Install bundler for ruby package management
+        run: gem install bundler -v "$(grep -A 1 "BUNDLED WITH" ./Gemfile.lock | tail -n 1)"
+        working-directory: ./android
+
+      - name: Install dotenv
+        run: gem install dotenv
+        working-directory: ./android
+        
+      - name: Cache ruby gems
+        uses: actions/cache@v1
+        id: ruby_cache
+        with:
+          path: android/vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/android/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
+
+      - name: Install Ruby Dependencies
+        if: steps.ruby_cache.outputs.cache-hit != 'true'
+        run: |
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+        working-directory: ./android
+        
+      ## Set up MN
+      - name: Setup MN env
+        run: ./bin/set_ha.sh mn ${{ secrets.ACCESS_TOKEN }}
+
+      - name: Save signing keys
+        run: ./bin/save_keystore.sh ${{ secrets.KEYSTORE_MN }}
+
+      - name: Increment version code
+        uses: maierj/fastlane-action@v1.4.0
+        with:
+          lane: 'release_increment_version_code'
+          subdirectory: 'android'
+          bundle-install-path: 'vendor/bundle'
+
+      - name: Build android release APK
+        uses: maierj/fastlane-action@v1.4.0
+        with:
+          lane: 'release_apk'
+          subdirectory: 'android'
+          bundle-install-path: 'vendor/bundle'
+          options: '{ "keyStorePassword" : "${{ secrets.KEYSTORE_PASSWORD_MN }}", "keyAlias" : "${{ secrets.KEY_ALIAS }}", "keyPassword" : "${{ secrets.KEY_PASSWORD_MN }}"}'
+
+      - name: Distribute APK
+        uses: maierj/fastlane-action@v1.4.0
+        with:
+          lane: 'android_alpha_apk'
+          subdirectory: 'android'
+          bundle-install-path: 'vendor/bundle'
+          options: '{ "apiToken": "${{ secrets.APPCENTER_TOKEN }}", "appName" : "packcheck-bt-mn" }'
+
+      # - name: Build android release AAB
+      #   uses: maierj/fastlane-action@v1.4.0
+      #   with:
+      #     lane: 'release_aab'
+      #     subdirectory: 'android'
+      #     bundle-install-path: 'vendor/bundle'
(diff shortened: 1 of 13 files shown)


.github/workflows/distribute.yml
```diff
@@ -0,0 +1,111 @@
...
+        with:
+          ruby-version: 2.7
+
+      - name: Install bundler for ruby package management
+        run: gem install bundler -v "$(grep -A 1 "BUNDLED WITH" ./Gemfile.lock | tail -n 1)"
+        working-directory: ./android
+
+      - name: Install dotenv
+        run: gem install dotenv
+        working-directory: ./android
```
juancruzgs, line 49: I had to add this step to be able to run `set_ha.sh`

.github/workflows/distribute.yml
```diff
@@ -0,0 +1,111 @@
...
+          bundle-install-path: 'vendor/bundle'
+          options: '{ "keyStorePassword" : "${{ secrets.KEYSTORE_PASSWORD_MN }}", "keyAlias" : "${{ secrets.KEY_ALIAS }}", "keyPassword" : "${{ secrets.KEY_PASSWORD_MN }}"}'
+
+      - name: Distribute APK
+        uses: maierj/fastlane-action@v1.4.0
+        with:
+          lane: 'android_alpha_apk'
+          subdirectory: 'android'
+          bundle-install-path: 'vendor/bundle'
+          options: '{ "apiToken": "${{ secrets.APPCENTER_TOKEN }}", "appName" : "packcheck-bt-mn" }'
```
juancruzgs, line 95: This section can be repeated to create APKs for different HAs
Github Action doesn't provide a way to share it yet https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851

.github/workflows/distribute.yml
```diff
@@ -0,0 +1,111 @@
...
+      #     bundle-install-path: 'vendor/bundle'
+      #     options: '{ "keyStorePassword" : "${{ secrets.KEYSTORE_PASSWORD_MN }}", "keyAlias" : "${{ secrets.KEY_ALIAS }}", "keyPassword" : "${{ secrets.KEY_PASSWORD_MN }}"}'          
+
+      # - name: Distribute AAB
+      #   uses: maierj/fastlane-action@v1.4.0
+      #   with:
+      #     lane: 'android_alpha_aab'
+      #     subdirectory: 'android'
+      #     bundle-install-path: 'vendor/bundle'
+      #     options: '{ "apiToken": "${{ secrets.APPCENTER_TOKEN }}", "appName" : "packcheck-bt-mn" }'
```
juancruzgs, line 111: We can't upload AAB files to AppCenter https://github.com/microsoft/appcenter/issues/955.
We should discuss which platform we can use to host those AAB files (they are required by Play Store).

Some options are:
- Google Drive (we need to set up https://github.com/google/skicka)
- Slack (someone with admin permissions will have to set up a bot and store the token as Github Secret)
- Github Actions Artifacts (those artifacts are public, maybe we can set up a hook that stores them in a private repository)

.github/workflows/distribute.yml
```diff
@@ -0,0 +1,111 @@
...
+      - name: Install Ruby Dependencies
+        if: steps.ruby_cache.outputs.cache-hit != 'true'
+        run: |
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+        working-directory: ./android
+        
+      ## Set up MN
+      - name: Setup MN env
+        run: ./bin/set_ha.sh mn ${{ secrets.ACCESS_TOKEN }}
```
juancruzgs, line 69: Added an optional parameter to send a token to this script

juancruzgs:
If this code looks good I'll add the configuration to distribute the rest of the HA builds

.github/workflows/distribute.yml
juancruzgs: We plan to move the distribution logic to Bitrise and store private artifacts there
https://trello.com/c/4jAoEC7f/276-set-up-a-ci-deployment-pipeline

bin/fetch_ha_env.sh
```diff
@@ -21,9 +21,10 @@ require 'dotenv'
 Dotenv.load
 
 HA_LABEL = ARGV[0]
+ACCESS_TOKEN = ARGV[1] # Optional parameter used by CI
 
 def fetch_env
-  token = ENV.fetch("ACCESS_TOKEN")
+  token = ACCESS_TOKEN.nil? ? ENV.fetch("ACCESS_TOKEN") : ACCESS_TOKEN
```
aledustet, line 27:
```suggestion
  token = ACCESS_TOKEN || ENV.fetch("ACCESS_TOKEN")
```
if access token has a value is `truthy` so it will short circuit, totally not important just wanted to share the option 😄 

You could also do:

```
ACCESS_TOKEN ||= ENV.fetch("ACCESS_TOKEN")
```

And then use `ACCESS_TOKEN` from that point forward 🤷

.github/workflows/distribute.yml
aledustet: Super left-field question here, but is there a world(I hope not) where a particular build should only be distributed for a particular set of `HA`?

.github/workflows/distribute.yml
```diff
@@ -0,0 +1,111 @@
...
+          subdirectory: 'android'
+          bundle-install-path: 'vendor/bundle'
+
+      - name: Build android release APK
+        uses: maierj/fastlane-action@v1.4.0
+        with:
+          lane: 'release_apk'
+          subdirectory: 'android'
+          bundle-install-path: 'vendor/bundle'
+          options: '{ "keyStorePassword" : "${{ secrets.KEYSTORE_PASSWORD_MN }}", "keyAlias" : "${{ secrets.KEY_ALIAS }}", "keyPassword" : "${{ secrets.KEY_PASSWORD_MN }}"}'
```
aledustet, line 87: this will make it impossible for persons outside the organization to trigger when they merge code to develop. It is probably the intended behavior so I'm not worried about it, it is the same with the lokalise push script, just wanted to make sure we make sure that if an external contribution needs to cut a build we would have to merge something else after or, do it ourselves.

aledustet:
I also think that this tackles at least art of https://trello.com/c/H2ZtptDC 🎉  would you mind checking those boxes @juancruzgs ?

.github/workflows/distribute.yml
juancruzgs: We would have to add different git branches with specific configurations in that case I think

bin/fetch_ha_env.sh
juancruzgs: Excellent!
I was looking for something like that, but my knowledge in Ruby is 0

.github/workflows/distribute.yml
```diff
@@ -0,0 +1,111 @@
+name: Distribute
+
+on:
+  push:
+    branches: [develop]
```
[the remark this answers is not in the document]
juancruzgs: I like the idea of having builds for each commit that the QA team can test (if they want to). In that way they can test specific commits and find regressions easily.
We could set up a `master` branch that reflects what is already deployed to the Play Store (and it can automatically make the deploys in the future).

This won't cause PRs to fail because it only runs when something is merged to develop, and external contributors are not allowed to push directly to develop.

That being said, I'm open to changing the workflow if this doesn't work or you don't feel comfortable with it

bin/set_ha.sh
```diff
@@ -19,7 +19,8 @@
 require 'open3'
 
 HA_LABEL = ARGV[0]
+ACCESS_TOKEN = ARGV[1] # Optional parameter used by CI
```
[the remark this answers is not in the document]
juancruzgs: Makes sense! I'll change it

juancruzgs:
Distribution workflow migrated to [Bitrise](https://app.bitrise.io/app/6da8ef1aa79776ae/workflow_editor#!/workflows?workflow_id=deploy).
Pull request checks will still run in Github Actions, no changes made there.

This PR uses **fastlane** to create APKs and upload them to AppCenter.
I wanted to make the process agnostic to the CI tool we use, we just have to call the same fastlane commands if we switch to a different CI tool.

On each merge to develop we do:
- Set environment variables
- Increment version code (Build number)
- Create release APK
- Upload it to AppCenter and Bitrise Artifacts
- Create release AAB
- Upload it to Bitirse Artifacts

I had to change some versions to match the versions installed in Bitrise's virtual machine.
- **Ruby:** changed from 2.7 to 2.6.5. https://github.com/bitrise-io/bitrise.io/blob/master/system_reports/osx-xcode-12.0.x.log#L727
- **NDK:** Changed from 21.3.6528147 to 21.0.6113669. https://github.com/bitrise-io/bitrise.io/blob/master/system_reports/osx-xcode-12.0.x.log#L1220

.github/workflows/distribute.yml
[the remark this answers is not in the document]
johnschoeman: that makes sense to me! yeah having a build for every commit is helpful for QA when tracking down issues, so we should do that, especially if there are no trade offs to be made.

PR #132, opened 2020-08-03
Remove unused dependencies

juancruzgs:
#### Description:

There are some dependencies that we are only being used by GPS.
It's important to remove them because they are requesting WRITE_EXTERNAL_STORAGE permission and we cannot submit the app to the Play Store until we remove it.

Dependencies:
- react-native-fs: Added here  https://github.com/Path-Check/covid-safe-paths/pull/34/files
- react-native-zip-archive
- rn-fetch-blob

#### Linked issues:

https://trello.com/c/6LIDHYdQ/274-android-permissions


octohub:
Using the [APK Info](https://play.google.com/store/apps/details?id=com.themelisx.apkinfo&hl=en_US) app I can see that the `android.permission.WRITE_EXTERNAL_STORAGE` is still being requested even after removing this dependency. Can you look into whether something else is requesting this permission? 

Screenshot_20200803-162105

aledustet:
Could we be missing a pod install here?

juancruzgs:
You are right @octohub, I found there's another library called `rn-fetch-blob` that also requests WRITE_EXTERNAL_STORAGE permission. When you open the AndroidManifest there is a tab called `Merged Manifest` where you can see all the permissions the app requests.

@aledustet Do you know if it's safe to remove that dependency?

aledustet:
@juancruzgs that on is used to access the file system which strikes me as a reminiscence from the export days, but not sure if any of the debug screen functionality supports for exporting any dump of data to the user? If tha t's not the case I would say we can probably get rid of it

octohub:
Hey @aledustet, @juancruzgs and I spoke to @JacobJaffe, this is good to remove!

PR #133, opened 2020-08-03
Fix language list

emilianojuan:
When trying to choose a language in iOS devices with a top notch the list overlaps with it, I added a SafeAreaView there to prevent this issue

issue:
Screen Shot 2020-08-03 at 16 14 53
fix:
Screen Shot 2020-08-03 at 16 13 55

### DIFF
--- a/src/More/LanguageSelection.tsx
+++ b/src/More/LanguageSelection.tsx
@@ -1,5 +1,11 @@
 import React from "react"
-import { FlatList, View, StyleSheet, TouchableHighlight, SafeAreaView } from "react-native"
+import {
+  FlatList,
+  View,
+  StyleSheet,
+  TouchableHighlight,
+  SafeAreaView,
+} from "react-native"
 import { useTranslation } from "react-i18next"
 import { useNavigation } from "@react-navigation/native"
 import { useStatusBarEffect } from "../navigation"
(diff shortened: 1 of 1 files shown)


src/More/LanguageSelection.tsx
```diff
@@ -2,11 +2,13 @@ import React from "react"
 import { FlatList, View, StyleSheet, TouchableHighlight } from "react-native"
 import { useTranslation } from "react-i18next"
 import { useNavigation } from "@react-navigation/native"
+import { useStatusBarEffect } from "../navigation"
 
 import { getLocaleList, setUserLocaleOverride } from "../locales/languages"
 import { GlobalText } from "../components/GlobalText"
 
 import { Colors, Spacing, Typography } from "../styles"
+import { SafeAreaView } from "react-native-safe-area-context"
```
johnschoeman, line 11: what do you think about keeping this dependency at the top with the other third party dependencies?

devinjameson: This can actually come directly from "react-native" @emilianojuan

src/More/LanguageSelection.tsx
```diff
@@ -2,11 +2,13 @@ import React from "react"
 import { FlatList, View, StyleSheet, TouchableHighlight } from "react-native"
```
devinjameson, line 2:
```suggestion
import { FlatList, View, StyleSheet, TouchableHighlight, SafeAreaView } from "react-native"
```

src/More/LanguageSelection.tsx
devinjameson, line 11:
```suggestion
```

PR #134, opened 2020-08-03
Remove react-native-push-notification

juancruzgs:
#### Description:

This library adds a notification channel that is shown to the user but seems to be unused.
I haven't seen any references in the code, are we using it to get notifications from Firebase?


juancruzgs:
GPS uses it to trigger notifications from the RN side, so maybe it was added by them.
If we need it in the future we can add a native module

PR #139, opened 2020-08-04
Add form for users to report an issue

aledustet:
Why:
----
In the event the users would like to communicate some feedback to the team, there isn't currently a place to do so inside the application. We need a way to create tickets in our customer support portal.

How:
----
The zendesk API has a way to create anonymous requests. We have the endpoint and the account and need to send a request with the required subject, and descriptions of the problem. Also, the OS info and the version of the app will be included

This PR:
----
- Add zendeskAPI helper to submit feedback
- Add Report Issue form to the more stack
- Add extra custom fields to the feedback request
- Add validations to Feedback form for required fields
- Add tests for the submit and error path of the feedback form
- Navigate back when feedback is submitted
- Switch input of subject for input of email
- Send the application name as an argument of the request

Reviewers:
----
- The app version here is hardcoded since we are waiting to remove the library we were using for getting this info and didn't seem prudent to add yet another dependency.

[trello](https://trello.com/c/9irPiTYI)

Large GIF (436x914)

Co-authored-by: Thomas G Paresi <tparesi@gmail.com>


src/More/FeedbackForm.tsx
```diff
@@ -0,0 +1,268 @@
...
+  const [isDisabled, setIsDisabled] = useState(true)
+  const [errorMessage, setErrorMessage] = useState(defaultErrorMessage)
+
+  const validate = () => {
+    const hasSubject = subject.trim().length > 0
+    const hasBody = body.trim().length > 0
+
+    if (hasSubject && hasBody) {
+      setIsDisabled(false)
+    } else setIsDisabled(true)
```
johnschoeman, line 41:
```suggestion
    } else {
    setIsDisabled(true)
    }
```

src/More/FeedbackForm.tsx
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  useEffect(validate, [subject, body])
```
johnschoeman, line 44: 💯

src/More/zendeskAPI.ts
```diff
@@ -0,0 +1,75 @@
+import env from "react-native-config"
+
+const defaultHeaders = {
+  "content-type": "application/json",
+}
+
+type EnvironmentProps = {
```
johnschoeman, line 7: what do you think about using an interface here?

src/More/zendeskAPI.ts
```diff
@@ -0,0 +1,75 @@
...
+interface NetworkFailure<U> {
+  kind: "failure"
+  error: U
+}
+
+export type FeedbackError = "Unknown" | "ZendeskError"
+
+export type NetworkResponse<U = "Unknown"> = NetworkSuccess | NetworkFailure<U>
+
+const OS_FIELD_KEY = "360033622032"
```
johnschoeman, line 33: Are these zendesk specific things? will they ever change in the future? should they be included in the .env?

src/More/zendeskAPI.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+interface NetworkFailure<U> {
+  kind: "failure"
+  error: U
+}
+
+export type FeedbackError = "Unknown" | "ZendeskError"
+
+export type NetworkResponse<U = "Unknown"> = NetworkSuccess | NetworkFailure<U>
```
johnschoeman, line 31:
```suggestion
export type NetworkResponse<Error = "Unknown"> = NetworkSuccess | NetworkFailure<Error>
```
or perhaps `T`,  the `U` jumps out as and artifact of a refactor that wasnt fully finished.

src/More/zendeskAPI.ts
aledustet: I don't know if they will change in the future, but if they do we will most likely have to make changes to the implementation. Because if the concept of the field changes I would think that the value should be different also.

src/More/zendeskAPI.ts
aledustet: Not sure why? I'm not opposed but since the most notable difference from an interface and a type(that I know of) is that re-declarations result in merging I try to use types for final declarations. But definitely up for it.

src/More/ReportIssueForm.tsx
```diff
@@ -0,0 +1,255 @@
...
+import { useNavigation } from "@react-navigation/native"
+
+import { GlobalText } from "../components/GlobalText"
+import { Button } from "../components/Button"
+import { reportAnIssue, ReportIssueError } from "./zendeskAPI"
+
+import { Spacing, Layout, Forms, Colors, Outlines, Typography } from "../styles"
+
+const defaultErrorMessage = " "
+const EMAIL_REGEX = RegExp(/\S+@\S+\.\S+/)
```
johnschoeman, line 23: should we be rolling our own email validation? is this something the zendesk api could handle?

src/More/zendeskAPI.ts
johnschoeman: cool, thanks for the reasoning. I also dont have a strong preference.

src/More/ReportIssueForm.tsx
aledustet: It does, actually, we could potentially gather the feedback from the errors and then display it back, not sure about the localization aspect of it. Maybe is worth exploring that.

aledustet: @johnschoeman updated that logic

.github/workflows/test.yml
```diff
@@ -52,10 +52,10 @@ jobs:
         if: steps.cache.outputs.cache-hit != 'true'
         run: yarn --frozen-lockfile
 
-      - name: Set up Ruby 2.6
+      - name: Set up Ruby 2.7
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.6.5
+          ruby-version: 2.7.1
```
juancruzgs, line 58: This change will break builds in Bitrise, their virtual machines use version `2.6.5`
We can migrate to 2.7.1 but we will have to add an script to install it on Bitrise

juancruzgs: This is script is not being used in Bitrise, but change on the iOS Gemfile will break it I think

aledustet: Then we should update all the versions for it, I'll open a PR with that changes to those lock files, thanks for the heads up I thought we overlooked it

PR #140, opened 2020-08-04
Refactor button component

devinjameson:
Why: our current implementation of the Button component is a bit hard to
understand.

This commit:
- Updates the Button component
- Implements the Button component in the EulaModal screen

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -16,8 +16,8 @@ interface ButtonProps {
   onPress: () => void
   loading?: boolean
   disabled?: boolean
-  buttonStyle?: ViewStyle
-  textStyle?: TextStyle
+  customButtonStyle?: ViewStyle
+  customTextStyle?: TextStyle
   invert?: boolean
   testID?: string
 }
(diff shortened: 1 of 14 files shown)


src/components/Button.tsx
```diff
@@ -27,20 +27,37 @@ export const Button: FunctionComponent<ButtonProps> = ({
...
-  const buttonStyles =
-    disabled || loading
-      ? { ...buttonColorStyle, ...style.buttonDisabled, ...buttonStyle }
-      : { ...buttonColorStyle, ...style.buttonEnabled, ...buttonStyle }
-  const buttonTextStyle =
-    disabled || loading
-      ? { ...style.text, ...style.textDisabled, ...textStyle }
-      : { ...style.text, ...style.textEnabled, ...textStyle }
+  const determineButtonStyle = (): ViewStyle => {
+    if (invert && (disabled || loading)) {
```
johnschoeman, line 36: This is a great improvement! :100:

a next step could be to introduce a union type or maybe a tuple to capture the possible states in the type system, which would provide some clarity in the possible states and help keep impossible state impossible.  something like

```typescript
type ButtonState = Enabled | Disabled | InvertedEnabled | InvertedDisabled
...

const determineButtonState = (invert: boolean, disabled: boolean, loading: boolean): ButtonState => {
  const isEnabled = !(disabled || loading)
  if (isEnabled && !invert) {
     return Enabled
  } else if (...

}

const determineButtonStyle = (state: ButtonState): ViewStyle => {
  switch (state): {
     case Enabled: {
         return style.buttonEnabled
     ...
```

Theres probably a better factoring and this is certainly not a required thing for this pr, just thought you might be interested in the thought.

devinjameson: Yeah this is a great idea! Thanks for sharing. Definitely will implement in a future PR. Yes, I believe we should lean into the `Button` component especially now that we are implementing a gradient button. This will keep the `LinearGradient` craziness scoped to the `Button` component. We also have a ton of button styling just floating around the codebase right now so it feels right to handle all of that within the `Button` component instead of having disabled and enabled states in every stylesheet that might require those states.

PR #141, opened 2020-08-05
Remove 'react-native-device-info' dependency

mattThousand:
### Why
From Google PlayStore review (build 7/30/2020) they note the RNDevice module has the ability to leak a phone number.

### This Commit
This commit removes the `react-native-device-info` dependency and replaces the functionality with native modules for each platform.


src/More/About.spec.tsx
```diff
@@ -1,44 +1,57 @@
...
 import AboutScreen from "./About"
-
-jest.mock("react-native-device-info", () => {
-  return {
-    getApplicationName: jest.fn(),
-    getBuildNumber: jest.fn(),
-    getVersion: jest.fn(),
-  }
-})
+import * as NativeModule from "../gaen/nativeModule"
```
johnschoeman, line 6: would it make sense to introduce another native module here? This doesnt seem related to gaen specifically, and it might be nice to allow them to be treated separately.  so something like
```
import * as DeviceInfo from "./deviceInfoNativeModule"
```

src/More/About.spec.tsx
```diff
@@ -1,44 +1,57 @@
...
-    ;(getApplicationName as jest.Mock).mockReturnValueOnce(applicationName)
+    const applicationNameSpy = jest
+      .spyOn(NativeModule, "getApplicationName")
+      .mockReturnValue(Promise.resolve(applicationName))
 
     const { getByText } = render(<AboutScreen />)
 
-    expect(getByText(applicationName)).toBeDefined()
+    act(() => {
+      jest.advanceTimersByTime(1000)
```
johnschoeman, line 21: this seems a bit odd to need to manually advance the time?  doesnt wait have a 5 second timeout?

mattThousand:
> LGTM, a few thoughts.
> 
> Might makes sense to ping @aledustet for a review on the jest specs, he is our resident jest expert.

Yes, I'd welcome help here. Bit of an inscrutable (for me) test failure having to do with my updates to the `About.spec.tsx`). Failure is reproducible locally by running `yarn test --watch` but not `yarn test` 🤔

src/More/Licenses.tsx
```diff
@@ -8,16 +8,26 @@ import {
...
 
 import { GlobalText } from "../components/GlobalText"
 
+import { getApplicationName } from "../gaen/nativeModule"
 import { Colors, Spacing, Typography } from "../styles"
 import { Icons } from "../assets"
 
 const Licenses: FunctionComponent = () => {
   const { t } = useTranslation()
+  const [applicationName, setApplicationName] = useState("")
```
devinjameson, line 21: Not needed for this PR but I wonder if it would be good to have a `useApplicationName` hook at some point, given that we'll be using the name across the app.

src/More/About.spec.tsx
```diff
@@ -1,44 +1,46 @@
...
 
 describe("About", () => {
-  it("shows the name of the application", () => {
+  it("shows the name of the application", async () => {
     const applicationName = "application name"
 
-    ;(getApplicationName as jest.Mock).mockReturnValueOnce(applicationName)
+    const applicationNameSpy = jest
+      .spyOn(NativeModule, "getApplicationName")
+      .mockReturnValue(Promise.resolve(applicationName))
```
aledustet, line 14:
```suggestion
      .mockResolvedValueOnce(applicationName)
```

src/More/About.spec.tsx
```diff
@@ -1,44 +1,46 @@
...
+  it("shows the build and version number of the application", async () => {
+    const buildNumber = "8"
+    const versionNumber = "0.18"
+    const fullString = `${versionNumber} (${buildNumber})`
 
-    ;(getBuildNumber as jest.Mock).mockReturnValueOnce(buildNumber)
-    ;(getVersion as jest.Mock).mockReturnValueOnce(versionNumber)
+    const buildNumberSpy = jest
+      .spyOn(NativeModule, "getBuildNumber")
+      .mockReturnValue(Promise.resolve(buildNumber))
```
aledustet, line 31:
```suggestion
      .mockResolvedValueOnce(buildNumber)
```

src/More/About.spec.tsx
```diff
@@ -1,44 +1,46 @@
...
+    const versionSpy = jest
+      .spyOn(NativeModule, "getVersion")
+      .mockReturnValue(Promise.resolve(versionNumber))
```
aledustet, line 34:
```suggestion
      .mockResolvedValueOnce(versionNumber)
```

src/More/Licenses.spec.tsx
```diff
@@ -1,30 +1,29 @@
...
 
 describe("LicensesScreen", () => {
-  it("shows the name of the application", () => {
+  it("shows the name of the application", async () => {
     const applicationName = "application name"
 
-    ;(getApplicationName as jest.Mock).mockReturnValueOnce(applicationName)
+    const applicationNameSpy = jest
+      .spyOn(NativeModule, "getApplicationName")
+      .mockReturnValue(Promise.resolve(applicationName))
```
aledustet, line 20:
```suggestion
      .mockResolvedValueOnce(applicationName)
```

yarn.lock
```diff
@@ -4757,7 +4745,6 @@ jest-leak-detector@^25.5.0:
 jest-matcher-utils@^24.0.0:
   version "24.9.0"
   resolved "https://registry.yarnpkg.com/jest-matcher-utils/-/jest-matcher-utils-24.9.0.tgz#f5b3661d5e628dffe6dd65251dfdae0e87c3a073"
-  integrity sha512-OZz2IXsu6eaiMAwe67c1T+5tUAtQyQx27/EMEkbFAGiw52tB9em+uGbzpcgYVpA8wl0hlxKPZxrly4CXU/GjHA==
```
aledustet, line 4760: why are all of these gone here 🤔  is it the yarn version? Or maybe the command to update them?

src/More/About.spec.tsx
mattThousand: the method returns a `Promise<string>` here; it registers an error when passed a `string`

aledustet:
@mattThousand let me know if you want to 🍐  on your specs.

PR #142, opened 2020-08-05
Update button component styling

devinjameson:
Why: our button styling is changing based on new visual design
requirements.

This commit:
- Updates our button component styling
- Uses our button component in several more locations in place of
  TouchableOpacity

Screen Shot 2020-08-05 at 12 51 53 PM
Screen Shot 2020-08-05 at 12 51 59 PM
Screen Shot 2020-08-05 at 12 52 06 PM

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -6,10 +6,11 @@ import {
   TextStyle,
   TouchableOpacity,
 } from "react-native"
+import LinearGradient from "react-native-linear-gradient"
 
 import { GlobalText } from "./GlobalText"
 
-import { Buttons, Typography } from "../styles"
+import { Colors, Buttons, Typography } from "../styles"
 
 interface ButtonProps {
   label: string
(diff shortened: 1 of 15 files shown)


src/components/Button.tsx
```diff
@@ -32,15 +33,15 @@ export const Button: FunctionComponent<ButtonProps> = ({
   invert,
   testID,
 }) => {
-  const determineButtonStyle = (): ViewStyle => {
+  const determineGradient = (): [string, string] => {
     if (invert && (disabled || loading)) {
-      return style.buttonInvertedDisabled
+      return [Colors.mediumGray, Colors.lighterGray]
```
aledustet, line 38: Would it make sense to extract these tuples as constants? Probably an overkill but there's at least a couple repeated that might be tricky to realize if more changes are made in the future

devinjameson: Good call! Will handle in my next PR.

PR #144, opened 2020-08-05
Implement revamped onboarding flow

devinjameson:
Why: our onboarding flow has changed.

This commit:
- Implements most of the revamped onboarding flow, with the exception of the EulaModal screen
- Updates various app-wide styles

onboarding
onboarding2


src/locales/en.json
```diff
@@ -125,34 +127,41 @@
...
-    "launch_screen2_subheader_bluetooth": "Nearby phones use Bluetooth to share and save random number strings to recall devices nearby.\n\nIf a person reports a positive diagnosis, their phone's recent random IDs are flagged to anonymously notify others. No locations or personal info is shared.",
     "launch_screen3_header_bluetooth": "Receive notifications when someone you encountered later chooses to report a positive diagnosis.",
     "launch_screen3_subheader_bluetooth": "Learn details about the encounter, such as the date, duration, and Bluetooth signal strength.",
     "launch_screen4_header_bluetooth": "If you test positive for the virus that causes COVID-19, you can choose to share your diagnosis anonymously.",
     "launch_screen4_subheader_bluetooth": "This helps others in your community contain the spread of the virus.",
     "launch_set_up_phone_bluetooth": "Next",
     "legal_page_address": "PathCheck Foundation\n58 Day Street\nBox 441621\nSomerville, MA 02144",
     "more_icon": "Icon of three horizontal dots",
+    "onboarding_screen1_header": "Together, we can contain the spread of COVID-19",
```
devinjameson, line 150: These should probably be moved to the `onboarding` section.

src/navigation/OnboardingStack.tsx
```diff
@@ -1,62 +1,154 @@
...
+      />
+      <Stack.Screen
+        name={OnboardingScreens.GetNotified}
+        component={GetNotified}
+      />
+      <Stack.Screen
+        name={OnboardingScreens.ValueProposition}
+        component={ValueProposition}
+      />
+      <Stack.Screen
```
devinjameson, line 113: Curious of anyone's thoughts re: these two screens, and eventually the `EulaModal` as well, living in their own "App Setup" stack.

src/locales/en.json
johnschoeman: there is an automated script to handle some of this moving around, let me know if you want to pair on it.

src/navigation/OnboardingStack.tsx
johnschoeman: That would make sense to me. They are pretty distinct functionally from the onboarding info screens, so I would lean towards them being grouped separately.

src/Onboarding/ExplanationScreen.tsx
```diff
@@ -1,197 +1,150 @@
...
-  const determineIconStyle = (iconStyle: IconStyle): ViewStyle => {
-    switch (iconStyle) {
-      case IconStyle.Blue:
-        return { ...style.icon, ...style.blueIcon }
-      case IconStyle.Gold:
-        return { ...style.icon, ...style.goldIcon }
+  const { t } = useTranslation()
+  const navigation = useNavigation()
+
+  const determineCircleStyle = (circlePosition: number): ViewStyle => {
```
johnschoeman, line 45: This strikes me as coupling the data model and a the view layer a bit. what do you think of an api a bit more like:

```typescript
const PositionDots = (currentPosition: number, positionCount: number): FunctionComponent<P> => {
  const determineCircleStyle = (idx: number, position: number) => {
        if idx === position {
            return style.circleActive
        } else {
           return style.circleInactive
        }
   }

   return <View style={style.circles}>
         { [1, ... , positionCount].map((idx) => {
               <View style={determineCircleStyle(idx, currentPosition)} />
           }
          }
    </View>
```

Something like this allows the view component for the bottom dots to not have any knowledge of the dataModel, `explanationScreenContent.screenNumber` and both can change independently.

src/Onboarding/ActivateProximityTracing.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+import { Spacing, Typography, Buttons, Colors } from "../styles"
+
+const ActivateProximityTracing: FunctionComponent = () => {
+  const { t } = useTranslation()
+  const navigation = useNavigation()
+
+  const { exposureNotifications } = usePermissionsContext()
+
+  const headerText = t("onboarding.proximity_tracing_header")
```
aledustet, line 25: Super small comment here but how do you feel about naming the variables closer to the translation keys? It makes the relationship for the values a bit better to read. Maybe just inlining the translations all together?

src/Onboarding/ExplanationScreen.tsx
aledustet: Solid feedback here, I also think that this will make it easier to extend in the future and separates the logic part nicely.

src/navigation/OnboardingStack.tsx
```diff
@@ -1,62 +1,154 @@
...
-)
+const OnboardingStack: FunctionComponent = () => {
+  const { t } = useTranslation()
+  const navigation = useNavigation()
+
+  const onboardingScreenOptions: StackNavigationOptions = {
+    headerShown: false,
+  }
+
+  const HeaderRight = (stepNumber: number) => {
```
aledustet, line 43: How do we feel about creating an enum or a union type to hold the screens? It feels like a stretch to have indexes since we are hardcoding the values in it. Could be something like:
``` typescript
enum OnboardingStep = {
  ValuePropositionStep = 1
  ...
}
type OnboardingStep = "ValuePropositionStep" | ...
```
This way you could tie a number with a meaning and we're not passing only cold numbers around, this is probably not a critical improvement but it will give you some safety here that we won't miss any screen if one gets added. Happy to talk more about it. You could probably even tie it to a subset of the screen names that will  be on the navigator, in case an update happens on the type system only and then we will get a warning here that we have an incomplete resolver.

src/Onboarding/ExplanationScreen.tsx
devinjameson: Thanks for the thoughtful feedback! Thoughts on faad0fda98b793690d80389bae588320b705130d?

src/Onboarding/ActivateProximityTracing.tsx
devinjameson: 💯

src/navigation/OnboardingStack.tsx
devinjameson: Great call thoughts on ac0fedbc50f023157dca92e24e8cb4b556a7fa3f?

src/navigation/OnboardingStack.tsx
devinjameson: Could I get your thoughts on 288893b105d9c4e998dc1d40acdcba9ffeba6db4?

src/navigation/OnboardingStack.tsx
devinjameson: Actually @aledustet I think this implements what you suggested in case you're interested.

devinjameson: Sorted it out thanks!

PR #145, opened 2020-08-06
Remove rn device info dependency

aledustet:
Why:
----
The `react-native-device-info` package has security vulnerabilities potentially leaking phone numbers on the android build. We need to display the info provided by this package without the security vulnerabilities.

How:
----
Native modules are added on the android and ios side exposing the build number, version number and the package name to display to users.

This PR:
----
- Remove RNDeviceInfo dependency.
- Implement device info native module (iOS)
- Implement device info native module (Android)
- Add hook to consume application name and version info

Screen Shot 2020-08-06 at 12 09 55 PM

### DIFF
--- a/src/More/useApplicationInfo.ts
+++ b/src/More/useApplicationInfo.ts
@@ -1,0 +1,52 @@
+import { useEffect, useState } from "react"
+
+import {
+  getVersion,
+  getBuildNumber,
+  getApplicationName,
+} from "../gaen/nativeModule"
+
+interface ApplicationInfo {
+  applicationName: string
+  versionInfo: string
+}
+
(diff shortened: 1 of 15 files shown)


src/More/useApplicationInfo.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const useApplicationInfo = (): ApplicationInfo => {
+  const { applicationName } = useApplicationName()
+  const { versionInfo } = useVersionInfo()
+
+  return { applicationName, versionInfo }
+}
+
+export { useApplicationInfo, useApplicationName, useVersionInfo }
```
devinjameson, line 52: 😍

PR #149, opened 2020-08-06
(Android) Implement exposure history methods

juancruzgs:
#### Description:

Implement Android methods to get current exposures and last detection date.
No changes made on the UI / Javascript side.

#### How to test:
If you got an exposure notification in the last 14 days you can go to:
- Exposure history (tab in the middle of the screen)
- EN Debug Menu > Show Exposures

If you go to `EN Debug Menu > Detect Exposures Now` it should update the header that says "Updated X minutes ago" (after an app restart)


juancruzgs:
The new methods are only getting called once the app is restarted.
I think they should be called from Javascript each time the Activity is resumed, in order to reflect the current state.
Should we create a ticket to do that?

johnschoeman:
> The new methods are only getting called once the app is restarted.
> I think they should be called from Javascript each time the Activity is resumed, in order to reflect the current state.
> Should we create a ticket to do that?

Yes, that would be great!

PR #153, opened 2020-08-07
Expose method to check if bluetooth is enabled

juancruzgs:
#### Description:

React Native needs a method to check if the bluetooth is enabled or not, this PR exposes its status.
Also expose a method to check if location is enabled on Android (GAEN requires both to work)

### DIFF
--- a/android/app/src/main/java/covidsafepaths/bt/exposurenotifications/ExposureNotificationsModule.java
+++ b/android/app/src/main/java/covidsafepaths/bt/exposurenotifications/ExposureNotificationsModule.java
@@ -1,6 +1,7 @@
 package covidsafepaths.bt.exposurenotifications;
 
 import android.app.Activity;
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/covidsafepaths/bt/exposurenotifications/ExposureNotificationsModule.java
```diff
@@ -105,6 +106,11 @@ public void getCurrentENPermissionsStatus(final Callback callback) {
                         });
     }
 
+    @ReactMethod
```
mattThousand, line 109: Should this go in the `DeviceInfo` module instead?

juancruzgs: Done!

PR #154, opened 2020-08-07
(iOS) Generate random padding data for each key submission 

mattThousand:
### Why 
Per https://google.github.io/exposure-notifications-server/server_functional_requirements.html, the key submission request should contain random data to obscure the size of the request to network packet sniffers

### This Commit
This commit generates random data with each key submission request to be added to the payload

### DIFF
--- a/ios/BT/API/Requests/DiagnosisKeyRequests.swift
+++ b/ios/BT/API/Requests/DiagnosisKeyRequests.swift
@@ -60,8 +60,18 @@ enum DiagnosisKeyListRequest: APIRequest {
         "appPackageName": Bundle.main.bundleIdentifier!,
         "verificationPayload": certificate,
         "hmackey": hmacKey,
-        "padding": String(decoding: Data(), as: UTF8.self)
+        "padding": Data.randomPadding(size: 1000)
       ]
     }
   }
+}
+
+private extension Data {
+
+  static func randomPadding(size: Int) -> String {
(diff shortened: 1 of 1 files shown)


ios/BT/API/Requests/DiagnosisKeyRequests.swift
```diff
@@ -60,8 +60,18 @@ enum DiagnosisKeyListRequest: APIRequest {
         "appPackageName": Bundle.main.bundleIdentifier!,
         "verificationPayload": certificate,
         "hmackey": hmacKey,
-        "padding": String(decoding: Data(), as: UTF8.self)
+        "padding": Data.randomPadding(size: 1000)
```
emilianojuan, line 63: can you use a constant for that 1000 and maybe a short description of why that number?

ios/BT/API/Requests/DiagnosisKeyRequests.swift
```diff
@@ -53,19 +53,27 @@ enum DiagnosisKeyListRequest: APIRequest {
...
+        "padding": Data.randomPadding(size: .paddingSize())
       ]
     }
   }
 }
 
+private extension Int {
+  /// Per https://google.github.io/exposure-notifications-server/server_functional_requirements.html
+  /// we add random data to obscure the size of the request (recommended size is ~1-2kb)
+  static func paddingSize() -> Int {
```
emilianojuan, line 72: nice

PR #156, opened 2020-08-07
Update home screen design

devinjameson:
Why: the design of our home screen has changed.

This commit:
- Updates the design of the home screen
- Updates the home screen spec
- Updates various other global styles

Co-Authored-By: amandabeiner <aebeiner@gmail.com>

home


src/Home/Home.spec.tsx
```diff
@@ -1,101 +1,255 @@
...
 
-import { ENPermissionStatus } from "../PermissionsContext"
 import Home from "./Home"
+import { PermissionsContext, ENPermissionStatus } from "../PermissionsContext"
+import { PermissionStatus } from "../permissionStatus"
+import { isBluetoothEnabled } from "../gaen/nativeModule"
+import { isPlatformiOS } from "../utils/index"
 
-afterEach(cleanup)
+jest.mock("@react-navigation/native")
```
aledustet, line 14: Some of this mocking was being shielded by having a screen component and a "launcher" component. I agree that this one is testing the native API more closely, but it also makes these tests way harder to reason and definitely harder to mock. Solid work here getting it to pass 👏

src/Home/Home.spec.tsx
```diff
@@ -1,101 +1,255 @@
...
-      fireEvent.press(button)
-      await wait(() => {
-        expect(requestPermission).toHaveBeenCalled()
+      const fixBluetoothButton = getByTestId("home-bluetooth-status-container")
+      const alert = jest.spyOn(Alert, "alert")
+
+      await wait(async () => {
+        fireEvent.press(fixBluetoothButton)
+        await wait(() => {
+          expect(alert).toHaveBeenCalled()
```
aledustet, line 162: Since we have a couple of alert calls on the component how do you feel about adding arguments to this assertion?

src/Home/Home.tsx
```diff
@@ -49,59 +73,281 @@ const Home = ({
...
+        {
+          text: t("common.okay"),
+        },
+      ],
+    )
+  }
+
+  const { applicationName } = useApplicationInfo()
+  const appDownloadLink = "https://pathcheck.org"
+  // TODO: Replace with actual app download link
```
aledustet, line 90: I created this ticket on Trello https://trello.com/c/YVm6ejAI would you mind reviewing and then we could potentially remove the TODO?

devinjameson: Perfect thank you for doing that!

src/Home/Home.spec.tsx
devinjameson: Sounds good!

src/Home/Home.tsx
```diff
@@ -49,59 +75,280 @@ const Home = ({
...
+      [
+        {
+          text: t("common.okay"),
+        },
+      ],
+    )
+  }
+
+  const { applicationName } = useApplicationInfo()
+  const appDownloadLink = "https://pathcheck.org"
```
johnschoeman, line 91: should this be coming in from a variable or constant?

src/Home/Home.tsx
```diff
@@ -49,59 +75,280 @@ const Home = ({
...
+      t("home.bluetooth.bluetooth_disabled_error_message"),
+      [
+        {
+          text: t("common.okay"),
+        },
+      ],
+    )
+  }
+
+  const { applicationName } = useApplicationInfo()
```
johnschoeman, line 90: what do you think about grouping this with the other application data?

src/Home/index.tsx
```diff
@@ -1,59 +0,0 @@
...
-
-import { usePermissionsContext } from "../PermissionsContext"
-import { useStatusBarEffect } from "../navigation"
-import Home from "./Home"
-
-import { Icons, Images } from "../assets"
-import { Spacing, Layout } from "../styles"
-
-const HomeScreen = (): JSX.Element => {
-  useStatusBarEffect("light-content")
```
johnschoeman, line 14: I am personally of two minds on the `ConnectorComponent` vs `PresentationalComponent` setup for screens. So happy with with change in general, but could you provide some context on why we are inlining the components for the home screen here?

src/Home/Home.tsx
```diff
@@ -49,59 +75,280 @@ const Home = ({
...
 }
 
+interface ActivationStatusProps {
+  headerText: string
+  isActive: boolean
+  fixAction: () => void
+  testID: string
+}
+
+const ActivationStatusSection: FunctionComponent<ActivationStatusProps> = ({
```
johnschoeman, line 203: Activation is a new domain concept for me. Does this mean 'Things that require permissions to be active?' if so what do you think about `PermissionStatusIndicator` for a name here?

Either way, is `Activation` a word that used used in the business domain? or is this something we are introducing in the codebase?

src/Home/index.tsx
devinjameson: I'm not sure if I'm answering your question, but it didn't seem like the `Home/index` structure was adding anything useful. Seems more straightforward to just include it all in one component.

src/Home/Home.tsx
devinjameson: It means a section that displays an active or disabled state to the user, e.g. "Bluetooth is active" or "Proximity tracing is active". IMO `PermissionStatusIndicator` wouldn't work because Bluetooth isn't really a permission. We're just telling the user whether Bluetooth is on or off because EN uses it. Does that make sense?

src/Home/Home.tsx
devinjameson: There's a card cut to route to the actual app download link, which will probably include getting this from the env.

src/Home/Home.tsx
johnschoeman: it does, thanks for the explanation!

src/Home/index.tsx
johnschoeman: makes sense to me!  

---

for context the reason they were separated initially was for FunctionalCore/ImperativeShell separation of concerns.  (https://www.destroyallsoftware.com/screencasts/catalog/functional-core-imperative-shell, if you are interested) Basically trying to put all the IO in spot and the rest in another as this makes it easier to test/reason about.  I agree that its a bit awkward to do this at the component level, it really wants to be at a module level, but with hooks n things that can be hard to do.

devinjameson: Interesting. Thanks for this context. I haven't read much about this concept at all so will check it out tonight!

PR #158, opened 2020-08-09
Upgrade React Native to 0.62.2

devinjameson:


devinjameson:
> 👏 👏 💪
> 
> Are the ios unit tests running locally?

Yep! Not sure why they failed. Re-running now.

PR #161, opened 2020-08-10
Update remote and onboarding lokalise keys

aledustet:
Why:
----
Some of the onboarding flow lokalise keys are not in use and not nested under the onboarding key. As a consequence of a malfunctioning of the push script, some keys are left orphan on the lokalise project.

This Commit:
----
- Moved onboarding translations
- Run i18n:extract
- Run i18n:pull after cleaning up unused keys

Reviewers:
----
This one goes hand and hand with a cleanup of unused and obsolete keys on the lokalise project.


android/app/src/main/res/values-el/strings.xml
```diff
@@ -4,6 +4,4 @@
   <string name="exposure_notification_channel_name">Ειδοποίηση έκθεσης</string>
   <string name="exposure_notification_message">Κάποιος που βρέθηκε κοντά σας έχει διαγνωστεί θετικά με COVID-19. Πατήστε για περισσότερες πληροφορίες.</string>
   <string name="exposure_notification_title">Πιθανή έκθεση στον COVID-19</string>
-  <!-- Channel label for notifications shown in App > Notifications > Path Check BT -->
-  <string name="notification_channel_description">Ειδοποιήσεις PathCheck BT</string>
```
juancruzgs, line 8: This string should be removed from all the translation folders.
I can do it later, doesn't need to be included in this PR

aledustet: Not sure why those are still on any other translations? Feels like we did a pull at some point but didn't commit the results, because those keys are no longer in lokalise

PR #162, opened 2020-08-10
CI/CD Updates

mattThousand:
### Why
We'd like to update our Fastfile for our new CI/CD pipeline

### This Commit
This commit removes unused code from our Fastfile, as well as an unused `Swift` package


ios/Podfile.lock
```diff
@@ -1,523 +0,0 @@
-PODS:
```
johnschoeman, line 1: Should we be removing this file?

ios/fastlane/Fastfile
```diff
@@ -15,184 +15,27 @@
...
-      build_ios_app(
-        workspace: WORKSPACE,
-        scheme: GPS_STAGING_SCHEME,
-        configuration: "Staging",
-        export_method: "ad-hoc",
-        silent: true,
-        clean: true,
-        output_name: "COVIDSafePaths-staging.ipa",
+  lane :deploy_to_appcenter do |options|
+    # Push to AppCenter
```
johnschoeman, line 20: what do you think about removing this comment? The code seems self documenting enough here?

PR #163, opened 2020-08-10
Add local push to simulate exposure EN debug button.

mattThousand:
### Why
We need a way to simulate an exposure notification "on demand" and demo the corresponding Exposure Summary and Exposure Details screens to guide discussion about the level of configuration jurisdictions would like (per https://trello.com/c/hg1M9jhp/311-feature-flag-to-demo-gaen-exposure-notifications)

### This Commit
This commit updates the "simulate exposure" debug button to show a push as it would appear to a user when receiving a real exposure notification, and ensures that the button is visible in the builds that will go to testflight

### DIFF
--- a/src/More/ENDebugMenu.tsx
+++ b/src/More/ENDebugMenu.tsx
@@ -98,16 +98,16 @@ const ENDebugMenu = ({ navigation }: ENDebugMenuProps): JSX.Element => {
         <ScrollView>
           <View style={style.section}>
             <DebugMenuListItem
+              label="Simulate Exposure"
+              onPress={handleOnPressSimulationButton(
+                NativeModule.simulateExposure,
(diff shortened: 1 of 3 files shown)


src/More/ENDebugMenu.tsx
```diff
@@ -98,16 +98,16 @@ const ENDebugMenu = ({ navigation }: ENDebugMenuProps): JSX.Element => {
         <ScrollView>
           <View style={style.section}>
             <DebugMenuListItem
-              label="Show Last Processed File Path"
+              label="Simulate Exposure"
```
johnschoeman, line 101: Should these labels be translated? It seems like this is intended to be used by clients directly so it might need to be a higher level of polish on the functionality.

PR #170, opened 2020-08-11
Update tab bar styling

devinjameson:
Why: our tab bar styling has changed.

This commit:
- Changes the style of the tab bar to a white background with violet icons
- Restructures the TabBarNav folder with svgs to match the way we're
organizing our other svgs
-Refactors the way we handle focused and unfocused tab bar icon states

tabbar


src/assets/svgs/TabBarNav/index.ts
```diff
@@ -1,7 +1,13 @@
...
-export { LocationsActive, LocationsInactive } from "./Locations"
-export { MoreActive, MoreInactive } from "./More"
-export { ShieldActive, ShieldInactive } from "./Shield"
-export { NotificationIcon } from "./NotificationIcon"
-export { SelfAssessmentActive, SelfAssessmentInactive } from "./SelfAssessment"
+import Exposure from "./exposure"
+import QuestionMark from "./questionMark"
+import HorizontalDots from "./horizontalDots"
+import House from "./house"
+import CheckInBox from "./checkInBox"
```
amandabeiner, line 5: 💯 love this refactor and pattern

src/navigation/MainTabNavigator.tsx
```diff
@@ -48,7 +52,8 @@ const MainTabNavigator: FunctionComponent = () => {
   const HomeIcon: FunctionComponent<TabBarIconProps> = ({ focused, size }) => {
     return (
       <SvgXml
-        xml={focused ? Icons.HomeActive : Icons.HomeInactive}
+        xml={TabBarIcons.House}
+        fill={determineIconFill(focused)}
```
amandabeiner, line 56: Love taking advantage of the fill prop here 🙌  I don't think this would be too bad to inline:
`fill={focused ? Colors.primaryViolet : Colors.quaternaryViolet}` and might have the slight benefit of keeping the decision making logic close to the render logic, but this is 100% a style preference and your call!

src/navigation/MainTabNavigator.tsx
```diff
@@ -79,9 +85,8 @@ const MainTabNavigator: FunctionComponent = () => {
   }) => {
     return (
       <SvgXml
-        xml={
-          focused ? Icons.SelfAssessmentActive : Icons.SelfAssessmentInactive
-        }
+        xml={TabBarIcons.CheckInBox}
+        fill={determineIconFill(focused)}
```
amandabeiner, line 89: Oooh now I see that you extracted this function since it's called multiple times. One option might be to have a base `TabIcon` component in this file that always accepts a `fill` prop and handles the badge and fill logic, then have your `SelfAssessmentIcon` render that component. 

This is up to your discretion, though, I think it's fine as is!

src/navigation/MainTabNavigator.tsx
devinjameson: Yeah I actually went back and forth between these two options and went with wrapping it in a function to avoid repeating the colors. I could go either way though!

src/navigation/MainTabNavigator.tsx
devinjameson: Oh I see that's a good idea. I'm going to try that and will tag you in a commit.

devinjameson: Thoughts on 5b8303c?

devinjameson: Oops I mean 933998f29abe11a8eacc8cd0cc73a2d0dea921cb!

src/assets/svgs/TabBarNav/index.ts
devinjameson: Thanks!

src/navigation/MainTabNavigator.tsx
amandabeiner: @devinjameson yesss looks great 👏

devinjameson: Thanks!

amandabeiner:
> This version is more code of course but it feels more React-y not to have function calls within props.

@devinjameson I also agree with this! I tend to like moving the branching of display logic up a bit more, like you've done here, so that you don't have a lot of if/elses in every prop as new use cases arise. It can make it a little easier to keep track of what a component is responsible for, like you said!

PR #184, opened 2020-08-11
Update onboarding flow

devinjameson:
Why:
- We want a Welcome screen that matches the rest of the onboarding flow
- We want users to be able to change the language at any time during the onboarding flow

This commit:
- Updates the design of the Welcome screen
- Adds language selection functionality throughout the onboarding flow

~~Note: there is one small issue in this commit that arises when navigating back to the More menu from the language selection screen. The header animates in a strange way. This will be fixed in a future commit.~~
Solved ✅

onboarding


src/More/LanguageSelection.tsx
```diff
@@ -1,71 +1,139 @@
...
+      <TouchableOpacity
+        style={style.languageButton}
+        onPress={handleOnSelectLanguage}
+      >
+        <GlobalText style={languageButtonTextStyles}>{item.label}</GlobalText>
+      </TouchableOpacity>
+    )
   }
 
+  const headerHeight = 70
```
johnschoeman, line 60: should these styles live in the stylesheet?

src/Onboarding/Welcome.tsx
```diff
@@ -18,87 +13,74 @@ import { Spacing, Colors, Typography, Outlines } from "../styles"
...
+    <View style={style.container}>
+      <TouchableOpacity
+        onPress={() => navigation.navigate(Screens.LanguageSelection)}
+        style={style.languageButtonContainer}
+      >
+        <GlobalText style={style.languageButtonText}>{languageName}</GlobalText>
+      </TouchableOpacity>
+      <View>
+        <Image
+          source={Images.MinnesotaMap}
```
johnschoeman, line 35: Does this need to be conditional on the build? we probably dont want other HAs to have this background image?

devinjameson: Yeah it does, I was going to cut a card for this work. Thoughts?

src/More/LanguageSelection.tsx
devinjameson: Oh, yeah the should. For some reason I was getting an error when trying to do that before, but looks good now.

PR #185, opened 2020-08-11
(Android) Update app name in launcher

juancruzgs:
#### Description:

We were always displaying the app name "PatchCheck BT" in the launcher.
This PR will show the `DISPLAY_NAME` that is defined in the env file.

#### How to test:

Install the app and look at the text below the app icon in the launcher.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -21,7 +21,6 @@
 
         <activity
             android:name="covidsafepaths.bt.SplashActivity"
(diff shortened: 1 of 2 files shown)


tparesi:
I know this is different, but the Splash Screen for Android is still showing PathCheck BT- some drawable stuff I couldn't figure out.

juancruzgs:
You are right @tparesi , that's a SVG file we have in the native side.
I'll create a ticket to dynamically change it

PR #186, opened 2020-08-12
(Android) Add revisionToken to submit keys requests

juancruzgs:
#### Description:

Currently the `exposureKeyModule.postDiagnosisKeys` method submits the exposure keys to the exposure notifications server but it is not adding the revisionToken on the request. This token will be returned from the exposure notification server once the new keys are submitted and it has to be used on subsequent updates of the exposure keys.

This PR saves the `revisionToken` in a shared preference and uses it in subsequent requests

#### Linked issues:

https://trello.com/c/ogfTrV4b/333-when-the-post-diagnosis-keys-process-ends-and-th-user-shares-the-keys-with-the-exposures-server-we-need-to-store-a-revision-toke

#### How to test:

Verify that `verifyToken` is not sent in the first request and it's added in subsequent requests.

### DIFF
--- a/android/app/src/main/java/covidsafepaths/bt/exposurenotifications/network/DiagnosisKeyUploader.java
+++ b/android/app/src/main/java/covidsafepaths/bt/exposurenotifications/network/DiagnosisKeyUploader.java
@@ -31,7 +31,7 @@ import com.android.volley.Response.ErrorListener;
 import com.android.volley.Response.Listener;
 import com.android.volley.VolleyError;
 import com.android.volley.toolbox.HttpHeaderParser;
-import com.android.volley.toolbox.JsonRequest;
+import com.android.volley.toolbox.JsonObjectRequest;
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.BaseEncoding;
 import com.google.common.util.concurrent.FluentFuture;
@@ -49,6 +49,7 @@ import java.security.SecureRandom;
 
 import covidsafepaths.bt.exposurenotifications.ExposureKeyModule;
 import covidsafepaths.bt.exposurenotifications.common.AppExecutors;
+import covidsafepaths.bt.exposurenotifications.storage.ExposureNotificationSharedPreferences;
 import covidsafepaths.bt.exposurenotifications.utils.Util;
 
 /**
@@ -73,10 +74,12 @@ public class DiagnosisKeyUploader {
 
     private final Context context;
     private final Uris uris;
+    private ExposureNotificationSharedPreferences prefs;
 
     public DiagnosisKeyUploader(Context context) {
         this.context = context;
         uris = new Uris(context);
+        prefs = new ExposureNotificationSharedPreferences(context);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


android/app/src/main/java/covidsafepaths/bt/exposurenotifications/network/DiagnosisKeyUploader.java
```diff
@@ -189,7 +192,8 @@ public DiagnosisKeyUploader(Context context) {
                         .put("hmackey", ExposureKeyModule.hmacKey)
                         .put("padding", randomBase64Data(Util.getRandomNumber()))
                         .put("regions", regionJson)
-                        .put("appPackageName", BuildConfig.ANDROID_APPLICATION_ID);
+                        .put("appPackageName", BuildConfig.ANDROID_APPLICATION_ID)
+                        .put("revisionToken", prefs.getRevisionToken(null));
```
aledustet, line 196: Will the null value be turned into the empty string when converted to JSON?

juancruzgs: It won't be sent if the value is `null`.
Should we send it as an empty string?

mattThousand:
@juancruzgs would it be prudent to store this in realm instead of storing it unencrypted in SharedPrefs?

juancruzgs:
I think I can use https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences to encrypt its values.
I didn't want to use RealmObjects to store a single value because I have to add migrations and queries. It seems overkill to me, but I'll implement it that way if I cannot use `EncryptedSharedPreferences`

android/app/src/main/java/covidsafepaths/bt/exposurenotifications/network/DiagnosisKeyUploader.java
aledustet: They do `len(data.RevisionToken)` on the go side, my newly acquired go knowledge is pretty close to 0 but seems to me that they are measuring the len, if I run code with nil it throws an error, but the type of the request is a string so I don't know how the encoding works so I would vote for an empty string.

juancruzgs: Oh I saw that you added it in the card description. My bad, I didn't read it.
I'll add it now, thanks!

juancruzgs:
Revision token is now stored encrypted in Realm.
We had a `RealmObject` to store key-value pairs so I'm using it

android/app/src/main/java/covidsafepaths/bt/exposurenotifications/storage/Migration.kt
```diff
@@ -1,18 +1,32 @@
...
+            version++
+        }
+
+        if (version == 2L) {
+            schema.get("KeyValues")
+                ?.renameField("lastDownloadedKeyZipFileName", "value")
+                ?.removePrimaryKey()
+                ?.transform { obj -> obj.setString("id", KeyValues.LAST_PROCESSED_FILE_NAME_KEY) }
+                ?.addPrimaryKey("id")
+        }
```
octohub, line 30: Do we need to worry about migrations yet given this isn't being used by real users?

juancruzgs: We don't need to worry, but I didn't want to force the QA team to re-install the app since they will lose all the keys they have (and will have to wait 24 hours to test again).

PR #187, opened 2020-08-12
Short-circuit key uploading process for empty set 

aledustet:
Why:
----
If no keys are uploaded to the exposure notifications server, the process raise an [error] for an empty set of keys. Once we validate a test code, we should stop the process from continuing since no data will cause an error.

This Commit:
----
- Navigate to the affected user flow completed screen if no exposure keys are available to be uploaded on the device
- Move zendesk API closer to where it is used

[error]: https://github.com/google/exposure-notifications-server/blob/fd8ade2817d1e148336ba4076c1f0989b35797d1/pkg/verification/utils.go#L57


src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -64,6 +64,12 @@ const CodeInputForm: FunctionComponent = () => {
       if (response.kind === "success") {
         const token = response.body.token
         const exposureKeys = await strategy.getExposureKeys()
+
+        if (exposureKeys.length === 0) {
```
aledustet, line 68: I stopped the code here, but the verification server seems ok with having an empty set of keys, should we still run it down and then skip the publish step? Not 100% familiar with the verification server inner works.

src/AffectedUserFlow/CodeInput/CodeInputForm.spec.tsx
```diff
@@ -39,57 +39,105 @@ describe("CodeInputForm", () => {
...
+          )
+          expect(navigateSpy).toHaveBeenCalledWith(
+            Screens.AffectedUserPublishConsent,
+          )
+        })
+      })
+    })
+
+    describe("when the device does not contains exposure keys", () => {
+      it("navigates directly to the complete screen", async () => {
```
johnschoeman, line 106: Is this the behavior we want? Another option would be to not navigate at all and show a pop up that they have no exposure keys? I suppose this is mostly an edge case that happens during development so perhaps it doesnt matter enough to change?

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
johnschoeman: This seems okay for now to me, but imo it would be better to have the api handle these edge cases rather than the client.

johnschoeman (approved):
LGTM!

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
aledustet: I was feeling the same way but the way we are handling the error messages gives little to no detail to the client-side, I created a [ticket] for it and going to try it in android to see what's the level of detail we are getting. 

[ticket]: https://trello.com/c/KeLmnh0c

aledustet:
Closing this because the error happens sooner when we validate the token.

PR #189, opened 2020-08-12
Fix Home tests warnings

aledustet:
Why:
----
Some of the tests on the Home spec are showing warnings about act.

This Commit:
----
- Remove some of the async nesting on waitFor blocks
- Mock the useApplicationInfo hook on the home tests


devinjameson:
Nice!

PR #190, opened 2020-08-12
Add revision token to key submission request (iOS)

mattThousand:
#### Description:
Currently the `exposureKeyModule.postDiagnosisKeys` method submits the exposure keys to the exposure notifications server but it is not adding the revisionToken on the request. This token will be returned from the exposure notification server once the new keys are submitted and it has to be used on subsequent updates of the exposure keys.

This PR saves the `revisionToken` in Realm and uses it in subsequent requests

#### Linked issues:

https://trello.com/c/ogfTrV4b/333-when-the-post-diagnosis-keys-process-ends-and-th-user-shares-the-keys-with-the-exposures-server-we-need-to-store-a-revision-toke

#### How to test:
Verify that `revisionToken` is not sent in the first request and it's added in subsequent requests.

### DIFF
--- a/ios/BT/ExposureManager.swift
+++ b/ios/BT/ExposureManager.swift
@@ -313,11 +313,15 @@ final class ExposureManager: NSObject {
         let currentKeys = allKeys.current()
         
         let regionCodes = ReactNativeConfig.env(for: .regionCodes)!.regionCodes
+
+        let revisionToken = BTSecureStorage.shared.userState.revisionToken
         
-        APIClient.shared.request(DiagnosisKeyListRequest.post(currentKeys.compactMap { $0.asCodableKey }, regionCodes, certificate, HMACKey),
+        APIClient.shared.request(DiagnosisKeyListRequest.post(currentKeys.compactMap { $0.asCodableKey }, regionCodes, certificate, HMACKey, revisionToken),
                                  requestType: .postKeys) { result in
                                   switch result {
-                                  case .success:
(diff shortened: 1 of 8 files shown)


ios/BT/ExposureManager.swift
```diff
@@ -313,11 +313,15 @@ final class ExposureManager: NSObject {
...
         
-        APIClient.shared.request(DiagnosisKeyListRequest.post(currentKeys.compactMap { $0.asCodableKey }, regionCodes, certificate, HMACKey),
+        APIClient.shared.request(DiagnosisKeyListRequest.post(currentKeys.compactMap { $0.asCodableKey }, regionCodes, certificate, HMACKey, revisionToken),
                                  requestType: .postKeys) { result in
                                   switch result {
-                                  case .success:
+                                  case .success(let response):
+                                    // Save revisionToken to use on subsequent key submission requests
+                                    BTSecureStorage.shared.revisionToken = response.revisionToken ?? .default
                                     resolve("Submitted: \(currentKeys.count) keys.")
```
aledustet, line 325: Maybe in the future, we could validate how many keys were inserted vs the keys we sent since they also send back an `insertedExposures` just food for thought 😄

aledustet, line 325: Maybe in the future, we could validate how many keys were inserted vs the keys we sent since they also send back an `insertedExposures` just food for thought 😄

PR #191, opened 2020-08-12
Update Bluetooth status on home screen on toggle

amandabeiner:
#### Description:
Why:
As a user, who has been prompted to enable bluetooth, I want to see my
new bluetooth status reflected in the home screen so that I know it is
working.

This commit:
Introduces the AppState listener from react native. When an iOS user
foregrounds the app (in order to activate bluetooth in their settings),
or an Android user focuses the app (in order to either activate their
bluetooth settings, or use the notification menu to toggle bluetooth),
the app will now refetch the bluetooth status and set it into state.

This also fixes a bug I experienced where the status was not being
properly set on Android, because `isBluetoothEnabled` returned a
boolean, and we were checking if `status === 'true'`

#### Linked issues:

Fixes [Trello](https://trello.com/c/ZPXtZnvi/320-when-i-do-not-have-bluetooth-enabled-and-the-app-tells-me-to-turn-it-on-i-want-to-see-that-turning-it-on-is-reflected-in-the-app)

#### Screenshots:
Aug-12-2020 13-00-38

#### How to test:
- With your bluetooth off, open the app on your device
- Turn your bluetooth on, and wait for it to activate
- revisit the app
- Expect the bluetooth status to have updated

Notes:
- I am unable to test this on an iPhone, as I do not have an iPhone and the simulator does not support bluetooth
- There is an edge case where if a user toggles their bluetooth settings and then quickly revisits the app, the bluetooth might not be active by the time they revisit the app, since it can take a second or two to activate. If we want to address this issue, we will have to create a native event emitter that will emit the updated bluetooth status regardless of application state.

### DIFF
--- a/src/Home/useBluetoothStatus.ts
+++ b/src/Home/useBluetoothStatus.ts
@@ -1,0 +1,33 @@
+import { useEffect, useState } from "react"
+import { Platform, AppState } from "react-native"
+import { isBluetoothEnabled } from "../gaen/nativeModule"
+
+export const useBluetoothStatus = (): boolean => {
+  const [btStatus, setBTStatus] = useState(false)
+  useEffect(() => {
+    const fetchBTEnabled = async () => {
+      const status = await fetchBTStatus()
+      setBTStatus(status)
+    }
+
+    AppState.addEventListener(determineOSListener(), () => fetchBTEnabled())
+    return AppState.removeEventListener(determineOSListener(), () =>
+      fetchBTEnabled(),
+    )
+  }, [])
+
+  return btStatus
+}
+
+const fetchBTStatus = async (): Promise<boolean> => {
+  return isBluetoothEnabled()
+}
+
+type ListenerMethod = "focus" | "change"
+const determineOSListener = (): ListenerMethod => {
+  return Platform.select({
+    ios: "change",
+    android: "focus",
+    default: "change",
+  })
+}
--- a/src/gaen/nativeModule.ts
+++ b/src/gaen/nativeModule.ts
@@ -140,8 +140,9 @@ export const getVersion = async (): Promise<string> => {
   return deviceInfoModule.getVersion()
 }
 
-export const isBluetoothEnabled = async (): Promise<string> => {
-  return deviceInfoModule.isBluetoothEnabled()
+export const isBluetoothEnabled = async (): Promise<boolean> => {
+  const bluetoothStatus = deviceInfoModule.isBluetoothEnabled()
+  return bluetoothStatus === true || bluetoothStatus === "true"
 }
 
 // Debug Module
--- a/src/Home/Home.spec.tsx
+++ b/src/Home/Home.spec.tsx
@@ -13,8 +13,8 @@ import "@testing-library/jest-native/extend-expect"
 import Home from "./Home"
 import { PermissionsContext, ENPermissionStatus } from "../PermissionsContext"
 import { PermissionStatus } from "../permissionStatus"
-import { isBluetoothEnabled } from "../gaen/nativeModule"
 import { isPlatformiOS } from "../utils/index"
+import { useBluetoothStatus } from "./useBluetoothStatus"
 
 jest.mock("@react-navigation/native")
 ;(useNavigation as jest.Mock).mockReturnValue({ navigate: jest.fn() })
@@ -34,12 +34,13 @@ jest.mock("../More/useApplicationInfo", () => {
     },
   }
 })
+jest.mock("./useBluetoothStatus.ts")
 
 describe("Home", () => {
   describe("When the enPermissionStatus is enabled and authorized and Bluetooth is on", () => {
     it("renders an active message", async () => {
-      const isBluetoothOn = "true"
-      ;(isBluetoothEnabled as jest.Mock).mockResolvedValueOnce(isBluetoothOn)
+      const isBluetoothOn = true
+      ;(useBluetoothStatus as jest.Mock).mockReturnValue(isBluetoothOn)
 
       const enPermissionStatus: ENPermissionStatus = ["AUTHORIZED", "ENABLED"]
       const permissionProviderValue = createPermissionProviderValue(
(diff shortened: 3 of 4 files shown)


src/Home/useBluetoothStatus.ts
```diff
@@ -0,0 +1,31 @@
...
+export const useBluetoothStatus = (): boolean => {
+  const [btStatus, setBTStatus] = useState(false)
+  useEffect(() => {
+    const fetchBTEnabled = async () => {
+      const status = await fetchBTStatus()
+      setBTStatus(status)
+    }
+
+    AppState.addEventListener(getOSListener(), () => fetchBTEnabled())
+    return AppState.removeEventListener(getOSListener(), () => fetchBTEnabled())
```
devinjameson, line 14: For my own learning, can you explain what this line does?

src/Home/useBluetoothStatus.ts
```diff
@@ -0,0 +1,31 @@
...
+  }, [])
+
+  return btStatus
+}
+
+const fetchBTStatus = async (): Promise<boolean> => {
+  const status = await isBluetoothEnabled()
+  return status === true || status === "true"
```
devinjameson, line 22: Good catch here!

src/Home/useBluetoothStatus.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const useBluetoothStatus = (): boolean => {
+  const [btStatus, setBTStatus] = useState(false)
+  useEffect(() => {
+    const fetchBTEnabled = async () => {
+      const status = await fetchBTStatus()
+      setBTStatus(status)
+    }
+
+    AppState.addEventListener(getOSListener(), () => fetchBTEnabled())
```
devinjameson, line 13: What are your thoughts on `determineOSListener` here instead of `getOSListener`? I tend to find the former more clear.

src/Home/useBluetoothStatus.ts
```diff
@@ -0,0 +1,31 @@
...
+  return btStatus
+}
+
+const fetchBTStatus = async (): Promise<boolean> => {
+  const status = await isBluetoothEnabled()
+  return status === true || status === "true"
+}
+
+const getOSListener = (): "focus" | "change" => {
+  return Platform.select({
```
devinjameson, line 25: Never seen this `Platform.select` thing before, very cool!

src/Home/useBluetoothStatus.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return btStatus
+}
+
+const fetchBTStatus = async (): Promise<boolean> => {
+  const status = await isBluetoothEnabled()
+  return status === true || status === "true"
+}
+
+const getOSListener = (): "focus" | "change" => {
```
devinjameson, line 24: Again for my own education, thoughts on

`const getOSListener = (): "focus" | "change" =>`

vs

```
type ListenerMethod = "focus" | "change"

const getOSListener = (): ListenerMethod =>
```

src/gaen/nativeModule.ts
```diff
@@ -140,8 +140,9 @@ export const getVersion = async (): Promise<string> => {
   return deviceInfoModule.getVersion()
 }
 
-export const isBluetoothEnabled = async (): Promise<string> => {
-  return deviceInfoModule.isBluetoothEnabled()
+export const isBluetoothEnabled = async (): Promise<boolean> => {
+  const bluetoothStatus = deviceInfoModule.isBluetoothEnabled()
+  return bluetoothStatus === true || bluetoothStatus === "true"
```
amandabeiner, line 145: I was running into an issue (on Android, idk if this is OS specific), where this function was returning the boolean value `true`, then failing to update the status because `true === "true"` is false.

src/Home/useBluetoothStatus.ts
amandabeiner: Absolutely! A function returned from the `useEffect` hook is called before the component unmounts (it's similar to `componentWillUnmount` from the olden days), so it's often used to define what "clean up" actions you want to run before this component is removed from the tree. In this case, since we are adding an event listener, we we can remove the event listener so that it doesn't hang around and keep running these fetch requests when the user navigates away from the Home screen.

src/Home/useBluetoothStatus.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const fetchBTStatus = async (): Promise<boolean> => {
+  return isBluetoothEnabled()
+}
+
+const getOSListener = (): "focus" | "change" => {
+  return Platform.select({
+    ios: "change",
+    android: "focus",
```
amandabeiner, line 27: `focus` is android-only, so it can't be used for `ios`. `focus` and `blur` events account for both foregrounding the app after it has been backgrounded (like `change`), and also re-focusing the app after pulling down the notification bar (where Android users can toggle bluetooth on/off).

src/Home/useBluetoothStatus.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const useBluetoothStatus = (): boolean => {
+  const [btStatus, setBTStatus] = useState(false)
+  useEffect(() => {
+    const fetchBTEnabled = async () => {
+      const status = await fetchBTStatus()
+      setBTStatus(status)
+    }
+
+    AppState.addEventListener(getOSListener(), () => fetchBTEnabled())
```
amandabeiner, line 13: Hooking into AppState because certain change events, such as toggling BT via the Android notification bar, do not affect the React component lifecycle and will not cause a rerender/update.

src/Home/useBluetoothStatus.ts
devinjameson: Ohhhh I see that makes sense. Thank you! 🙏🏼

src/Home/useBluetoothStatus.ts
amandabeiner: Yes! I like both these suggestions, thanks!

src/gaen/nativeModule.ts
juancruzgs: I can change the return type from `Boolean` to `String` in the native side if it's better for you.
I thought iOS was returning a boolean too

amandabeiner: I personally would prefer to return a `Boolean` rather than a `String`, so I think you made the right call! I'm fine with either keeping this as it is (`status === true || status === "true"`) or updating the iOS to return a boolean rather than a string. What do you think?

juancruzgs: @amandabeiner I also prefer to use a Boolean instead of an String here.
@mattThousand Is iOS returning a `String` when `isBluetoothEnabled` is called? If so, is it possible to change it to a `Boolean`?

PR #192, opened 2020-08-12
Use generic launch image instead of MN specific image

devinjameson:
Why: we would like the image presented on the app launch screen to be generic until we have a white-labeling solution.

## Before
Screen Shot 2020-08-12 at 1 01 11 PM

## After
Screen Shot 2020-08-12 at 12 57 32 PM


src/Onboarding/Welcome.tsx
```diff
@@ -20,8 +26,18 @@ const Welcome: FunctionComponent = () => {
   } = useTranslation()
   const languageName = getLocalNames()[localeCode]
   const appName = env.IN_APP_NAME || "PathCheck"
+  const region = env.REGION_CODES
   useStatusBarEffect("dark-content")
 
+  const determineImageSource = (): ImageSourcePropType => {
```
tparesi, line 32: hmm - this seems like it could get quite cumbersome as the number of regions grow. I am thinking we shouldn't even be storing environment dependent assets in the repo. What if there is a welcomeImageUrl in the env that pulls the image in?

devinjameson: Yeah that's definitely a concern. This probably wants to be part of a broader discussion around handling app customization. Although I'm not sure I'm opposed to storing environment dependent assets in the repo. What do you see as the downside of this?

Also can yan you clarify how the `welcomeImageURL` thing would work?

devinjameson: @johnschoeman Not sure if you've done any thinking about white labeling solutions, but thought you may have some thoughts or insight here.

tparesi: Seems like the white labeling is going to involve lots of assets unique to jurisdictions (like the app icon). If we store all those assets locally, we would then be the ones managing that part of the app for them. I thought we were trying to avoid this with the use of the .env.bt file. Is there a way to include this into the setup_ha script - pulling in assets?

devinjameson: That makes total sense. There's some conversation about this in Slack so maybe we should just use the generic image for now and come back to this in a more holistic way.

devinjameson: @tparesi Lmk if you think this is good to merge for now.

PR #194, opened 2020-08-12
Use the app or play store link for sharing the app

tparesi:
Why?
------
When sharing the app, send the user to the app store to download the app - HA dependent.

#### Linked issues:

[Trello](https://trello.com/c/YVm6ejAI/312-when-i-press-the-share-link-on-the-home-screen-i-would-like-to-get-the-actual-store-link-for-the-application)
[Pathcheck Mobile Resource PR](https://github.com/Path-Check/pathcheck-mobile-resources/pull/11)

### DIFF
--- a/src/Home/Home.tsx
+++ b/src/Home/Home.tsx
@@ -10,11 +10,13 @@ import {
   Image,
   View,
   Share,
+  Platform,
 } from "react-native"
 import { useSafeAreaInsets } from "react-native-safe-area-context"
 import { useTranslation } from "react-i18next"
 import { SvgXml } from "react-native-svg"
 import { useNavigation } from "@react-navigation/native"
+import env from "react-native-config"
 
 import {
   usePermissionsContext,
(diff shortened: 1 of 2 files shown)


tparesi:
I believe I used the correct URL's - added them to Pathcheck Mobile Resources repo - BUT since the app is not in the app store, I don't believe they will work yet. Not sure the best way to proceed on this.

src/Home/Home.tsx
```diff
@@ -88,9 +90,13 @@ const HomeScreen: FunctionComponent = () => {
     )
   }
 
-  const appDownloadLink = "https://pathcheck.org"
-
   const handleOnPressShare = async () => {
+    const isIOS = Platform.OS === "ios"
+
+    const appDownloadLink = isIOS
+      ? env.IOS_APP_STORE_URL
```
johnschoeman, line 97: what happens when these links are undefined?

PR #197, opened 2020-08-13
 Refactor Exposure Manager

emilianojuan:
#### Description:

Refactor Exposure Manager to enable testability
- Adds code coverage, ExposureManager and SecureStore coverage is at ~ 90%
- Adds Unit Test for lifecycle and status of the manager
- Refactor signatures to decouple RN from ExposureManager
- Introduce dependency injection and protocol oriented programming to facilitate mocking

https://trello.com/c/nIUkkuLw/277-ios-codebase-cleanup


ios/BT/Info.plist
```diff
@@ -10,7 +10,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleDisplayName</key>
-	<string>PathCheck BT</string>
+	<string>PathCheck MN</string>
```
mattThousand, line 13: Looks like this may have been included by accident?

ios/BT/Storage/BTSecureStorage.swift
```diff
@@ -2,7 +2,7 @@ import ExposureNotification
 import Foundation
 import RealmSwift
 
-final class BTSecureStorage: SafePathsSecureStorage {
+class BTSecureStorage: SafePathsSecureStorage {
```
mattThousand, line 5: why not `final`? 🤔  Are we subclassing `BTSecureStorage` anywhere?

ios/BT/bridge/ENPermissionsModule.m
```diff
@@ -10,11 +11,19 @@ @implementation ENPermissionsModule
 RCT_EXPORT_MODULE();
 
 RCT_EXPORT_METHOD(requestExposureNotificationAuthorization: (RCTResponseSenderBlock)callback) {
-  [[ExposureManager shared] requestExposureNotificationAuthorizationWithEnabled:YES callback:callback];
+  [[ExposureManager shared] requestExposureNotificationAuthorizationWithEnabled:YES callback:^(NSError * _Nullable error) {
```
mattThousand, line 14: It would be great if we could refactor these to use `Promise`s instead of callbacks (would require changing the corresponding function signatures in  `nativeModule.ts`)

ios/BT/Extensions/Exposure Notifications/ENManager+Extensions.swift
```diff
@@ -0,0 +1,26 @@
...
+
+  func authorizationStatus() -> ENAuthorizationStatus
+
+  var exposureNotificationEnabled: Bool { get }
+  func setExposureNotificationEnabled(_ enabled: Bool, completionHandler: @escaping ENErrorHandler)
+  func detectExposures(configuration: ENExposureConfiguration, diagnosisKeyURLs: [URL], completionHandler: @escaping ENDetectExposuresHandler) -> Progress
+  func getExposureInfo(summary: ENExposureDetectionSummary, userExplanation: String, completionHandler: @escaping ENGetExposureInfoHandler) -> Progress
+  func getDiagnosisKeys(completionHandler: @escaping ENGetDiagnosisKeysHandler)
+  func getTestDiagnosisKeys(completionHandler: @escaping ENGetDiagnosisKeysHandler)
+}
```
mattThousand, line 19: 👍

ios/BT/Storage/BTSecureStorage.swift
emilianojuan: we are making BTSecureStorage not final so we can mock it, I can create a protocol and make this implementation final, what do you think?

mattThousand: ah I see - thanks for explaining - either approach is fine 👍

ios/BT/bridge/ENPermissionsModule.m
emilianojuan: that will also make an impact on the Android side? This change is so to divide the native component from the RN side if we have RN callbacks I think we are mixing the layers

ios/BT/Info.plist
emilianojuan: yep

ios/BT/bridge/ENPermissionsModule.m
mattThousand: good question - yes, it would require changes on the android side. My thinking is that it would be good for the native layer to have a consistent API (`nativeModule` currently has a mixture of promises and callbacks), but I would say that it's low priority, and as you point out it would be a little disruptive because it would require coordination with thee Android side

emilianojuan: I'll speak with Juan about this, maybe we can coordinate the effort

juancruzgs:
This method is still using Callbacks `exposureHistoryModule.getCurrentExposures`, the others are looking good
I'll make those changes in the Android side

ios/BT/API/APIClient.swift
```diff
@@ -6,12 +6,35 @@ enum RequestType {
   exposureConfiguration
 }
 
-final class APIClient {
+protocol APIClient {
```
mattThousand, line 9: 💯

ios/BT/API/Model/DownloadedPackage.swift
```diff
@@ -14,7 +19,7 @@ struct DownloadedPackage {
       return nil
     }
     do {
-      self = try archive.extractKeyPackage()
+      self = try archive.extractKeyPackage() as! DownloadedPackageImpl
```
mattThousand, line 22: hm, is there a way we can avoid force-unwrapping here? 🤔

emilianojuan: done!

ios/BT/ExposureManager.swift
```diff
@@ -349,118 +468,194 @@ final class ExposureManager: NSObject {
...
+
+  func cleanup() {
+    // Delete downloaded files from file system
+    localUncompressedURLs.cleanup()
+    localUncompressedURLs = []
+    downloadedPackages = []
+  }
+}
+
+protocol ExposureManagerDebuggable {
```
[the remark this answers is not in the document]
emilianojuan: the problem there is that we use private ExposureManager properties, I can change their access to public (because fileprivate wont be enough)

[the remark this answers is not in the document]
mattThousand: hm, I think I'd like us to do that, just to avoid clutter in this file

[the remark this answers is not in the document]
emilianojuan: done!

PR #199, opened 2020-08-13
(Android) Upgrade realm dependency

juancruzgs:
#### Description:

Upgrade Realm version to `7.0.1`. Changelog [here](https://github.com/realm/realm-java/blob/master/CHANGELOG.md)
Removed [Stetho](https://github.com/uPhyca/stetho-realm) since it doesn't support the latest Realm version and it seems it's no longer maintained.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -6,7 +6,6 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
 
     <uses-permission android:name="android.permission.BLUETOOTH" />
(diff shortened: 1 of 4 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,6 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
 
     <uses-permission android:name="android.permission.BLUETOOTH" />
-    <uses-permission android:name="android.permission.INTERNET" />
```
juancruzgs, line 9: This permission was duplicated

PR #201, opened 2020-08-13
Add empty state text to exposure list debug screen

amandabeiner:
#### Description:
Why:
As a tester, I want to be clearly told when there are no exposures to
view in the debug menu, so that I know that it is working as expected.

This commit:
Adds empty state text to alert a debug menu user that when the device
has no current exposures. It also adds a test for this screen, and
refactors the debug screen to use the exposure information that we are
already storing in app context instead of re-fetching them.

#### Linked issues:

Fixes: [Trello](uhttps://trello.com/c/fQzc2wFR/298-as-a-tester-when-i-go-to-the-show-exposures-debug-screen-and-there-are-no-exposures-i-see-some-no-exposures-indicatorrl)

#### Screenshots:

Simulator Screen Shot - iPhone 11 - 2020-08-13 at 12 48 55
Simulator Screen Shot - iPhone 11 - 2020-08-13 at 13 34 29

#### How to test:
- Open the app and clear exposures
- visit the "See exposures" screen
- Expect to see "no exposures" text
- Simulate exposure
- Return to the "See exposures" screen
- Expect to see the new simulated exposure

### DIFF
--- a/src/More/ExposureListDebugScreen.spec.tsx
+++ b/src/More/ExposureListDebugScreen.spec.tsx
@@ -1,0 +1,43 @@
+import React from "react"
+import { render } from "@testing-library/react-native"
+import "@testing-library/jest-native/extend-expect"
+import ExposureListDebugScreen from "./ExposureListDebugScreen"
+import { ExposureContext } from "../ExposureContext"
+import { factories } from "../factories"
+
+describe("ExposureListDebugScreen", () => {
+  describe("when there are no exposures", () => {
+    it("Displays a helpful message", () => {
+      const exposureProviderValue = factories.exposureContext.build()
+      const { getByText } = render(
+        <ExposureContext.Provider value={exposureProviderValue}>
+          <ExposureListDebugScreen />
+        </ExposureContext.Provider>,
+      )
+
+      expect(getByText("No exposure data to display")).toBeDefined()
+    })
+  })
+
+  describe("when there is one or more exposure", () => {
+    it("displays a list of exposures", () => {
+      const exposure1 = factories.exposureDatum.build({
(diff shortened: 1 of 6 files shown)


src/More/ExposureListDebugScreen.tsx
```diff
@@ -1,95 +1,41 @@
...
-  id: string
-  date: string
-}
-
-const ExposureListDebugScreen = ({
-  navigation,
-}: ENLocalExposureScreenProp): JSX.Element => {
-  const initialExposures: DebugExposure[] = []
-
-  const fetchExposures = async () => {
```
amandabeiner, line 34: Not sure if I'm missing something here, but is there a reason why we would want to fetch exposures independently in this screen when we are already using the same function to set current exposures in the `ExposureContext`? It seems like we already have the information

amandabeiner: Cool, yeah the function is already exposed through the context, so I can add that back into a useEffect hook. This would be easy to move to a button press in the future if we wanted!

src/More/ExposureListDebugScreen.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+        <ExposureContext.Provider value={exposureProviderValue}>
+          <ExposureListDebugScreen />
+        </ExposureContext.Provider>,
+      )
+
+      expect(getByText("No exposure data to display")).toBeDefined()
+    })
+  })
+
+  describe("whent here is one or more exposure", () => {
```
devinjameson, line 22:
```suggestion
  describe("when there is one or more exposure", () => {
```

PR #202, opened 2020-08-13
Achieve parity for iOS and Android design

devinjameson:
Why: as a user on either iOS or Android, I would like to have a comparable experience.

This commit:
- Updates a bunch of styling to achieve design parity

onboarding

main


src/Activation/ActivateProximityTracing.tsx
```diff
@@ -43,7 +43,7 @@ const ActivateProximityTracing: FunctionComponent = () => {
   }
 
   return (
-    <SafeAreaView>
+    <SafeAreaView style={{ backgroundColor: Colors.primaryBackground }}>
```
devinjameson, line 46: I think I like this pattern more than `style.safeArea` and then specifying `backgroundColor` in the styles. This will only ever take a `backgroundColor` so it feels more straightforward this way.

src/Onboarding/OnboardingScreen.tsx
```diff
@@ -155,23 +170,55 @@ const PositionDots: FunctionComponent<PositionDotsProps> = ({
   )
 }
 
+const isIOS = Platform.OS === "ios"
+const skipButtonSpacingTop = isIOS ? Spacing.xSmall : Spacing.xxLarge
```
devinjameson, line 174: Android handles `absolute` positioning differently from iOS in this particular circumstance. I think it's a bug. We need to do this to make the designs look the same.

src/Activation/ActivateProximityTracing.tsx
aledustet: My thoughts on this are that if we extract it to a style key, extending it implies, going to the key and adding the new style, while leaving it here, when it gets extended the person that changes is going to have to move it to the styles(create a variable if it doesn't exist) and then change this line. Also, consistency, if we always have it on the style object there are no opinions, just one pattern to follow, one less thing to consider.

src/Onboarding/OnboardingScreen.tsx
aledustet: How do you feel about using a `Platform.select`? When we use one of the two platforms it kind of feels like we treat one as the default and the other as the special case(guilty of that 😄  )

src/Activation/ActivateProximityTracing.tsx
devinjameson: Good points! Will update.

src/Onboarding/OnboardingScreen.tsx
devinjameson: Totally, good call.

src/Onboarding/OnboardingScreen.tsx
```diff
@@ -155,23 +166,69 @@ const PositionDots: FunctionComponent<PositionDotsProps> = ({
   )
 }
```
devinjameson, line 168: @aledustet I definitely like this pattern a lot more.

src/Onboarding/OnboardingScreen.tsx
```diff
@@ -155,23 +166,69 @@ const PositionDots: FunctionComponent<PositionDotsProps> = ({
...
 
+const skipButtonSpacingTop = Platform.select({
```
devinjameson, line 169: I def like this pattern a lot more @aledustet

PR #204, opened 2020-08-13
Add Bluetooth and Proximity Tracing info screens to Home stack

devinjameson:
Why: as a user, I would like to be able to learn more about how Bluetooth and Proximity Tracing work.

This commit:
- Creates a `Home` stack which includes the `Home` screen as well as the new `BluetoothInfo` and `ProximityTracingInfo` screens
- Adds buttons to navigate from the Home screen these two new info screens

infoScreens


src/navigation/HomeStack.tsx
```diff
@@ -0,0 +1,46 @@
...
+  TransitionPresets,
+} from "@react-navigation/stack"
+
+import { HomeScreens, HomeScreen } from "./index"
+import Home from "../Home/Home"
+import BluetoothInfo from "../Home/BluetoothInfo"
+import ProximityTracingInfo from "../Home/ProximityTracingInfo"
+
+type HomeStackParams = {
+  [key in HomeScreen]: undefined
```
johnschoeman, line 13: why is this undefined?

devinjameson: I think because none of these screens take params. But actually this can just be removed altogether, so I'll do that.

PR #214, opened 2020-08-17
Update Affected User Flow styling

devinjameson:
Why: the styling of the affected user flow has changed.

This commit:
- Updates styling throughout the affected user flow.


src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -116,61 +124,86 @@ const CodeInputForm: FunctionComponent = () => {
 
   const isDisabled = code.length !== codeLength
 
+  const codeInputBorderStyle = isFocused
+    ? { borderColor: Colors.secondaryViolet }
+    : null
+  const codeInputStyle = { ...style.codeInput, ...codeInputBorderStyle }
```
amandabeiner, line 130: This is minor, but it feels a little difficult to follow having style declarations up here and also down below in the `style` definition. What do you think about refactoring `styles` into a function that can take an argument like `codeInputIsFocused` and return the correct styles? Then up here you can just call `const styles = getStyles(isFocused)`, and have all of the styles being set in one place?

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -190,15 +223,29 @@ const indicatorWidth = 120
...
+    position: "absolute",
+    top: Layout.oneTwentiethHeight,
+    right: 0,
+  },
+  cancelButtonInnerContainer: {
+    padding: Spacing.medium,
   },
   headerContainer: {
-    marginBottom: Spacing.xxxHuge,
+    marginBottom: Spacing.xxLarge,
```
amandabeiner, line 248: 🥳 🦜

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
devinjameson: @achen178 and I paired on an improvement to this, but I had trouble understanding exactly what you're recommending here. I'm going to merge this in for now, but would love to pair on it later perhaps!

PR #216, opened 2020-08-17
Update example.env

johnschoeman:


example.env.bt
```diff
@@ -1,8 +1,9 @@
-TRACING_STRATEGY=bt
 DOWNLOAD_PATH=pc
 DISPLAY_SELF_ASSESSMENT=true
+PRIVACY_POLICY_URL='https://pathcheck.org/privacy-policy/'
 
 DISPLAY_NAME='PathCheck'
+IN_APP_NAME='PathCheck'
```
aledustet, line 6: I'm removing this oops 🤣

PR #217, opened 2020-08-17
Remove eula from activation stack

johnschoeman:
Why:
We would like to suppress the EULA for now until the product
requirements for it can be fully defined.

This commit:
Refactors the activation stack and removes the eula screen. We kept the
eula component in the repo as we expect to be reintroducing it in an
upcoming iteration.

### DIFF
--- a/src/Activation/AcceptEula.tsx
+++ b/src/Activation/AcceptEula.tsx
@@ -70,6 +70,10 @@ const AcceptEula: FunctionComponent = () => {
     loadEula()
   }, [eulaPath])
 
+  const handleOnPressNext = () => {
+    navigation.navigate(ActivationScreens.ActivateProximityTracing)
+  }
+
   const checkboxIcon = boxChecked
     ? Icons.CheckboxChecked
     : Icons.CheckboxUnchecked
@@ -110,9 +114,7 @@ const AcceptEula: FunctionComponent = () => {
           </GlobalText>
         </TouchableOpacity>
         <Button
-          onPress={() =>
-            navigation.navigate(ActivationScreens.ActivateProximityTracing)
-          }
+          onPress={handleOnPressNext}
           disabled={!boxChecked}
           label={t("common.continue")}
         />
--- a/src/navigation/ActivationStack.tsx
+++ b/src/navigation/ActivationStack.tsx
@@ -11,7 +11,6 @@ import { useTranslation } from "react-i18next"
 import { GlobalText } from "../components"
 import { Stacks, ActivationScreen, ActivationScreens } from "./index"
 
-import AcceptEula from "../Activation/AcceptEula"
 import ActivateProximityTracing from "../Activation/ActivateProximityTracing"
 import NotificationPermissions from "../Activation/NotificationPermissions"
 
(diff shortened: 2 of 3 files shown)


src/Activation/AcceptEula.tsx
```diff
@@ -70,6 +70,10 @@ const AcceptEula: FunctionComponent = () => {
     loadEula()
   }, [eulaPath])
 
+  const handleOnPressNext = () => {
+    navigation.navigate(ActivationScreens.ActivateProximityTracing)
```
devinjameson, line 74: Does this want to navigate to the next screen in the stack somehow? Instead of having this screen hardcoded?

devinjameson: Same with the navigation from `ActivateProximityTracing` to the next screen.

aledustet: I see where you're coming from @devinjameson you mean like having a navigate next property on each one of the onboarding/activation screen props? and then we could call navigate with the next screen prop? If this is owned by each screen and it's reused I would be onboard.

src/navigation/ActivationStack.tsx
```diff
@@ -24,35 +23,42 @@ type ActivationStackParams = {
...
+}
+
+const activationStepsIOS: ActivationStep[] = [
+  activateProximityTracing,
+  notificationPermissions,
+]
+
+const activationStepsAndroid: ActivationStep[] = [activateProximityTracing]
+
+const activationSteps =
```
aledustet, line 48: I'm a bit concerned about having logic among the declarations of this file, is there another place we could do this?

src/navigation/ActivationStack.tsx
```diff
@@ -24,35 +23,42 @@ type ActivationStackParams = {
...
-          return t("onboarding.step", { currentStep: 1, totalSteps })
-      }
-    }
-
+  const HeaderRight = (currentStep: number, totalSteps: number) => {
     return (
       <View style={style.headerRight}>
         <GlobalText style={style.headerRightText}>
-          {determineStepText()}
+          {t("onboarding.step", { currentStep, totalSteps })}
```
aledustet, line 61: probably not important but should we have an `activation` namespace for this?

src/Activation/AcceptEula.tsx
johnschoeman: sounds like a great idea to me.

src/navigation/ActivationStack.tsx
johnschoeman: i agree its a bit weird, we can look for a better spot.

src/Activation/AcceptEula.tsx
johnschoeman: though i do want to be careful not to over engineer this. We are working with react-navigation after all, we likely wont be able to get to only one call site for routing given the lack of a schema.

src/navigation/ActivationStack.tsx
johnschoeman: To keep this pr small, gonna roll back the `firstActivationScreen` concept, which will allow this to live in the component. We lose the single source of truth for the route progression, but we can revisit handling the routing in a more thoughtful way in a future commit.

src/navigation/ActivationStack.tsx
johnschoeman: agree, but also going to leave for a future commit.

PR #218, opened 2020-08-17
Refactor app colors

devinjameson:
Why: the way we were had implemented color variables was challenging to work with and resulted in confusion and a lack of parity between the codebase and the style guide.

This commit:
- Refactors colors based on the style guide from the design team
- Removes layers of abstraction from `colors.ts`
- Changes our color names to use a different format. Instead of:

```
const shamrockGreen = "#d8f9c3"
const primaryGreen = shamrockGreen
```

we are now using:

```
const success100 = "#d8f9c3"
```


ios/Podfile.lock
```diff
@@ -12,12 +12,12 @@ PODS:
     - React-Core (= 0.63.2)
```
johnschoeman, line 12: Probably dont want to include this one?

devinjameson: How do you mean?

PR #221, opened 2020-08-18
Send AppCenter destinations as a parameter

juancruzgs:
#### Description:

QA team is not getting new builds because the group is not added as a destination in each release.
This PR allows us to configure the `destinations` array from Bitrise


juancruzgs:
Merge on hold until we upgrade the plan in Bitrise

PR #223, opened 2020-08-18
Use HA specific content on About and Legal screens

devinjameson:
Why: prior to this commit the About and Legal screens were using placeholder content.

aboutlegal


src/More/About.tsx
```diff
@@ -18,9 +12,11 @@ export const AboutScreen: FunctionComponent = () => {
   const { t } = useTranslation()
 
   const osInfo = `${Platform.OS} v${Platform.Version}`
-  const pathCheckWebAddress = "pathcheck.org"
-  const pathCheckUrl = "https://pathcheck.org/"
   const { applicationName, versionInfo } = useApplicationInfo()
+  const healthAuthorityName =
```
devinjameson, line 16: @aledustet Will all this stuff be wrapped up in the config context? It's starting to feel pretty brittle.

aledustet: Definitely, this all os a candidate, for this screen however there might be changes coming soon but we should have this done soon.

src/More/Licenses.tsx
```diff
@@ -20,12 +20,16 @@ const Licenses: FunctionComponent = () => {
...
-  const pathCheckUrl = "https://pathcheck.org/"
   const privacyPolicyUrl =
     env.PRIVACY_POLICY_URL || "https://pathcheck.org/privacy-policy/"
+  const healthAuthorityName =
+    env.GAEN_AUTHORITY_NAME || "PathCheck Organization"
+  const haAddressLine1 = env.GAEN_AUTHORITY_ADDRESS_LINE_1 || "58 Day Street"
+  const haAddressLine2 = env.GAEN_AUTHORITY_ADDRESS_LINE_2 || "Box 441621"
+  const haAddressLine3 =
+    env.GAEN_AUTHORITY_ADDRESS_LINE_3 || "Somerville, MA 02144"
+  const haAddressLine4 = env.GAEN_AUTHORITY_ADDRESS_LINE_4 || "USA"
```
aledustet, line 31: I'm kind of worried about this but we will not solve international addresses when the moment comes that a different country won't have all the components we have here. Probably this ship sailed but did we consider having one address with `\n` to format it?

devinjameson: Yeah we're actually removing all this stuff anyways but that's a good point!

PR #228, opened 2020-08-18
Add button to change app language on home screen

devinjameson:
Why: as a user, I would like to be able to change the app language from the home screen.

This commit:
- Adds a button to change the language to the home screen
- Updates the home screen styling to accommodate this button

NOTE: When the language is changed to Spanish, the header text is too long and so the description text gets cut off. This is a separate issue. @johnschoeman You may want to know about this given that it's part of the translations conversation we were having earlier.

language


src/Home/Home.tsx
```diff
@@ -118,7 +123,24 @@ const HomeScreen: FunctionComponent = () => {
 
   const bottomContainerStyle = {
     ...style.bottomContainer,
-    maxHeight: insets.bottom + Layout.screenHeight * 0.475,
+    maxHeight: Platform.select({
```
devinjameson, line 126: It seems like there are a bunch of "magic numbers" here, but this is the only way I can think off to make this design work. It gets really complex when you're working with different screen sizes and supporting iOS + Android, plus we have this background image, absolutely positioned elements, etc.

johnschoeman: one thing you can consider is using named constants. for example:
```typescript
const maxHeightIOS = insets.bottom + Layout.screenHeight * 0.475
const maxHeightAndroid = insets.bottom + Layout.screenHeight * 0.475 - 50
```

while still magic number-ish you have at least provided the reader a bit of context on the reasoning.

devinjameson: Good call!

PR #230, opened 2020-08-19
(iOS) Differentiate between dev and prod revision tokens

mattThousand:
### Why 
We need to securely preserve the revision token returned from the verification server across app installs, and for testing purposes, we need apps with the same bundle id to differentiate a verification code for use with the dev server against a verification code for use with the production server.

### This Commit
This commit creates a service that allows for storing the revision token in a bundle-id specific keychain per server (dev or production)


ios/BT/Storage/KeychainService.swift
```diff
@@ -0,0 +1,25 @@
+import KeychainAccess
+
+final class KeychainService {
+
+  static let shared = KeychainService()
+
+  private lazy var keychain = Keychain(service: "\(Bundle.main.bundleIdentifier!).keychainService")
```
aledustet, line 7: I see, so we will be storing the token on a combination of bundle identifier and then dev and prod environments. Sounds good but it seems to me that we will still have the problem if we change the server URL on the environment variables while we are at the dev environment we will run into the same issue. Is it possible to store the key using a combination of the bundle identifier and the exposures server? Could we store the tokens in some sort of dictionary where the keys are the exposure server URL and that could be persisted per install? Again not bulletproof but will mimic the behavior that is being problematic to us which is using the same keys, on the same server with the wrong revision token.

PR #233, opened 2020-08-19
Make onboarding scrollable

zhaccc:
#### Description:

The onboarding screen was not scrollable on smaller mobile devices so we made this fix.

### DIFF
--- a/src/Onboarding/OnboardingScreen.tsx
+++ b/src/Onboarding/OnboardingScreen.tsx
@@ -68,26 +68,28 @@ const OnboardingScreen: FunctionComponent<OnboardingScreenProps> = ({
     <>
       <SafeAreaView style={style.topSafeArea} />
       <SafeAreaView style={style.bottomSafeArea}>
-        <TouchableOpacity
-          onPress={() => navigation.navigate(Screens.LanguageSelection)}
-          style={style.languageButtonContainer}
-        >
-          <GlobalText style={style.languageButtonText}>
-            {languageName}
-          </GlobalText>
-        </TouchableOpacity>
-        <TouchableOpacity
-          onPress={() =>
-            navigation.navigate(Stacks.Activation, {
-              screen: ActivationScreens.AcceptEula,
-            })
-          }
-          style={style.skipButtonContainer}
-        >
-          <GlobalText style={style.skipButtonText}>
-            {t("common.skip")}
-          </GlobalText>
-        </TouchableOpacity>
+        <View style={style.header}>
+          <TouchableOpacity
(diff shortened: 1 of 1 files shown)


src/Onboarding/OnboardingScreen.tsx
```diff
@@ -187,6 +185,11 @@ const style = StyleSheet.create({
...
     backgroundColor: Colors.secondary10,
   },
+  header: {
+    flex: 0.1,
+    flexDirection: "row",
+    justifyContent: "space-between",
+  },
   languageButtonContainer: {
     ...Outlines.ovalBorder,
     position: "absolute",
```
zhaccc, line 195: I think we should avoid `absolute` positioning this will break the styles across the devices. Maybe we should open PR to fix this?

devinjameson:
Overall I think this code looks good! I wonder if we want to make this scrollable or just make the content small enough so that it fits on any screen size. Perhaps there is a way to make the height of the image conditional on the screen height? May be worth checking with Art on the design team.

src/Onboarding/OnboardingScreen.tsx
```diff
@@ -170,13 +172,15 @@ const skipButtonSpacingTop = Platform.select({
...
 const languageButtonSpacingTop = Platform.select({
   ios: Spacing.small,
   android: Spacing.huge,
 })
-const imageMarginTop = Platform.select({
-  ios: 60,
-  android: 90,
+
+const headerFlex = Platform.select({
+  ios: 0.1,
```
devinjameson, line 182: A pattern we've been using elsewhere is `Layout.screenHeight * 0.1` or `Layout.oneTenthHeight`. That may not work for this situation, but wanted to make you aware of it just in case.

devinjameson:
> I wonder if we want to make this scrollable or just make the content small enough so that it fits on any screen size. Perhaps we could make the height of the image conditional on the screen height? May be worth checking with Art on the design team

@zhaccc I take this back actually, considering that we have to support very large text sizes for accessibility.

zhaccc:
> Overall I think this code looks good! I wonder if we want to make this scrollable or just make the content small enough so that it fits on any screen size. Perhaps we could make the height of the image conditional on the screen height? May be worth checking with Art on the design team.

I agree, but let's just merge this for now because it quickly hotfixes this particular problem.

src/Onboarding/OnboardingScreen.tsx
zhaccc: Oh, thanks! Yeah, won't work in this particular case but this is good to know!

PR #234, opened 2020-08-19
Tweak home screen design issues

achen178:
Why:
The home screen currently has two notable design inconsistencies. They were fixed In order to have a more fluid and consistent user experience. 

This commit:
- Changes the main inactive state image on the home screen to match the colors within the app
- Adjusts the layout of the right icons in order to have better visual consistency

#### Before:
Simulator Screen Shot - iPhone 8 - 2020-08-19 at 11 23 30

#### After:
Simulator Screen Shot - iPhone 8 - 2020-08-19 at 11 23 19


src/Home/Home.tsx
```diff
@@ -377,7 +377,9 @@ const style = StyleSheet.create({
     ...Typography.header4,
   },
   shareIconContainer: {
-    paddingHorizontal: Spacing.medium,
+    width: 60,
```
devinjameson, line 380: Thoughts on this living in a variable? Something like `const activationStatusRightWidth = 60` right above the stylesheet?

achen178: Hmm, yeah I can see that -- I'll make the change

PR #238, opened 2020-08-19
Refactor PermissionsContext

devinjameson:
Why: prior to this commit, the PermissionsContext was exposing the `ENPermissionStatus` directly, which has a type of `[ENAuthorization, ENEnablement]`, which in turn have types:

```
type ENAuthorization = `UNAUTHORIZED` | `AUTHORIZED`
type ENEnablement = `DISABLED` | `ENABLED`
```

This resulted in complexity at the view layer because we had to import various types from the `PermissionsContext` and, for example, check to see if the `ENAuthorization === "AUTHORIZED"` or `ENEnablement === "DISABLED"`.

This commit instead exposes an `ENAuthorizationEnablementStatus`, of type `{ authorization: boolean, enablement: boolean }`.

The result of this is that we can check if exposure notifications are authorized with `exposureNotifications.status.authorization`, or if they are enabled with `exposureNotifications.status.enablement`.

### DIFF
--- a/src/Home/Home.spec.tsx
+++ b/src/Home/Home.spec.tsx
@@ -40,14 +40,14 @@ jest.mock("../More/useApplicationInfo", () => {
 jest.mock("./useBluetoothStatus.ts")
 
 describe("Home", () => {
-  describe("When the exposure notification permissions are enabled and authorized and Bluetooth is on", () => {
+  describe("When the exposure notification permissions are enabled and the app is authorized and Bluetooth is on", () => {
     it("renders an active message", async () => {
       const isBluetoothOn = true
       ;(useBluetoothStatus as jest.Mock).mockReturnValue(isBluetoothOn)
 
       const isENAuthorizedAndEnabled: ENAuthorizationEnablementStatus = {
-        authorization: true,
-        enablement: true,
+        authorized: true,
+        enabled: true,
       }
       const permissionProviderValue = createPermissionProviderValue(
         isENAuthorizedAndEnabled,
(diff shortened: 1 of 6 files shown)


src/Home/Home.spec.tsx
```diff
@@ -37,14 +40,17 @@ jest.mock("../More/useApplicationInfo", () => {
 jest.mock("./useBluetoothStatus.ts")
 
 describe("Home", () => {
-  describe("When the enPermissionStatus is enabled and authorized and Bluetooth is on", () => {
+  describe("When the exposure notification permissions are enabled and authorized and Bluetooth is on", () => {
```
johnschoeman, line 43:
```suggestion
  describe("When the exposure notification permissions are enabled and the app is authorized and Bluetooth is on", () => {
```

src/Home/Home.spec.tsx
```diff
@@ -80,17 +86,17 @@ describe("Home", () => {
     })
   })
 
-  describe("When the enPermissionStatus is not enabled and not authorized", () => {
+  describe("When the exposure notification permissions are not enabled and not authorized", () => {
```
johnschoeman, line 89:
```suggestion
  describe("When the exposure notification permissions are not enabled and the app is not authorized", () => {
```

src/PermissionsContext.tsx
```diff
@@ -13,15 +13,32 @@ import {
...
+  "DISABLED",
+]
+
+const toAuthorizationEnablementStatus = (
+  enPermissionStatus: ENPermissionStatus,
+): ENAuthorizationEnablementStatus => {
+  const isENAuthorized = enPermissionStatus[0] === "AUTHORIZED"
+  const isENEnabled = enPermissionStatus[1] === "ENABLED"
+
+  return { authorization: isENAuthorized, enablement: isENEnabled }
```
johnschoeman, line 32: what do you think about
```suggestion
  return { authorized: isENAuthorized, enabled: isENEnabled }
```
?

devinjameson: Yeah this is better! Naming is hard.

src/Home/Home.spec.tsx
devinjameson: 👍

src/Home/Home.spec.tsx
devinjameson: 👍

PR #239, opened 2020-08-20
Add configuration context

aledustet:
Why:
----
The configuration logic needs several safeguards on the way it is being used across the application, from defensive programming to decisions on how to consume it. Some of the operations are changing rapidly and it is hard to keep up. Also, the `react-native-config` package is very hard to test properly. To sum up:

We need a way to abstract the way we get configuration items for the application. This will be an item subject to changes in the next iterations. A few problems we are facing:

- The complexity of "interpreting" environment is scattered across the codebase
- Some logic is leaking from the environment into the business logic
- If any environment variable triggered a change in behavior testing that becomes cumbersome
- Lack of types from environment variables, everything is a string

How:
----
This is attempting to move that logic from just reading string variables to moving the repeated logic decisions into a domain we can control. Right now is just decisions on what platform dictates what environment variable to load, but if this is a pattern we'll keep using is best to move it to a "testable" and more controlled area. We also anticipate that more "decisions" will be made in the load phase of a configuration.

Caveats:
----
- The `helpers` are not included in this first pass since it would involve passing arguments since these are not components. Still working through the best way to send these configs only used in helpers. Comments welcome!
- We are still missing validation of config for a given HA during build time. That is coming in subsequent PR's as we work through the best way of doing this on CI and Fastlane.

This Commit:
----
- Migrate Home screen to use the new exposure context
- Add region codes and app package name to context
- Move usage of env on about and legal screen
- Move usage of env on the main tab navigator
- Renamed variables to include `helathAuthority`
- Move usage of env on More screens to use configuration context
- Remove env from ExposureHistory screens to use the context
- Remove usage of env from Share screen on self assessment
- Transform self assessment share to typescript

### DIFF
--- a/src/ConfigurationContext.tsx
+++ b/src/ConfigurationContext.tsx
@@ -1,0 +1,73 @@
+import React, { FunctionComponent, createContext, useContext } from "react"
+import { Platform } from "react-native"
+import env from "react-native-config"
+
+export interface Configuration {
+  appDownloadLink: string
+  appPackageName: string
+  displayReportAnIssue: boolean
+  displaySelfAssessment: boolean
+  healthAuthorityAdviceUrl: string
+  healthAuthorityName: string
+  healthAuthorityPrivacyPolicyUrl: string
+  regionCodes: string[]
+}
+
+const initialState = {
+  appDownloadLink: "",
+  appPackageName: "",
+  displayReportAnIssue: false,
+  displaySelfAssessment: false,
+  healthAuthorityAdviceUrl: "",
+  healthAuthorityName: "",
+  healthAuthorityPrivacyPolicyUrl: "",
+  regionCodes: [],
+}
+
+const ConfigurationContext = createContext<Configuration>(initialState)
+
+const ConfigurationProvider: FunctionComponent = ({ children }) => {
+  const {
+    AUTHORITY_ADVICE_URL: healthAuthorityAdviceUrl,
+    GAEN_AUTHORITY_NAME: healthAuthorityName,
+    PRIVACY_POLICY_URL: healthAuthorityPrivacyPolicyUrl,
+  } = env
+  const displaySelfAssessment = env.DISPLAY_SELF_ASSESSMENT === "true"
+  const displayReportAnIssue = env.DISPLAY_REPORT_AN_ISSUE === "true"
+  const appDownloadLink = Platform.select({
+    ios: env.IOS_APP_STORE_URL,
+    android: env.ANDROID_PLAY_STORE_URL,
+  }) as string
+  const appPackageName = Platform.select({
+    ios: env.IOS_BUNDLE_ID,
+    android: env.ANDROID_APPLICATION_ID,
+  }) as string
+  const regionCodes = env.REGION_CODES.split(",")
+
+  return (
+    <ConfigurationContext.Provider
+      value={{
+        appDownloadLink,
+        appPackageName,
+        displayReportAnIssue,
+        displaySelfAssessment,
+        healthAuthorityAdviceUrl,
+        healthAuthorityName,
+        healthAuthorityPrivacyPolicyUrl,
+        regionCodes,
+      }}
+    >
+      {children}
+    </ConfigurationContext.Provider>
+  )
+}
+
+const useConfigurationContext = (): Configuration => {
+  const context = useContext(ConfigurationContext)
+  if (context === undefined) {
+    throw new Error("ConfigurationContext must be used with a provider")
+  }
+  return context
+}
+
+export { ConfigurationContext, ConfigurationProvider, useConfigurationContext }
--- a/src/ExposureHistory/History/NoExposures.tsx
+++ b/src/ExposureHistory/History/NoExposures.tsx
@@ -1,5 +1,4 @@
 import React, { FunctionComponent } from "react"
-import env from "react-native-config"
 import { Linking, View, StyleSheet, TouchableOpacity } from "react-native"
 import { SvgXml } from "react-native-svg"
 import { useTranslation } from "react-i18next"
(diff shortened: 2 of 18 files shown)


src/Home/Home.spec.tsx
```diff
@@ -37,6 +39,32 @@ jest.mock("../More/useApplicationInfo", () => {
 jest.mock("./useBluetoothStatus.ts")
 
 describe("Home", () => {
+  it("allows users to share the application", () => {
```
aledustet, line 42: This is an example on how we can improve testing logic dependant on the values from the environment

src/ConfigurationContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+  authorityName: "",
+  authorityAdviceUrl: "",
+}
+
+const ConfigurationContext = createContext<Configuration>(initialState)
+
+const DEFAULT_DOWNLOAD_LINK = ""
+const DEFAULT_AUTHORITY_PRIVACY_URL = ""
+const DEFAULT_AUTHORITY_NAME = ""
+const DEFAULT_AUTHORITY_ADVICE_URL = ""
```
aledustet, line 28: This is another benefit of controlling the environment here, we could:
- Display warnings on inconsistent configurations in a dev environment
- Have a safeguard for miss-configured builds, so no blowing up if something is missing while we fix it.

aledustet: Maybe we don't need a default since these configs should be vetted in advance but the paranoid in me likes this.

src/ConfigurationContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const ConfigurationProvider: FunctionComponent = ({ children }) => {
```
aledustet, line 30: Another benefit of having this here is that we could pivot, or enrich our use of the config package with external or a static file we can set on our setup scripts

src/ConfigurationContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const authorityPrivacyPolicyUrl =
```
devinjameson, line 31: Do you think this would be better or worse? Seems like it might get tough to read as this grows.

```
  const {
    PRIVACY_POLICY_URL: authorityPrivacyPolicyUrl = DEFAULT_AUTHORITY_PRIVACY_URL,
    GAEN_AUTHORITY_NAME: authorityName = DEFAULT_AUTHORITY_NAME,
    AUTHORITY_ADVICE_URL: authorityAdviceUrl = DEFAULT_AUTHORITY_ADVICE_URL,
  } = env

  const displaySelfAssessment = env.DISPLAY_SELF_ASSESSMENT === "true"
  const displayReportAnIssue = env.DISPLAY_REPORT_AN_ISSUE === "true"

  const appDownloadLink =
    Platform.select({
      ios: env.IOS_APP_STORE_URL,
      android: env.ANDROID_PLAY_STORE_URL,
    }) || DEFAULT_DOWNLOAD_LINK

```

src/ConfigurationContext.tsx
johnschoeman: I dig this a lot. It would also be nice to see some build time checking as well though, if we dont have a valid config file it would be nice to fail at compile time.

src/ConfigurationContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+    env.PRIVACY_POLICY_URL || DEFAULT_AUTHORITY_PRIVACY_URL
+  const authorityName = env.GAEN_AUTHORITY_NAME || DEFAULT_AUTHORITY_NAME
```
johnschoeman, line 33: what do you think about be less liberal with the defaults? if we fail to provide an authority name, would we rather the app fail to build than to accidentally ship a `PathCheck`?

src/ConfigurationContext.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { FunctionComponent, createContext, useContext } from "react"
+import { Platform } from "react-native"
+import env from "react-native-config"
+
+export interface Configuration {
```
johnschoeman, line 5: 💯

src/ConfigurationContext.tsx
aledustet: HMW fail here? When? Is this a case of yeah this builds and then when I press play we show an error? Do we have any tests running after an environment is added?

johnschoeman: What do you think about a `bin/config_is_valid.sh` script that is run before builds on ci fastlane?  If its fast we might also want to add it to `yarn validate`

johnschoeman: so not a perfect solution but, my main concern is we are setting ourselves up to forget to update a config somewhere and ship a build that runs but shows the wrong app name or the like, which would not be a great look.

aledustet: 100% agree with you on this one, my only concern is where to do this. Since it would only make sense to validate prior-creating an android or ios build. The binaries we have right now for scripting are made to be run on demand when someone is about to cut a build but we don't necessarily need those before each build. Also, those live far from the logic of the application.

aledustet: Should we do it on the place we store the configs?

johnschoeman: that could make sense as well.  dont have a strong opinion, but leaning towards keeping all the ci in one spot.

src/ExposureHistory/History/NoExposures.tsx
```diff
@@ -31,16 +27,21 @@ const NoExposures: FunctionComponent = () => {
...
+    Linking.openURL(healthAuthorityAdviceUrl)
   }
 
   return (
     <View style={style.card}>
       <GlobalText style={style.cardHeaderText}>
         {t("exposure_history.protect_yourself_and_others")}
       </GlobalText>
-      {Boolean(healthAuthorityLink) && (
+      {Boolean(healthAuthorityAdviceUrl) && (
```
devinjameson, line 44: Does this need the Boolean constructor?

PR #242, opened 2020-08-20
Add scaffolding for Analytics work

amandabeiner:
#### Description:
Why:
As a product team, I want to be sure that the analytics information I am
collecting respects user privacy and jurisdictional restrictions.

This commit:
Introduces an AnalyticsContext which will will access information
about a user's analytics settings. Right now this is just whether they
have consented to send us analytics, but in the future this will include
more granular privacy settings (ability to consent to some analytics but
not others), and blanket restrictions for jurisdictions (such as HAs in
the EU opting out of analytics altogether)

It also introduces an analytics module, which will serve as a source of
truth for what kinds of events we support, and how to communicate with
the server to fire off those events. This is pretty empty right now, and
is currently serving only to test that we are only tracking analytics
when we have permission to do so.

Note that this code is currently not in use anywhere. We will build off
of it once we finalize tracking requirements with product and the server
team.

#### Linked issues:

Fixes: [Trello](https://trello.com/c/U80GlN4m/398-as-a-product-team-i-want-to-be-sure-that-we-are-capturing-analytics-in-accordance-with-user-consent-and-the-jurisdiction-laws)

#### How to test:

- The app should compile
- Tests should pass


src/analytics/index.ts
```diff
@@ -0,0 +1,3 @@
+import * as actions from "./actions"
```
amandabeiner, line 1: This will soon also house `events` so that we can have type safety around what event strings can be sent. For now, this module is pretty empty, but separating it out makes it easier to check that it's being called as expected in the tests.

src/utils/storage.ts
```diff
@@ -47,3 +47,36 @@ export async function setIsOnboardingComplete(): Promise<void> {
...
+export async function getAnalyticsConsent(): Promise<boolean> {
+  const userConsented = await getStoreData(ANALYTICS_CONSENT)
+  return consentToBoolean(userConsented)
+}
+
+export async function setAnalyticsConsent(consent: boolean): Promise<void> {
+  return setStoreData(ANALYTICS_CONSENT, booleanToConsent(consent))
+}
+
+const consentToBoolean = (consent: string | null): boolean => {
```
aledustet, line 64: 🔥

aledustet:
Not sure why the CI is not kicking in, happened the same to me 🤷‍♂️ might need a force push

PR #243, opened 2020-08-20
Update typography across the app

devinjameson:
Why: our typography styles were hard to use, not consistent within the app, and not in sync with our style guide.

This commit: 
- Updates `typography.ts` to use new variables for typographic styles
- Uses these new typographic styles across the app
- Makes a fix for the OnboardingScreen header
- Removes some button styling that is no longer used

### DIFF
--- a/src/Onboarding/OnboardingScreen.tsx
+++ b/src/Onboarding/OnboardingScreen.tsx
@@ -83,7 +83,6 @@ const OnboardingScreen: FunctionComponent<OnboardingScreenProps> = ({
                 screen: ActivationScreens.AcceptEula,
               })
             }
-            style={style.skipButtonContainer}
           >
             <GlobalText style={style.skipButtonText}>
               {t("common.skip")}
@@ -168,20 +167,7 @@ const PositionDots: FunctionComponent<PositionDotsProps> = ({
   )
 }
 
-const skipButtonSpacingTop = Platform.select({
-  ios: Spacing.xSmall,
-  android: Spacing.xxLarge,
-})
-
-const languageButtonSpacingTop = Platform.select({
-  ios: Spacing.small,
-  android: Spacing.huge,
-})
-
-const headerFlex = Platform.select({
-  ios: 0.1,
-  android: 0.15,
-})
+const headerHeight = Platform.select({ ios: 65, android: 90, default: 65 })
 
 const style = StyleSheet.create({
   topSafeArea: {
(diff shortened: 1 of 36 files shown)


src/navigation/MainTabNavigator.tsx
```diff
@@ -25,7 +25,7 @@ const MainTabNavigator: FunctionComponent = () => {
   const { userHasNewExposure } = useExposureContext()
   const insets = useSafeAreaInsets()
   const displaySelfAssessment = env.DISPLAY_SELF_ASSESSMENT === "true"
-  const displayReportAnIssue = env.DISPLAY_REPORT_AN_ISSUE === "true"
+  const displayReportAnIssue = env.DISPLAY_REPORT_AN_ISSUE === "true" || true
```
johnschoeman, line 28: should this be removed?

src/Onboarding/OnboardingScreen.tsx
```diff
@@ -168,20 +167,7 @@ const PositionDots: FunctionComponent<PositionDotsProps> = ({
...
-const languageButtonSpacingTop = Platform.select({
-  ios: Spacing.small,
-  android: Spacing.huge,
-})
-
-const headerFlex = Platform.select({
-  ios: 0.1,
-  android: 0.15,
-})
+const headerHeight = Platform.select({ ios: 65, android: 90, default: 65 })
```
johnschoeman, line 170: can we add names for these numbers?

devinjameson: I'm kinda on the fence here. Is it not clear enough as is?

devinjameson: Like, we could have:

```
const iosHeaderHeight = 65
const androidHeaderHeight = 65
const headerHeight = Platform.select({ ios: iosHeaderHeight, android: androidHeaderHeight, default: iosHeaderHeight )}
```

But I'm not sure I like that more.

johnschoeman: yeah a bit of overkill in this case, so totally your call and fine either way.

personally, I tend towards choosing to be black n white in gray areas as it tends to encourage better practices in the long term, so even though its kinda dumb to used a named constant for the number in this case, because using named constants for numbers is good idea in general, its makes sense to do it even in the 1 in 100 cases where its a bit too much.

devinjameson: That makes sense! I'm down with the change.

PR #244, opened 2020-08-20
Add bugsnag and logs to the key submission process

aledustet:
Why:
----
We are lacking visibility on the data flow during the submission of exposure keys. We need a way to gain insight on what are the errors, on which environment and what are the problems we are having. Namely which exchange is happening among all actors in the process:
- OS and application with keys
- Verification server and application
- Exposure Keys server and application

How:
----
By adding a logs service like `bugsnag` we not only have the ability to log the data flow among all the actors but also gain info on how the users are behaving and what other errors we might be missing.

This Commit:
----
- Install and setup `bugsnag`
- Add `Logger` service object to add relevant metadata to the session and send errors
- Add logs to the submission process and fix tests

Caveat:
----
This first implementation with bugsnag is not uploading symbol files, we'll address that on future iterations. There's a [guide] available.

[guide]: https://docs.bugsnag.com/platforms/react-native/react-native/showing-full-stacktraces/#uploading-source-maps

### DIFF
--- a/src/AffectedUserFlow/PublishConsent/PublishConsentScreen.spec.tsx
+++ b/src/AffectedUserFlow/PublishConsent/PublishConsentScreen.spec.tsx
@@ -1,3 +1,4 @@
+jest.mock("../../logger.ts")
 import React from "react"
 import { render, fireEvent } from "@testing-library/react-native"
 import { useNavigation } from "@react-navigation/native"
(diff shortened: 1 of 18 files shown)


src/AffectedUserFlow/PublishConsent/PublishConsentScreen.spec.tsx
```diff
@@ -1,3 +1,4 @@
+jest.mock("../../logger.ts")
```
johnschoeman, line 1: should this live below the imports? or does it need to be mocked first?

PR #246, opened 2020-08-20
Move post diagnosis keys to js layer

aledustet:
Why:
----
The logic to submit keys to the backend is duplicated on the native layer.
We would like to consolidate all the network calls to the js layer, in order to make all those calls easier to test, instrument and maintain. We need to display the errors generated on the servers, so it is no longer a generic error message being displayed.

This Commit:
----
- Add a context to temporarily hold exposure keys for publishing
- Remove the call to the submit diagnosis keys on the native layer
- Add the post diagnosis keys request to the exposureNotificationAPI module
- Add error handling and complete response logging


src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
```diff
@@ -28,35 +29,64 @@ import {
...
+        }
+      }
     } catch (e) {
       setIsLoading(false)
       const errorMessage = e.message
       Logger.addMetadata("publishKeys", {
         errorMessage,
       })
-      Logger.error("Incomplete Key Submission")
+      Logger.error(`Incomplete Key Submission ${errorMessage}`)
```
johnschoeman, line 89: we might want to format this something like `IncompleteKeySumbission.${errorMessage}` for error aggression and search reasons.

src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
```diff
@@ -28,35 +29,64 @@ import {
...
+        revisionToken,
+      )
       setIsLoading(false)
-      navigation.navigate(AffectedUserFlowScreens.AffectedUserComplete)
+      if (response.kind === "success") {
+        storeRevisionToken(response.body.revisionToken)
+        navigation.navigate(AffectedUserFlowScreens.AffectedUserComplete)
+      } else {
+        const errorMessage = response.message
+        Logger.addMetadata("publishKeys", {
```
aledustet, line 75: Do we want to add the revision token to this error log?

mattThousand: totally ^^

PR #251, opened 2020-08-21
(Android) Remove unused methods/classes/variables

juancruzgs:
#### Description:

Now that submission logic has been moved to JS layer there are some classes we don't need anymore.
This PR cleans up all the unused methods, classes & variables on the Android side.


mattThousand:
> @mattThousand should we do the same in iOS?

@emilianojuan yes 👍

PR #256, opened 2020-08-24
Add LinearGradient background to Welcome screen

devinjameson:
This commit:
- Adds a `GradientBackground` component
- Uses `GradientBackground` on the Welcome screen
- Uses `useAngle` wherever we use `LinearGradient` currently

Screen Shot 2020-08-24 at 12 11 08 PM

### DIFF
--- a/src/Onboarding/Welcome.tsx
+++ b/src/Onboarding/Welcome.tsx
@@ -5,13 +5,14 @@ import { useNavigation } from "@react-navigation/native"
 import LinearGradient from "react-native-linear-gradient"
 
 import { GlobalText } from "../components/GlobalText"
+import { Button } from "../components/Button"
+import { GradientBackground } from "../components"
 import { getLocalNames } from "../locales/languages"
-import { Button } from "../components"
 import { useApplicationName } from "../More/useApplicationInfo"
+import { Screens, OnboardingScreens, useStatusBarEffect } from "../navigation"
 
 import { Images } from "../assets"
-import { Spacing, Colors, Typography, Outlines } from "../styles"
-import { Screens, OnboardingScreens, useStatusBarEffect } from "../navigation"
+import { Layout, Spacing, Colors, Typography, Outlines } from "../styles"
 
 const Welcome: FunctionComponent = () => {
   const navigation = useNavigation()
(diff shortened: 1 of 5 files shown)


src/Onboarding/Welcome.tsx
```diff
@@ -23,42 +24,47 @@ const Welcome: FunctionComponent = () => {
...
-        <GlobalText style={style.languageButtonText}>{languageName}</GlobalText>
-      </TouchableOpacity>
-      <View>
-        <Image
-          source={Images.PeopleOnNetworkNodes}
-          style={style.image}
-          accessible
-          accessibilityLabel={t("onboarding.welcome_image_label")}
+    <>
+      <GradientBackground />
```
aledustet, line 28: Could we have this component render it's children? or would that mess with the layout?

src/components/GradientBackground.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { FunctionComponent } from "react"
+import { StyleSheet } from "react-native"
+
+import LinearGradient from "react-native-linear-gradient"
+
+import { Colors } from "../styles"
+
+const GradientBackground: FunctionComponent = () => {
```
johnschoeman, line 8: What do you think about keeping this inline until we have a few places where this is being used?

src/Onboarding/Welcome.tsx
devinjameson: That's what I tried first, and it would be ideal, but it messes up all the tests. For some reason all the `getBy` stuff doesn't work.

src/components/GradientBackground.tsx
devinjameson: Based on the Figma designs, it looks like we'll be using this in several other spots in the next iteration. Given that, I feel like it makes sense to leave as is.

src/Onboarding/Welcome.tsx
devinjameson: Can you think of a way around that @aledustet?

devinjameson: I'm going to get this merged in for now, but perhaps we can take a look at this once things cool down a bit for you @aledustet?

PR #257, opened 2020-08-24
(Android) Remove text and logo from splash screen

juancruzgs:
#### Description:

This PR removes the text "PathCheck BT" and the logo from Splash screen.
In the future each HA will have its own splash screen, but we can use this one as a placeholder.

**Old:**
image

**New:**
image


devinjameson:
Does this only remove the text on Android or does it do it for iOS as well?

juancruzgs:
Just for Android

devinjameson:
Just want to make sure this work is going to be handled on the iOS side as well @mattThousand

PR #259, opened 2020-08-24
Introduce staging configuration

mattThousand:
### Why
We'd like to suppress the debug menu for release builds, and also ensure that builds are not pointed to the release environment during testing

### This Commit
This commit introduces a `STAGING` env configuration that suppresses the debug menu and points to the dev server, as well as an `Adhoc` scheme for `iOS` that uses the `.env.bt` configuration

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -92,6 +92,8 @@ apply from: "../../node_modules/@bugsnag/react-native/bugsnag-react-native.gradl
 project.ext.react = [
     entryFile: "index.js",
     bundleInDebug: false,
+    bundleInStaging: true,
+    devDisabledInStaging: true,
     enableHermes: false,  // clean and rebuild if changing
 ]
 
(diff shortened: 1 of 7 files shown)


juancruzgs:
@mattThousand According to this [card](https://trello.com/c/vFqRhwLM/417-as-a-dev-team-when-we-make-a-release-build-end-users-do-not-see-debug-info-menus-etc) we shouldn't show the debug menu on staging builds , right?

android/app/build.gradle
```diff
@@ -6,6 +6,7 @@ apply plugin: 'realm-android'
 
 project.ext.envConfigFiles = [
     debug: ".env.bt",
+    staging: ".env.bt.staging",
```
juancruzgs, line 9: This won't compile I think, more changes are needed to add a new build variant
I can make those changes here

mattThousand: @juancruzgs thanks!

mattThousand:
> @mattThousand According to this [card](https://trello.com/c/vFqRhwLM/417-as-a-dev-team-when-we-make-a-release-build-end-users-do-not-see-debug-info-menus-etc) we shouldn't show the debug menu on staging builds , right?
> If it does show the menu, what's the difference between DEV and STAGING?

@juancruzgs thanks, you are right

PR #263, opened 2020-08-24
Updated android icons

aledustet:
- New android icons
- Changes the color


android/app/src/main/res/values/ic_launcher_background.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
juancruzgs, line 1: This file is not being used, it can be removed

PR #258, opened 2020-08-24
(Android) Disable bugsnag on release builds

juancruzgs:
#### Description:

This PR disables Bugsnag on release builds.

Bugsnag has the concept of "stages" to identify each environment. We will only notify Bugsnag when using the `development` stage.
https://docs.bugsnag.com/platforms/android/configuration-options/#releasestage
https://docs.bugsnag.com/platforms/android/configuration-options/#enabledreleasestages

I think this is enough to disable it and nothing else is needed on the JS side: 
https://docs.bugsnag.com/platforms/react-native/react-native/#further-configuration.

#### How to test:
If Bugsnag is enabled you will see logs like these:
```
2020-08-24 14:53:24.731 5831-6061/org.pathcheck.covidsafepaths.bt D/Bugsnag: Received NDK message com.bugsnag.android.StateEvent$StartSession@3d68367
2020-08-24 14:53:30.730 5831-5831/org.pathcheck.covidsafepaths.bt D/Bugsnag: Received NDK message com.bugsnag.android.StateEvent$AddBreadcrumb@b0691c9
2020-08-24 14:53:30.751 5831-5831/org.pathcheck.covidsafepaths.bt D/Bugsnag: Received NDK message com.bugsnag.android.StateEvent$AddBreadcrumb@72cbef
2020-08-24 14:53:31.020 5831-5831/org.pathcheck.covidsafepaths.bt D/Bugsnag: Received NDK message com.bugsnag.android.StateEvent$UpdateInForeground@8af66e8
```

That's the information that Bugsnag automatically captures and you won't see those logs if it's disabled.
https://docs.bugsnag.com/platforms/android/#logging-breadcrumbs
https://docs.bugsnag.com/platforms/android/#session-tracking

PR #264, opened 2020-08-24
Only use Bugsnag on the DEV environment

aledustet:
- Check for the env[DEV] before loading bugsnag

### DIFF
--- a/ios/AppDelegate.m
+++ b/ios/AppDelegate.m
@@ -17,6 +17,7 @@
 #import <RNSplashScreen.h>
 #import <BT-Swift.h>
 #import <Bugsnag/Bugsnag.h>
+#import "ReactNativeConfig.h"
 
 @implementation AppDelegate
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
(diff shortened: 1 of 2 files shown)


ios/AppDelegate.m
```diff
@@ -48,7 +49,11 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   [[ExposureManager shared] setup];
 
   [RNSplashScreen showSplash:@"LaunchScreen" inRootView:rootView];
-  [Bugsnag start];
+  
+  // Read dev environment variable
+  if ([[ReactNativeConfig envFor:@"DEV"]  isEqual: @"true"]) {
+    [Bugsnag start];
```
juancruzgs, line 55: I think it will cause a runtime error if the library is referenced and we didn't start it.
That's why I used `releaseStages` instead https://github.com/Path-Check/gaen-mobile/pull/258 and it's what I found in their repository https://github.com/bugsnag/bugsnag-js/issues/135

aledustet: You're 💯 right @juancruzgs but we would have to mimic that logic on the javascript side also since we need to start bugsnag in all those three places, this felt easier for this round but we definitely need a way to share that config around

PR #273, opened 2020-08-25
Remove remnants of Tracing Strategy

tparesi:
### Description:

Noticed a couple places referencing tracing strategy - removed them.


tparesi:
> :shipit:
> 
> I believe that the types file is still called `tracingStrategy` maybe we can add it to this one or a new PR , your call

@aledustet I was thinking this too - I will take care of it in this PR.

PR #274, opened 2020-08-25
Function Signature Mismatch Fix

mattThousand:
### Why
To accurately reflect EN permissions state on app launch

### This Commit
This commit fixes a function signature mismatch between `PermissionsContext` and `nativeModule`

### DIFF
--- a/src/gaen/nativeModule.ts
+++ b/src/gaen/nativeModule.ts
@@ -64,11 +64,12 @@ export const requestAuthorization = async (): Promise<string> => {
   return permissionsModule.requestExposureNotificationAuthorization()
 }
 
-export const getCurrentENPermissionsStatus = async (): Promise<
-  ENPermissionStatus
-> => {
+export const getCurrentENPermissionsStatus = async (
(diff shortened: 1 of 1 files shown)


src/gaen/nativeModule.ts
```diff
@@ -64,11 +64,12 @@ export const requestAuthorization = async (): Promise<string> => {
   return permissionsModule.requestExposureNotificationAuthorization()
 }
 
-export const getCurrentENPermissionsStatus = async (): Promise<
-  ENPermissionStatus
-> => {
+export const getCurrentENPermissionsStatus = async (
+  cb: (status: ENPermissionStatus) => void,
```
emilianojuan, line 68: we removed all the callbacks from the native modules, is this ok?

mattThousand: good question - yes, this callback happens in the JS layer

emilianojuan: ok, great! thanks for the explanation :)

PR #276, opened 2020-08-25
Add Location status to home screen for certain Android devices

devinjameson:
Why: some Android devices need location enabled in order for exposure notifications to work. We would like to tell these users whether Location is enabled.

This commit:
- Adds functions to `nativeModule.ts`: `isLocationEnabled` and `doesDeviceSupportLocationlessScanning`
- Adds a `useLocationStatus` hook that exposes `locationStatus` and and refetches it on app state change
- Adds a `useLocationIsNeededStatus` hook that exposes `isLocationNeeded`
- Adds a location status section to the home screen that is shown if the device does NOT support locationless scanning
- Adds an alert if the user taps to fix the location error
- Adds an info page for location

Screen Shot 2020-08-25 at 12 47 18 PM
Screen Shot 2020-08-25 at 12 47 31 PM
Screen Shot 2020-08-25 at 12 47 44 PM
Screen Shot 2020-08-25 at 12 47 56 PM


src/Home/Home.tsx
```diff
@@ -115,7 +136,7 @@ const HomeScreen: FunctionComponent = () => {
   }
 
   const handleOnPressFixBluetooth = () => {
-    showBluetoothDisabledAlert()
+    showFixBluetoothAlert()
```
johnschoeman, line 139: 🥇

src/Home/Home.tsx
```diff
@@ -24,6 +24,10 @@ import { GlobalText } from "../components/GlobalText"
 import { Button } from "../components/Button"
 import { isPlatformiOS } from "../utils/index"
 import { getLocalNames } from "../locales/languages"
+import { useConfigurationContext } from "../ConfigurationContext"
+import { useBluetoothStatus } from "./useBluetoothStatus"
```
johnschoeman, line 28: These three useThingStatus custom hooks are jumping out to me like they may want to be pulled into a common context to manage the ActivationState of the app.  I wouldnt make any changes to this pr, but wondering about your thoughts?

src/Home/Home.spec.tsx
```diff
@@ -355,16 +386,76 @@ describe("Home", () => {
...
+            "To activate Proximity Tracing, authorize COVID-19 Exposure Logging in the Settings app",
+            [
+              expect.objectContaining({ text: "Back" }),
+              expect.objectContaining({ text: "Open Settings" }),
+            ],
+          )
+        })
+      })
+    })
+    describe("when the device requires location permissions", () => {
```
johnschoeman, line 400:
```suggestion
    describe("when the device platform is android", () => {
      it("shows the location activation status", () => {
```

src/Home/Home.spec.tsx
```diff
@@ -355,16 +386,76 @@ describe("Home", () => {
...
+          "home-location-status-container",
+        )
+
+        const locationText = within(locationStatusContainer).getByText(
+          "Location",
+        )
+
+        expect(locationText).toBeDefined()
+      })
+      describe("when the location permissions are on", () => {
```
johnschoeman, line 419:
```suggestion
      describe("and the location permissions are on", () => {
        it("shows the location permission as enabled.", () => {
```

src/Home/Home.spec.tsx
```diff
@@ -355,16 +386,76 @@ describe("Home", () => {
...
+            "home-location-status-container",
+          )
+          const enabledText = within(locationStatusContainer).getByText(
+            "Enabled",
+          )
+
+          expect(enabledText).toBeDefined()
+        })
+      })
+      describe("when the location permissions are off", () => {
```
johnschoeman, line 440:
```suggestion
      describe("and the location permissions are off", () => {
        it("shows the locations permissions as disabled", () => {
```

And a thought: if we have the second two tests here, the permissions on and off cases, do we need the first test asserting that the location is present? or do the second two cases sufficiently cover this?

src/Home/Home.spec.tsx
```diff
@@ -95,27 +97,73 @@ describe("Home", () => {
...
+      expect(header).toHaveTextContent("Active")
+      expect(subheader).toHaveTextContent(
+        `${mockedApplicationName} will remain active after the app has been closed`,
+      )
+      expect(bluetoothEnabledText).toBeDefined()
+      expect(proximityTracingEnabledText).toBeDefined()
+    })
+  })
+
+  describe("When the exposure notification permissions are enabled and the app is authorized, Bluetooth is on, and Location is off, but the platform is iOS", () => {
```
johnschoeman, line 117: this is getting a bit long of a title, what do you think about something more like
```suggestion
  describe("When the device platform is iOS, () => {
    describe("and the exposure notification permissions are enabled and the app is authorized, Bluetooth is on, and Location is off", () => {
```

And wouldnt the location not affect the behavior for ios?

src/Home/Home.spec.tsx
devinjameson: Is this true? I thought this was just for certain versions of Android? Or am I mistaken?

src/Home/Home.tsx
devinjameson: I like that. Seems like a natural grouping. Would be fun to pair on.

src/Home/useLocationIsNeededStatus.ts
```diff
@@ -0,0 +1,27 @@
...
+  const [isLocationNeeded, setIsLocationNeeded] = useState(false)
+
+  useEffect(() => {
+    const determineLocationIsNeeded = async () => {
+      const supportsLocationlessScanning = await determineSupportsLocationlessScanning()
+      setIsLocationNeeded(!supportsLocationlessScanning)
+    }
+
+    determineLocationIsNeeded()
+  }, [])
```
devinjameson, line 16: Can someone explain why all this stuff needs to happen in the `useEffect` with the `[]` dependency array?

src/Home/useLocationIsNeededStatus.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  return isLocationNeeded
+}
+
+const determineSupportsLocationlessScanning = async (): Promise<boolean> => {
+  if (Platform.OS === "android") {
+    return doesDeviceSupportLocationlessScanning()
+  } else {
+    return true
```
devinjameson, line 25: I'm confused as to why we can (have to?) return `boolean` here when the function says it returns `Promise<boolean>`.

src/gaen/nativeModule.ts
```diff
@@ -142,6 +142,16 @@ export const isBluetoothEnabled = async (): Promise<boolean> => {
   return bluetoothStatus === true || bluetoothStatus === "true"
 }
 
+export const doesDeviceSupportLocationlessScanning = async (): Promise<
+  boolean
+> => {
+  return deviceInfoModule.deviceSupportsLocationlessScanning()
```
devinjameson, line 148: I'm probably just used to Elm, but why don't we have to account for the fact that the promise could reject here? E.g.

```
try { return deviceInfoModule.deviceSupportsLocationlessScanning() } catch { ... }
```

src/Home/Home.spec.tsx
devinjameson: I actually don't know. I'd be curious what your recommendation is here because I'm very much learning about how to properly test.

src/Home/Home.spec.tsx
devinjameson: Yes, I was thinking that this would test that. Does that make sense?

src/Home/Home.spec.tsx
johnschoeman: Good question! I am not sure, please only follow through on suggestions if they make sense. I do know that it will never be required on iOS. If its only a subset of android devices require it, are we showing location status on all android devices? should we?

src/Home/Home.spec.tsx
johnschoeman: mostly brought it up to get you thinking about it! as this is a topic that comes up a lot in testing. that is, what is the appropriate amount of coverage for a given feature? and how do we best achieve that coverage?

 On the one hand we want maximum coverage as to feel confident that our system works, but on the other hand if we over cover, then our test suite becomes slow and is a drag on productivity. So its a tradeoff we need to make and have some opinions on what are best practices around what to test and why.

typically we want to cover as much of what is possible as possible, but no more than the minimum. and its common for one test to have overlap with others, which is what is happening here as far as i can tell. In this case, imo, it seems like the second two tests fully cover the first, so the first is redundant, and could be removed. Of course I defer to you as the author of this commit to what makes the most sense though.

let me know if youd like to chat about it more irl!

src/Home/useLocationIsNeededStatus.ts
tparesi: Since the setIsLocationNeeded is set on a component mounting, we have to wait for it to be fetched before setting. The empty dependency array is akin to componentDidMount - meaning it won't be run again after it is mounted.

Not sure if that is what you were asking - but hope it helps.

devinjameson: Ah cool I see. So it prevents this from re-running on every state change?

tparesi: From the useEffect docs:

`If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the dependencies array always works.

If you pass an empty array ([]), the props and state inside the effect will always have their initial values. While passing [] as the second argument is closer to the familiar componentDidMount and componentWillUnmount mental model, there are usually better solutions to avoid re-running effects too often. Also, don’t forget that React defers running useEffect until after the browser has painted, so doing extra work is less of a problem.`

src/Home/Home.spec.tsx
devinjameson: This is super helpful, thank you! I agree. Will remove that test.

src/Home/Home.spec.tsx
devinjameson: We are only showing location status on Android devices if the device does not support locationless scanning.

devinjameson: I'm thinking this:

```
describe("when the device does not support locationless scanning", () => {
```

src/Home/Home.spec.tsx
devinjameson: Ended up removing this test because it was testing the same thing as the final tests.

src/Home/useLocationIsNeededStatus.ts
devinjameson: Thanks!

src/Home/Home.tsx
```diff
@@ -24,6 +24,10 @@ import { GlobalText } from "../components/GlobalText"
...
+import { useLocationStatus } from "./useLocationStatus"
+import { useLocationIsNeededStatus } from "./useLocationIsNeededStatus"
```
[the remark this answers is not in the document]
devinjameson: Mm I think you're onto something here. I'll take a wack at this and hopefully we can pair shortly!

src/gaen/nativeModule.ts
johnschoeman: This is a pretty deep question, so perhaps worth chatting about irl, and theres a few reasons worth pointing out here so I'll just mention that `deviceInfoModule.deviceSupportsLocationlessScanning()` is of type `any` so the compiler has no way of understanding what this returns.  you can play around with updating the type signature to anything else and notice that the complier wont complain at all.

You can kinda think of these functions in this file as wrappers around the IO edge of our system which doesnt have any type information inherently. So here we are doing the work of provide the basic type the compiler will check against for the rest of our system. Which is why type bugs are likely to be common in this file and are worth taking and extra look at and providing integration tests for.

devinjameson: Ah I see. Very interesting. Thanks for the thorough explanation!

PR #278, opened 2020-08-25
Use single share app link

johnschoeman:
Why:
Currently we are making a distinction between ios and android share
links, which doesnt meet the feature requirements as we do not know the
device of the user in which the app is being shared.

This commit:
Removes the casing on the device platform for the share app link and
replaces the two links in the env file with a single 'SHARE_APP_LINK'

Co-Authored-By: Arthur Gibson <arthur.gibson@pathcheck.org>

### without link
Screen Shot 2020-08-25 at 11 15 16 AM

### with link
Screen Shot 2020-08-25 at 11 42 26 AM

### DIFF
--- a/src/Home/Home.tsx
+++ b/src/Home/Home.tsx
@@ -130,6 +130,8 @@ const HomeScreen: FunctionComponent = () => {
     navigation.navigate(Screens.LanguageSelection)
   }
 
+  const showShareLink = Boolean(configuration.appDownloadLink)
+
   const isProximityTracingOn = isEnabledAndAuthorized
   const isBluetoothOn = btStatus
   const appIsActive = isProximityTracingOn && isBluetoothOn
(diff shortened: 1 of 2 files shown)


src/Home/Home.tsx
```diff
@@ -130,6 +130,8 @@ const HomeScreen: FunctionComponent = () => {
     navigation.navigate(Screens.LanguageSelection)
   }
 
+  const showShareLink = configuration.appDownloadLink !== null
```
aledustet, line 133: This is making a transformation on a value we set up, how do you feel about setting this setting higher? Or perhaps using the null value as a boolean with the casting `Boolean(appDownloadLink)` down below? It's probably fine but I'm worried that we keep pushing defensive code down the pipe when we could avoid it by setting nondisruptive values on the context(this null seems nondisruptive to me), maybe it's just because this screen is getting a little large and we could move the share link to a component that renders null if no download link is present? ... Sorry for the rant, I hope it makes sense 😬

johnschoeman: Makes a lot of sense! I agree that the explicit null checking is not the best pattern, so will make an update. thanks!

PR #279, opened 2020-08-25
Add spec for NotificationPermissions

devinjameson:
Why: we would like to have better test coverage.

### DIFF
--- a/src/Activation/NotificationPermissions.spec.tsx
+++ b/src/Activation/NotificationPermissions.spec.tsx
@@ -1,0 +1,63 @@
+import React from "react"
+import { render, fireEvent, waitFor } from "@testing-library/react-native"
+import NotificationPermissions from "./NotificationPermissions"
+
+import {
+  PermissionsContext,
+  PermissionsContextState,
+} from "../PermissionsContext"
+import { OnboardingContext, OnboardingContextState } from "../OnboardingContext"
+import { PermissionStatus } from "../permissionStatus"
+
+describe("NotificationPermissions", () => {
+  describe("when a user enables notifications,", () => {
+    it("requests permissions", () => {
+      const onboardingProviderValue = createOnboardingProviderValue()
+
+      const notificationRequestSpy = jest.fn()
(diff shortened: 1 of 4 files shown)


src/Activation/NotificationPermissions.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+            <NotificationPermissions />
+          </PermissionsContext.Provider>
+        </OnboardingContext.Provider>,
+      )
+
+      fireEvent.press(getByLabelText("Enable Notifications"))
+
+      expect(notificationRequestSpy).toHaveBeenCalled()
+    })
+    it("marks onboarding complete", async () => {
```
aledustet, line 34: Are all of these actions asynchronous? I believe that if we are testing an action from the users we could potentially leave all this tests without the `async`?(probably wrong, but maybe worth the shot)

src/Activation/NotificationPermissions.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+      fireEvent.press(getByText("Maybe later"))
+
+      await waitFor(() => {
+        expect(completeOnboardingSpy).toHaveBeenCalled()
+      })
+    })
+  })
+})
+
+const createPermissionProviderValue = (
```
aledustet, line 109: Maybe is just the pattern but don't we have these helpers in a few places, could we benefit from having a factory for these? Please ignore if there are just a couple of places where it makes sense to test the permissions context

devinjameson: Yeah that sounds good! I'll pick that up in a future PR.

src/Activation/NotificationPermissions.spec.tsx
devinjameson: Ah, so we can remove the `async` from the last test, but it needs to stay here because when a user enables notifications that request is async, so we have to wait for the `completeOnboardingSpy` to have been called _after_ that.

PR #280, opened 2020-08-25
Fix typo in default ios en config

johnschoeman:


mattThousand:
> This is now correct, but please make sure that we fetch from the online config rather than use the static in the app.

this is the "online config" that will live on and endpoint on the scoring, server, is that right?

PR #281, opened 2020-08-25
Add a validation to AffectedUser CodeInput form

tparesi:
#### Description:

Validate that the AffectedUser CodeInput is a number.

#### Linked issues:

[Trello](https://trello.com/c/mPffXwd3/379-as-a-user-when-i-enter-a-verification-code-if-that-code-is-not-formated-correctly-only-digits-i-am-told-that-theres-a-formating)

### DIFF
--- a/src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
+++ b/src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
@@ -49,10 +49,16 @@ const CodeInputForm: FunctionComponent = () => {
   const [isFocused, setIsFocused] = useState(false)
 
   const codeLength = 8
+  const codeIsInvalidLength = code.length !== codeLength
+  const codeContainsNonDigitChars = (code: string) => !code.match(/^\d+$/)
 
-  const handleOnChangeText = (code: string) => {
+  const handleOnChangeText = (newCode: string) => {
     setErrorMessage("")
-    setCode(code)
+    if (newCode && codeContainsNonDigitChars(newCode)) {
+      setErrorMessage(t("export.error.invalid_format"))
+    } else {
+      setCode(newCode)
+    }
   }
 
   const handleOnToggleFocus = () => {
@@ -144,7 +150,7 @@ const CodeInputForm: FunctionComponent = () => {
     }
   }
 
-  const isDisabled = code.length !== codeLength
+  const isDisabled = codeIsInvalidLength || codeContainsNonDigitChars(code)
 
   const codeInputFocusedStyle = isFocused && { ...style.codeInputFocused }
   const codeInputStyle = { ...style.codeInput, ...codeInputFocusedStyle }
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -82,6 +82,7 @@
     "enable_exposure_notifications_body": "You must enable exposure notifcations to report a positive test result.",
     "enable_exposure_notifications_title": "Enable exposure notifications to continue",
     "error": {
+      "invalid_format": "Invalid format",
       "invalid_code": "Try a different code",
       "unknown_code_verification_error": "Try a different code",
       "verification_code_used": "Verification code has already been used"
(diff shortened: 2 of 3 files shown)


src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -52,7 +52,11 @@ const CodeInputForm: FunctionComponent = () => {
 
   const handleOnChangeText = (code: string) => {
     setErrorMessage("")
-    setCode(code)
+    if (code && !code.match(/^\d+$/)) {
```
devinjameson, line 55: Thoughts on naming `!code.match(/^\d+$/)`?

```
const codeContainsNonDigitChars = !code.match(/^\d+$/)
if (code && codeContainsNonDigitChars) {
...
```

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -144,7 +148,7 @@ const CodeInputForm: FunctionComponent = () => {
     }
   }
 
-  const isDisabled = code.length !== codeLength
+  const isDisabled = code.length !== codeLength || !code.match(/^\d+$/)
```
devinjameson, line 151: You could also use `codeContainsNonDigitChars` here.

devinjameson: It may also be nice to name `code.length !== codeLength`

```
const codeIsInvalidLength = code.length !== codeLength
const isDisabled = codeIsInvalidLength || codeContainsNonDigitChars
```

src/locales/en.json
```diff
@@ -82,6 +82,7 @@
     "enable_exposure_notifications_body": "You must enable exposure notifcations to report a positive test result.",
     "enable_exposure_notifications_title": "Enable exposure notifications to continue",
     "error": {
+      "invalid_format": "Verification code can only contain numbers",
```
johnschoeman, line 85:
```suggestion
      "invalid_format": "Invalid format",
```
Mostly just worried about translations being more that one line in length. so fewer words > more words as a rule of thumb.

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
johnschoeman: might need to be something more like:
```typescript
const codeIsValid = code && !code.match(/^\d+$/)
```
as `!code.match(/^\d+$/)` would error if code is null

tparesi: @johnschoeman This check is there already - in the if statement.

@devinjameson Since the naming in this file uses code as local state as well as an argument for this function, I renamed code in this instance. Also, the codeContainsNonDigitChars has to be a function because of this. Let me know what you think of the changes.

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -49,10 +49,16 @@ const CodeInputForm: FunctionComponent = () => {
   const [isFocused, setIsFocused] = useState(false)
 
   const codeLength = 8
+  const codeIsInvalidLength = code.length !== codeLength
+  const codeContainsNonDigitChars = (code: string) => !code.match(/^\d+$/)
```
devinjameson, line 53: Nice! I tend to prefer explicit returns but that's totally a matter of preference.
```
const codeContainsNonDigitChars = (code: string) => { return !code.match(/^\d+$/) }
```

devinjameson: Just making sure re: @johnschoeman's comment, does `code.match(…)` not error if `code` is null?

tparesi: For the first use of codeContainsNonDigitChars in handleOnChangeText - we have an empty string check so that we don't show an error when the code is an empty string. This happens when deleting all the characters after typing.

For the second use as part of isDisabled, we do not need an empty string check because the valid length check handles that for us.

We should be good in both cases - since the code is never null, but an empty string.

PR #282, opened 2020-08-25
Add Script to get icon assets resources repo

aledustet:
Why:
----
We need to update the icons for each one of the builds corresponding to a different jurisdiction.

Caveat:
----
This process has only been tested locally and is still to test on the CI environment. Also there is currently only on jurisdiction with the assets available.

This Commit:
----
- Add a script to download assets from resources repo and copy to locations inside ios/android projects

### DIFF
--- a/bin/download_assets.sh
+++ b/bin/download_assets.sh
@@ -1,0 +1,201 @@
+#!/usr/bin/env ruby
+#
+# Download specific assets for the health authority
+#
+# Usage
+#
+#   bin/download_assets.sh <health-authority-label> <github-token>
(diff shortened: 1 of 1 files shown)


aledustet:
> LGTM!
> 
> The only thing that jumps out to me is the use of early returns which I always question if i can remove to reduce branching complexity. But would consider any refactors in a follow up commit.

Do you mean that we should have more or less of those early returns? Should we consider an actual 1 exit status, with short circuits for errors or you mean we should nest decisions on `if-else` blocks?

PR #283, opened 2020-08-26
Refresh exposure history screen

zhaccc:
**Why:**
Exposure history does not re-render unless we restart the app.

**What:**
Since the screens are mounted when the app starts and they newer unmount we relied on listeners to figure out when the app is in focused/blur state and caused re-render with the state change.
Also, we added pull to refresh functionality. To achieve that we compare previous exposure state with the current one to determine if we need to re-render the screen.


src/ExposureHistory/History/index.tsx
```diff
@@ -34,11 +36,26 @@ const History: FunctionComponent<HistoryProps> = ({
...
+
+  useEffect(() => {
+    prevExposuresRef.current = exposures
+  })
 
   const handleOnPressMoreInfo = () => {
     navigation.navigate(Screens.MoreInfo)
   }
 
+  const handleRefresh = () => {
```
devinjameson, line 50: Thoughts on this for consistency?

```suggestion
  const handleOnRefresh = () => {
```

src/ExposureHistory/History/index.tsx
```diff
@@ -51,6 +68,9 @@ const History: FunctionComponent<HistoryProps> = ({
           contentContainerStyle={style.contentContainer}
           style={style.container}
           alwaysBounceVertical={false}
+          refreshControl={
+            <RefreshControl refreshing={refreshing} onRefresh={handleRefresh} />
```
devinjameson, line 72:
```suggestion
            <RefreshControl refreshing={refreshing} onRefresh={handleOnRefresh} />
```

devinjameson:
It may be worth checking with design if they'd like an icon at the top of this screen indicating that you can pull to refresh.

src/ExposureHistory/ExposureDetail.tsx
```diff
@@ -27,7 +27,7 @@ const ExposureDetail: FunctionComponent = () => {
     healthAuthorityAdviceUrl,
   } = useConfigurationContext()
 
-  const [connectivity, setConnectivity] = useState<boolean | null | undefined>(
+  const [isConnected, setIsConnected] = useState<boolean | null | undefined>(
```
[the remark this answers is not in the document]
zhaccc: Definitely but let's do a separate task for this. I'll cut a card and handle it.

zhaccc:
> It may be worth checking with design if they'd like an icon at the top of this screen indicating that you can pull to refresh.

Good idea, I'll check with the team and we can cut a card for it!

PR #284, opened 2020-08-26
Convert splash screen and download icon asset

aledustet:
Why:
----
The splash screen for the app should be a blank screen with the logo of the app centered.
The status bar on android has a solid color, that is contrasting with the color from the top section on the views.

How:
----
The resource for ios that does this is the LaunchScreen, so it needs to be modified to a white canvas with a centered image. In this case the loglogoImage asset. For android is the launch_screen resource, that needs to be transformed in the same way.

The new resources will be available from the resources repo, and placed in the right spot by the download assets script.

This Commit:
----
- Refactor helpers to download files
- Convert ios splash screen to be a centered logo
- Update android splash screen to be a white background with the centered logo
- Copy ios logoImage to project from resources repo
- Use fixed colors for main layout on android
- Add splash background color
- Make status bar transparent for android
- Make the color for status bar transparent
- Delete unused bgImage

Screen Shot 2020-08-26 at 2 26 08 PMScreen Shot 2020-08-26 at 12 00 33 PMScreen Shot 2020-08-26 at 2 26 18 PM
Screen Shot 2020-08-26 at 2 46 28 PMScreen Shot 2020-08-26 at 3 02 52 PM


src/navigation/index.ts
```diff
@@ -183,18 +181,13 @@ export const Stacks: { [key in Stack]: Stack } = {
...
-      ? Colors.primaryLightBackground
-      : Colors.primary125
-
   useFocusEffect(
     useCallback(() => {
       StatusBar.setBarStyle(statusBarStyle)
       if (Platform.OS === "android") {
-        StatusBar.setBackgroundColor(backgroundColor)
+        StatusBar.setBackgroundColor("transparent")
         StatusBar.setTranslucent(true)
```
devinjameson, line 189: Is this redundant now? I remember @johnschoeman and I testing this a few weeks ago and it didn't seem to do anything.

src/navigation/index.ts
```diff
@@ -183,18 +181,13 @@ export const Stacks: { [key in Stack]: Stack } = {
...
-    statusBarStyle === "dark-content"
-      ? Colors.primaryLightBackground
-      : Colors.primary125
-
   useFocusEffect(
     useCallback(() => {
       StatusBar.setBarStyle(statusBarStyle)
       if (Platform.OS === "android") {
-        StatusBar.setBackgroundColor(backgroundColor)
+        StatusBar.setBackgroundColor("transparent")
```
devinjameson, line 188: 😍

android/app/src/main/res/values/splash.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <color type="color" name="splash_background">#ffffff</color>
```
juancruzgs, line 3: We could move this one to `colors.xml`, just to have everything in the same place

android/app/src/main/res/values/styles.xml
```diff
@@ -1,20 +1,18 @@
 <resources>
-
     <!-- Base application theme. -->
     <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
         <!-- Customize your theme here. -->
         <item name="android:textColor">#000000</item>
         <item name="android:windowLightStatusBar">false</item>
-        <item name="android:statusBarColor">@color/launcher_background</item>
-        <item name="android:windowBackground">@color/launcher_background</item>
+        <item name="android:statusBarColor">@color/splash_background</item>
```
juancruzgs, line 7: This will set a white status bar, is that what we want?

juancruzgs: We might need 2 colors, `splash_background` & `status_bar`

bin/helpers.rb
```diff
@@ -0,0 +1,64 @@
+def download_file(file_name, remote_url)
```
johnschoeman, line 1: what do you think about breaking this file up into something like `file_system_helpers.rb` and `remote_repo_helpers.rb` ?

android/app/src/main/res/values/splash.xml
aledustet: The thing is that we override colors per jurisdiction 😬

android/app/src/main/res/values/styles.xml
aledustet: We left it as one there because the status bar color only shows up on that little spot between splash screen and app, but you're totally right we should separate those values. We hardcoded it whit because it's only for that portion since it is overridden by the application as soon it loads.

src/navigation/index.ts
aledustet: The good news is that it actually is doing something 🎉

android/app/src/main/res/values/styles.xml
aledustet: @juancruzgs updated are we good to go?

bin/helpers.rb
aledustet: I don't feel strongly against it but it feels like overkill to re-organize helpers for helpers, I was just thinking of a bucket to not repeat copy helpers that we are using in one place now but the trend is upwards. Can modify it if we think is very necessary

android/app/src/main/res/values/splash.xml
juancruzgs: That makes sense then 💯

android/app/src/main/res/values/styles.xml
```diff
@@ -1,20 +1,18 @@
...
+        <item name="android:windowBackground">@color/splash_background</item>
     </style>
 
     <style name="SplashTheme" parent="Theme.AppCompat.Light.NoActionBar">
         <item name="android:windowLightStatusBar">true</item>
-        <item name="android:statusBarColor">@color/launcher_background</item>
-        <item name="android:windowBackground">@color/launcher_background</item>
-        <item name="android:background">@drawable/background_splash</item>
-        <item name="colorPrimaryDark">@color/launcher_background</item>
+        <item name="android:statusBarColor">@color/splash_background</item>
```
juancruzgs, line 13: This one should be `splash_status_bar` , right?

android/app/src/main/res/values/styles.xml
juancruzgs: Does ReactNative override this value and sets something different for the rest of the app?
I'm asking because this is defined to be used in the entire application from the native side

bin/helpers.rb
johnschoeman: Certainly not necessary.

android/app/src/main/res/values/styles.xml
aledustet: Yes, there are overrides per screen, each one will be responsible for setting the right color for the status bar.

juancruzgs: Okay I can remove these values later to avoid confusion
Looks good to me!

PR #285, opened 2020-08-26
(iOS) Implement Scoring Server API Contract

mattThousand:
### Why
We'd like to implement the scoring server's api contract to accommodate `v1` of the ExposureNotifications framework apple API

### This Commit
This commit implements the scoring server's api


ios/BT/API/Requests/ExposureDetectionSummaryRequests.swift
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+private extension ExposureDetectionSummary {
+
+  var asJsonBody: JSONObject {
+    return [
+        "date_received": startOfDateReceived,
+        "timezone_offset": timezoneOffset,
+        "attenuation_durations_seconds": [],
```
lmartensson, line 44: This needs to be the low/medium/high buckets instead right?

ios/BT/ExposureManager.swift
```diff
@@ -489,12 +487,12 @@ private extension ExposureManager {
...
-      case .success(let exposureConfiguration):
-        self?.exposureConfiguration = exposureConfiguration
-      case .failure(let error):
-        print("Error fetching exposure configuration: \(error)")
-      }
+                        switch result {
+                        case .success(let exposureConfiguration):
+                          self?.exposureConfiguration = exposureConfiguration
+                        case .failure(let error):
+                          print("Error fetching exposure configuration: \(error)")
```
lmartensson, line 494: Outside the scope of this pull request, but, what happens if there was an error fetching the config? Will it just keep using the previous one?

ios/BT/ExposureManager.swift
```diff
@@ -525,4 +523,61 @@ private extension ExposureManager {
...
+      // Call scoring server
+      apiClient.request(ExposureDetectionSummaryListRequest.post(exposureDetectionSummary,
+                                                                 unusedExposureSummaries),
+                        requestType: .scoring) { result in
+                          switch result {
+                          case .success(let scoringServerResponse):
+                            let summaries = scoringServerResponse.notifications.map { $0.exposureSummaries }.flatMap { $0 }
+                            completion(summaries)
+                          case .failure:
+                            completion([])
```
lmartensson, line 551: I think this is outside of the scope of this change, but, how hard would it be to try the same thing again later? (Perhaps the next time we fetch a new key batch.)
(Consider writing a note here that we should handle that case.)

ios/BT/ExposureManager.swift
```diff
@@ -525,4 +523,61 @@ private extension ExposureManager {
     localUncompressedURLs = []
     downloadedPackages = []
   }
+
+  func processExposureDetectionSummary(_ summary: ENExposureDetectionSummary?, completion: @escaping (([ExposureDetectionSummary]) -> Void)) {
```
lmartensson, line 527: Yup, this looks right!

ios/BT/ExposureManager.swift
```diff
@@ -525,4 +523,61 @@ private extension ExposureManager {
...
+                            let summaries = scoringServerResponse.notifications.map { $0.exposureSummaries }.flatMap { $0 }
+                            completion(summaries)
+                          case .failure:
+                            completion([])
+                          }
+      }
+    }
+  }
+
+  func processScoringServerResponse(_ summaries: [ExposureDetectionSummary],
```
lmartensson, line 557: Hmm, I'm a little confused here, but I also don't know Swift.

The first argument is the response from the server? Which should be a Notification, containing an array of summaries, but also other things.

ios/BT/ExposureManager.swift
```diff
@@ -525,4 +523,61 @@ private extension ExposureManager {
...
+    }
+
+    // Remove "used" exposure summaries from realm
+    BTSecureStorage.shared.pruneExposureDetectionSummaries(summaries)
+
+    // Store new exposures in realm
+    btSecureStorage.storeExposures(newExposures)
+
+    // Post notification to user
+    postNewExposureNotification()
```
lmartensson, line 575: I'm not entirely sure what happens at this point. Does this notify the user without any additional context?
I'm not really caught up on what the UI looks at this point, and I'm fine with whatever there - just want to make sure I understand what the code is meant to do.

ios/BT/ExposureManager.swift
```diff
@@ -525,4 +523,61 @@ private extension ExposureManager {
...
+    let newExposures = (summaries).map { summary in
+      Exposure(id: UUID().uuidString,
+               date: summary.startOfDateReceived)
+    }
+
+    // Remove "used" exposure summaries from realm
+    BTSecureStorage.shared.pruneExposureDetectionSummaries(summaries)
+
+    // Store new exposures in realm
+    btSecureStorage.storeExposures(newExposures)
```
lmartensson, line 572: By "exposures" here, you mean the exposures that led to a notification?
If so, I'd consider different naming here - as a reader, it's not obvious what the difference is here.
And consider storing the whole notification object instead?

ios/BT/API/Requests/ExposureDetectionSummaryRequests.swift
mattThousand: good catch, this is out of date - incorrect parameters and also I need to use camelcase

ios/BT/ExposureManager.swift
mattThousand: Good question - yes, if there is an error fetching the config from the server, the hard-coded placeholder value will be used: https://github.com/Path-Check/gaen-mobile/blob/a71338f67bdfac2e57377c1b4891b58a22e133d0/ios/BT/API/Model/ExposureConfiguration.swift#L19

ios/BT/ExposureManager.swift
mattThousand: ah - this piece is not related to fetching keys from the key server; it is the completion after the mobile client makes a request to the scoring server. If successful, if passes the `notification` objects returned from the server into the completion. If it fails due to a server error, we should log it, but the completion should still be executed

lmartensson: No, that's what I mean - the part about the key server was just a suggestion of when might be a good time to retry. (We don't actually need retries to be instant, since the user isn't actually interacting here!)

But basically, if the server were down for some reason, it would be nice to retry a bit later. Could be done as a follow-up change though.

ios/BT/ExposureManager.swift
lmartensson: Cool! I think this is good for now.

lmartensson: Perhaps add a comment here to say that that's the case?

ios/BT/ExposureManager.swift
```diff
@@ -525,4 +523,61 @@ private extension ExposureManager {
...
   }
+
+  func processExposureDetectionSummary(_ summary: ENExposureDetectionSummary?, completion: @escaping (([ExposureDetectionSummary]) -> Void)) {
+    if let summary = summary {
+
+      // Serialize exposure detection summary
+      let exposureDetectionSummary = ExposureDetectionSummary(summary)
+
+      // Prune exposure detection summaries >14 days from realm
+      BTSecureStorage.shared.pruneExposureDetectionSummaries()
```
emilianojuan, line 534: can we use btSecureStorage here to make this testable?

ios/BT/ExposureManager.swift
```diff
@@ -525,4 +523,61 @@ private extension ExposureManager {
...
+                                    progress: Progress,
+                                    completionHandler: @escaping ((ExposureResult) -> Void)) {
+
+    let newExposures = (summaries).map { summary in
+      Exposure(id: UUID().uuidString,
+               date: summary.startOfDateReceived)
+    }
+
+    // Remove "used" exposure summaries from realm
+    BTSecureStorage.shared.pruneExposureDetectionSummaries(summaries)
```
emilianojuan, line 569: same here

ios/BT/ExposureManager.swift
mattThousand: sorry - yes, this was from a commit before your refactor 👍

ios/BT/API/Model/ExposureDetectionSummary.swift
```diff
@@ -9,19 +9,19 @@ class ExposureDetectionSummary: Object, Codable {
...
+  @objc dynamic var attenuationDuration: AttenuationDuration? = AttenuationDuration()
+  @objc dynamic var maximumRiskScore: Int8 = 0
   @objc dynamic var riskScoreSumFullRange: Double = 0.0
 
   init(_ enExposureDetectionSummary: ENExposureDetectionSummary) {
     timezoneOffset = TimeZone.current.secondsFromGMT()
-    attenuationDuration.low = Int(truncating: enExposureDetectionSummary.attenuationDurations[0])
-    attenuationDuration.medium = Int(truncating: enExposureDetectionSummary.attenuationDurations[1])
-    attenuationDuration.high = Int(truncating: enExposureDetectionSummary.attenuationDurations[2])
+    attenuationDuration?.low = Int(truncating: enExposureDetectionSummary.attenuationDurations[0])
```
lmartensson, line 19: Ugh, so sorry - the swagger API does indeed say attenuationDuration but it should be plural: attenuationDurations. I think Matt might even have asked me about this. But, I guess attenuationDurations plural it is.
Could you please update this?

PR #286, opened 2020-08-26
Run download assets with the set_ha script

aledustet:
Why:
----
We need an automated way on CI to setup the right assets for the different jurisdictions. Currently the set_ha script is ran on each release build on Bitrise.

This Commit:
----
- Add `download_assets` to the set_ha script

Caveat:
----
If the script is executed for one of the jurisdictions that have all the assets ready, it will create a huge number of extra files that will pollute the local environment. We could change the script settings for local setup purposes(adding an env var to the .env file maybe), but for practical purposes we do need the assets to be copied over on the CI environment


johnschoeman:
for the extra files, should we gitignore them? or perhaps make a clean_up script to remove them?

aledustet:
> for the extra files, should we gitignore them? or perhaps make a clean_up script to remove them?

I don't know about the gitignore part since those are assets that are used to build the app and if at some point we want to change them maybe it becomes cumbersome to update? As for a cleanup script we could extract the abort for the `download_assets` into its own script

PR #287, opened 2020-08-26
Chore/add download assets to set ha

aledustet:
Why:
----
The splash screen for the app should be a blank screen with the logo of the app centered.
The status bar on android has a solid color, that is contrasting with the color from the top section on the views.

How:
----
The resource for ios that does this is the LaunchScreen, so it needs to be modified to a white canvas with a centered image. In this case the loglogoImage asset. For android is the launch_screen resource, that needs to be transformed in the same way.

The new resources will be available from the resources repo, and placed in the right spot by the download assets script.

We need an automated way on CI to setup the right assets for the different jurisdictions. Currently the set_ha script is ran on each release build on Bitrise.

Caveat:
----
If the script is executed for one of the jurisdictions that have all the assets ready, it will create a huge number of extra files that will pollute the local environment. We could change the script settings for local setup purposes(adding an env var to the .env file maybe), but for practical purposes we do need the assets to be copied over on the CI environment

This Commit:
----
- Refactor helpers to download files
- Convert ios splash screen to be a centered logo
- Update android splash screen to be a white background with the centered logo
- Copy ios logoImage to project from resources repo
- Use fixed colors for main layout on android
- Add splash background color
- Make status bar transparent for android
- Make the color for status bar transparent
- Delete unused bgImage
- Add `download_assets` to the set_ha script

Reviewers: 
----
Something happened(I pressed the close button) on the first or and then CI was not happy, so we reopened it from #284 


aledustet:
@johnschoeman someone(me) pressed the close and comment button 😅

PR #289, opened 2020-08-27
Refactor connectivity status

zhaccc:
**Why:**
Connection status functionality should be more reusable.

**How:**
Refactored connection status to be a custom hook.


src/hooks/useConnectionStatus.ts
```diff
@@ -0,0 +1,21 @@
+import { useState, useEffect } from "react"
+import NetInfo from "@react-native-community/netinfo"
+
+export const useConnectionStatus = (): boolean | null | undefined => {
+  const [isConnected, setIsConnected] = useState<boolean | null | undefined>(
```
aledustet, line 5: How do you feel about narrowing down the type, since we are setting up this value, we could always set a boolean and avoid the union type right?

src/hooks/useConnectionStatus.ts
```diff
@@ -0,0 +1,19 @@
...
+import NetInfo from "@react-native-community/netinfo"
+
+export const useConnectionStatus = (): boolean => {
+  const [isConnected, setIsConnected] = useState<boolean>(true)
+
+  useEffect(() => {
+    const unsubscribe = NetInfo.addEventListener((state) => {
+      // netInfo state comes as null while unresolved so to avoid flicker we only set component state
+      // if the netInfo state is resolved to boolean
+      if (typeof state.isInternetReachable === "boolean") {
```
aledustet, line 11: I feel like this type check could be avoided if we do something like:

```typescript
useEffect(() => {
  const unsubscribe = NetInfo.addEventListener((netInfo) => {
    setIsConnected(Boolean(netInfo?.isInternetReachable))
  })
  return unsubscribe
}, [])
```
Could that work? I feel like it is less hard to read the `Boolean` conversion than the type check.

src/hooks/useConnectionStatus.ts
```diff
@@ -0,0 +1,19 @@
+import { useState, useEffect } from "react"
+import NetInfo from "@react-native-community/netinfo"
+
+export const useConnectionStatus = (): boolean => {
+  const [isConnected, setIsConnected] = useState<boolean>(true)
```
aledustet, line 5: How do you feel about keeping the naming from the `NetInfo` library since we are doing a wrapper? Isn't `isInternetReachable` a little more descriptive and specific than `isConnected`?

src/hooks/useConnectionStatus.ts
zhaccc: Unfortunately, this won't work. `undefined` we get with optional chaining will convert to `false` and we will get flicker we were getting before we used this check here.

aledustet (approved):

PR #290, opened 2020-08-27
Clean up Home component

amandabeiner:
### Description:
Why:
As a dev team, we want the home screen to be clear an manageable so that we can make changes quickly

This commit:
Cleans up some pain points we noticed while working with the Home screen. This includes:
- Moving shared hooks into a `hooks` directory so that they are not imported from sibling modules
- Exporting common components consistently so that they can be imported consistently
- Breaking out the Home component into smaller child components to that it is more readable and editable
- Updating the stylesheet in the Home component to take arguments to move css logic out of the component and minimize the number of places where we are defining styles
- Updating the Permissions context to track `authorizedAndEnabled` status so that we don't redefine this logic in multiple places

Co-Authored-By: devinjameson <devin@thoughtbot.com>


devinjameson:
>  to move css logic out of the component

This is technically just "styling" logic. Unfortunately, we have no cascade in RN :'( That's kind of a pedantic note but thought it might be worth sharing!

src/More/About.spec.tsx
```diff
@@ -2,11 +2,11 @@ import React from "react"
 import { render } from "@testing-library/react-native"
 
 import AboutScreen from "./About"
-import { useApplicationInfo } from "./useApplicationInfo"
+import { useApplicationInfo } from "../hooks/useApplicationInfo"
```
aledustet, line 5: 🙏

src/Activation/ActivateProximityTracing.tsx
```diff
@@ -13,7 +13,7 @@ import { useNavigation } from "@react-navigation/native"
 
 import { usePermissionsContext } from "../PermissionsContext"
 import { useOnboardingContext } from "../OnboardingContext"
-import { useApplicationName } from "../More/useApplicationInfo"
+import { useApplicationName } from "../hooks/useApplicationInfo"
```
johnschoeman, line 16: So far we have been avoiding `directory by implementation` and preferring `directory by feature`. What do you think about keeping this convention the same? Is there a benefit to grouping all the hooks together, especially if a hook is only used in one place and is not intended to be used throughout the app?

src/Activation/NotificationPermissions.spec.tsx
```diff
@@ -109,7 +109,7 @@ const createPermissionProviderValue = (
       request: requestPermission,
     },
     exposureNotifications: {
-      status: { authorized: true, enabled: true },
+      status: { authorized: true, enabled: true, authorizedAndEnabled: true },
```
johnschoeman, line 112: Does this introduce a risk of getting into impossible states? e.g. `{ authorized: false, enabled: false, authorizedAndEnabled: true }`.  If we need to dry up logic, what do you think about introducting a helper function or maybe updating the exposed status type altogether? would a union type be a better fit than an object?

src/Home/ActivationStatus.tsx
```diff
@@ -0,0 +1,106 @@
+import React, { FunctionComponent } from "react"
+import { useTranslation } from "react-i18next"
+import { TouchableOpacity, View, StyleSheet } from "react-native"
+import { SvgXml } from "react-native-svg"
+
+import { GlobalText } from "../components"
+
+import { Icons } from "../assets"
+import { Colors, Iconography, Outlines, Spacing, Typography } from "../styles"
+interface ActivationStatusProps {
```
johnschoeman, line 10: missing space above?

src/Home/ShareLink.tsx
```diff
@@ -0,0 +1,98 @@
...
+        }),
+      })
+    } catch (error) {
+      throw new Error(error)
+    }
+  }
+
+  return (
+    <>
+      {showShareLink ? (
```
johnschoeman, line 35: could this conditional be moved up a level?

src/Home/LocationActivationStatus.tsx
```diff
@@ -0,0 +1,56 @@
...
+          text: t("common.settings"),
+          onPress: () => Linking.openSettings(),
+        },
+      ],
+    )
+  }
+
+  return (
+    <>
+      {showLocationStatus && (
```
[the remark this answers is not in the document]
amandabeiner: I'm not opposed to moving this logic back up per se, but I think this is part of why the Home component became so unreadable. It was using a _lot_ of state to determine which children to render, which  meant extra variable declarations and many ternaries and conditionals in the `return`. This became even harder to follow because the variable definitions were really far from where those variables were used.

In this case, I might prefer to have the ActivationStatus and the ShareLink be responsible for reading from context and displaying themselves correctly. I think it's easier to read because these components are much simpler, so there is only one conditional in the return and it is directly related to the component's display logic. What do you think?

cc @johnschoeman because we made a similar decision in the ShareLink

src/Home/ShareLink.tsx
```diff
@@ -0,0 +1,98 @@
...
+              height={Iconography.small}
+            />
+          </View>
+        </TouchableOpacity>
+      ) : null}
+    </>
+  )
+}
+
+// TODO revisit reusability here
```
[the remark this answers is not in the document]
amandabeiner: WHoops missed this! thanks!!

src/Activation/ActivateProximityTracing.tsx
amandabeiner: I agree that there's no benefit to grouping all the hooks together, and this PR does not move all of the hooks into this directory. This specific hook is currently very widely used throughout the app (Home, Activation, Onboarding, More, ReportIssue), so it is functioning as a shared resource.

I try to avoid the pattern currently in the app (importing a utility or component from a sibling directory), because it's easy to mistakenly think that the hook defined within that feature is only used in that feature, and to accidentally break its uses elsewhere. In this case, this shared hook lives in the `More` feature, but is imported into other features very often.

I was thinking about this `hooks` directory kind of like our `components` directory-- Not all components go into it, but ones that are widely reused throughout the app do. Is there a more clear place for it?

src/Activation/NotificationPermissions.spec.tsx
amandabeiner: Oooh I like this idea. Going to noodle on it for a sec 🤔

src/Home/LocationActivationStatus.tsx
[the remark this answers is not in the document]
aledustet: I'm sorry I meant a little higher in this exact component, not back up to the parent one, 💯 agree with you on this reasoning, moving it here and making this component responsible for its own logic is the perfect move. Just thought that since now is more compact and nice to understand doing:

``` typescript
if(!showLocationStatus) {
  return null
}
```

a little higher would not be terribly harder to read. But it's definitely not a big deal.

src/PermissionsContext.tsx
```diff
@@ -23,25 +23,32 @@ const initialENPermissionStatus: ENPermissionStatus = [
...
   }
+
+  return ENStatus.UNAUTHORIZED_DISABLED
 }
-export type ENAuthorizationEnablementStatus = {
-  authorized: boolean
-  enabled: boolean
-  authorizedAndEnabled: boolean
+
+export enum ENStatus {
```
johnschoeman, line 45: Yeah! this seems like a better fit for our use case.

an alternative would be a union type, which would just change the look n feel and not the type safety:
```typescript
export type ENStatus = "AuthorizedEnabled" | "AuthorizedDisabled" | "UnauthorizedDisabled"
```

src/Home/index.spec.tsx
```diff
@@ -322,11 +289,7 @@ describe("Home", () => {
   describe("When exposure notification permissions are unauthorized", () => {
     describe("and the platform is iOS", () => {
       it("shows an unauthorized alert", () => {
-        const isENAuthorizedAndEnabled: ENAuthorizationEnablementStatus = {
-          authorized: false,
-          enabled: false,
-          authorizedAndEnabled: false,
-        }
+        const isENAuthorizedAndEnabled = ENStatus.UNAUTHORIZED_DISABLED
```
johnschoeman, line 292:
```suggestion
        const enStatus = ENStatus.UNAUTHORIZED_DISABLED
```

src/Home/ProximityTracingActivationStatus.tsx
```diff
@@ -15,13 +14,10 @@ export const ProximityTracingActivationStatus: FunctionComponent = () => {
...
   const { t } = useTranslation()
 
-  const {
-    authorized: isAuthorized,
-    authorizedAndEnabled: isProximityTracingOn,
-  } = exposureNotifications.status
+  const { status } = exposureNotifications
 
   const handleOnPressFix = () => {
-    if (!isAuthorized && isPlatformiOS()) {
```
johnschoeman: do we need to keep the isPlatformiOS check here?

src/PermissionsContext.tsx
aledustet, line 45: types are cool 😎

devinjameson: 💯

src/Home/ProximityTracingActivationStatus.tsx
amandabeiner: We were thinking of that, but I don't think so? If the concept of having to "authorize" ENs is specific to iOS, the status should never be UNAUTHORIZED on android. Can you think of a reason that might be the case?

devinjameson:
I wonder if it's worth exposing some helpers from the `PermissionsContext` like:

```
const isENAuthorized = (status: ENStatus): boolean => {
  return status === ENStatus.AUTHORIZED_DISABLED || ENStatus.AUTHORIZED_ENABLED
}

const isENEnabled = (status: ENStatus): boolean => {
  return status === ENStatus.AUTHORIZED_ENABLED
}
```

src/Home/LocationActivationStatus.tsx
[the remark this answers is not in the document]
amandabeiner: Oooh I got it. That makes sense, thanks for clarifying. I will make this change!

PR #291, opened 2020-08-27
Pull exposure keys every 12 hours

juancruzgs:
#### Description:

This PR lowers the pull interval to 12 hours to avoid having to wait too much time to get an exposure notification.

#### Linked issues:

https://trello.com/c/TE9xVKd1/451-pull-keys-2x-per-day

#### How to test:

Check that the time shown in Exposure History screen is updated every 12 hours.


juancruzgs:
Good question.
According to the [documentation](https://developers.google.com/android/exposure-notifications/exposure-notifications-api#methods):

> Calls to `provideDiagnosisKeys()` are limited to six per day for token=TOKEN_A.

The API no longer accepts a TOKEN as a parameter, but I think they still have the same rate limit.
It shouldn't be an issue because even if the job runs every 12 hours it won't do anything if it already submitted the keys to the API (that will still happen once a day)

PR #292, opened 2020-08-27
Add helper for displaying configurable copy per HA

aledustet:
Why:
----
There are three configurable options on the application that each authority will set up independently, those user-facing text bodies need to be localized and should follow the same language preference that the application is using. They need to live on the resources repo since they are private to each HA.

This Commit:
----
- Add the download copy script
- Add the helper to consume a key, and fetch it from the copy json
- Add default values to useAuthorityCopy
- Display content from authority copy on legal and about if present
- Change surrounding scrollview for legal and about screens
- Update download file script
- Include download_copy_methods on the set_ha script

Reviewers:
----
This approach assumes that the content of the copy file will be verified prior usage. Also assumes that the supported locales are in sync with the ones that are present on the specific authority configuration.
The approach follows constraints expressed on the trello card. 

The new version respects the current behavior and should not change anything except for the content of the screen after the app is built per authority.

There's an accompanying PR here https://github.com/Path-Check/pathcheck-mobile-resources/pull/30

Co-authored-by: Thomas Paresi<thomas.paresi@thoughtbot.com>

### DIFF
--- a/src/configuration/copy.json
+++ b/src/configuration/copy.json
@@ -1,0 +1,11 @@
+{
+  "welcome_message": {
+    "en": ""
+  },
+  "about": {
+    "en": ""
+  },
+  "legal": {
+    "en": ""
+  }
+}
--- a/bin/download_copy.sh
+++ b/bin/download_copy.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env ruby
+#
+# Download the copy json for a specific HA and places it on the configuration
+# folder
+#
+# Usage
+#
+#   bin/download_copy.sh <health-authority-label> <github-token>
+#
+# Example
+#
+#   bin/download_assets.sh guam <token>
+#
+# Requirements
+#
+# 1. Remote access to the environment repo
+# 2. A github personal access token saved in `.env`:
+#    https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token
+
+require "open3"
+require "dotenv"
+require_relative "./download_copy_methods"
+Dotenv.load
+
+HA_LABEL = ARGV[0]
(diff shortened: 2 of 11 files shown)


src/configuration/useCopy.ts
```diff
@@ -0,0 +1,15 @@
+import { useTranslation } from "react-i18next"
+
+import copy from "./copy.json"
+
+type CopyKey = "welcome_message" | "about" | "legal"
+
+const useCopy = (key: CopyKey): string => {
```
johnschoeman, line 7: What do you think about a more specific name?

```suggestion
const useAuthorityCopy = (key: CopyKey): string => {
```

src/configuration/copy.json
```diff
@@ -0,0 +1,14 @@
+{
+  "welcome_message": {
+    "en": "Welcome to\nPathcheck",
```
johnschoeman, line 3: so noticing that this one has the app name hard coded, meaning that if we ever change the app name then there will be multiple places this needs to be updated.  Likely a low risk as the app name likely wont change often, but wanted to point it out.

aledustet: Yeah, I was debating about adding interpolation to these, but then the hook would not be as simple as we have it. Definitely open to extending it. One question I have is how might we coordinate to get this pattern used across.

johnschoeman: I think its low enough risk of churn to make the simpler implementation of hard coding preferable. Something we can address if its actually turns out to be an issue.

src/configuration/copy.json
```diff
@@ -0,0 +1,14 @@
+{
```
devinjameson, line 1: Thoughts on naming this file something like `authorityCopy` or something else more specific?

PR #297, opened 2020-08-28
Lokalise updates from Aug 28

alpita-masurkar:
#### Description:
Waiting on some missing Greek and es_PR translations, but this is the latest from August 28


alpita-masurkar:
@aledustet, the run failed on this PR and I went through the logs to find this:
Execution failed for task ':app:mergeDebugResources'.
> /home/runner/work/gaen-mobile/gaen-mobile/android/app/src/main/res/values-es-r419: Error: Invalid resource directory name

es-r419 was generated. Seems like this name doesn't comply or android doesn't like this name for some reason. The simplest solution to make this run work probably is to delete the es-r419 file from this PR. We're not using the Latin America version for now and we can just delete the file for now and resolve the issue. If you have any other suggestions, let me know.
Thank you.

aledustet:
> @aledustet, the run failed on this PR and I went through the logs to find this:
> Execution failed for task ':app:mergeDebugResources'.
> 
> > /home/runner/work/gaen-mobile/gaen-mobile/android/app/src/main/res/values-es-r419: Error: Invalid resource directory name
> 
> es-r419 was generated. Seems like this name doesn't comply or android doesn't like this name for some reason. The simplest solution to make this run work probably is to delete the es-r419 file from this PR. We're not using the Latin America version for now and we can just delete the file for now and resolve the issue. If you have any other suggestions, let me know.
> Thank you.

Sounds about right, for the moment can we delete it from the LOKALISE project so the automation tool keeps working? I think it would be difficult to communicate to the rest of the team to keep an eye on this. I can look into why this ISO is not a valid one. Thanks

alpita-masurkar:
done. deleted the es-r419 android file. it should build as expected now and should be ready for merge.
i remember our discussion on es-r419, let me know if you find out why android didn't like this ISO format :/ thanks @aledustet ❤️

android/app/src/main/res/values-da/strings.xml
```diff
@@ -1,4 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources>
-    <!-- Channel label for notifications shown in App > Notifications > Path Check BT -->
+  <!-- Shown on app info, app switcher, etc -->
```
johnschoeman, line 3: Should these values of "PathCheck BT" be included?

aledustet: I think if it is marked as non-translatable they are not used on the app. Maybe we should remove those keys all -together to avoid pulling them. I can take a look.

alpita-masurkar: If they are not used and if we don't need them, then we should remove them.
If I only remove them from Lokalise, they will get pulled in again on the next push to Lokalise. If you guys are planning to remove them from the repo, that'd be better so that the next pull from Lokalise will clean these keys out from all other language files.

juancruzgs: I can confirm we are not using these 2 strings, I've removed them from the code in the past but we should remove them from Lokalise too

alpita-masurkar: Thanks @juancruzgs for confirming this.

PR #300, opened 2020-08-28
Update response handling for keys posting

aledustet:
Why:
----
When the server returns a response stating that the keys sent were already inserted but no revision token was sent, we are displaying a scary and hard to understand error. This operation should not be treated as an error since the keys were added and therefore the exposed devices were notified and it should happen only if a positive was reported with a different jurisdiction. There is a disproportion of occurrences of it given the nature of how we test.

This Commit:
----
- Modify request management to handle no-op responses
- Add some copy to the error for the no-op response to the users

### DIFF
--- a/src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
+++ b/src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
@@ -29,7 +29,13 @@ import {
   Layout,
 } from "../../styles"
 import Logger from "../../logger"
-import * as ExposureAPI from "../exposureNotificationAPI"
+import {
+  postDiagnosisKeys,
+  PostKeysError,
+  PostKeysNoOp,
+  PostKeysSuccess,
+  PostKeysFailure,
+} from "../exposureNotificationAPI"
 
 interface PublishConsentFormProps {
   hmacKey: string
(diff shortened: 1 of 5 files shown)


src/AffectedUserFlow/exposureNotificationAPI.ts
```diff
@@ -9,25 +9,42 @@ const defaultHeaders = {
   accept: "application/json",
 }
 
+const EXISTING_KEYS_SENT_RESPONSE = "no revision token, but sent existing keys"
```
aledustet, line 12: This one ties directly to the case where:

- No new keys were sent
- No revision token was sent

This happens when there is a new submission on the same device for a different jurisdiction on a very short amount of time(typically testing) There is a different scenario where some keys are inserted that would stop being treated as an error on a future update.

src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
```diff
@@ -65,19 +122,13 @@ const PublishConsentForm: FunctionComponent<PublishConsentFormProps> = ({
       revisionToken,
     )
     setIsLoading(false)
-    if (response.kind === "success") {
-      storeRevisionToken(response.body.revisionToken)
+    if ((response as PostKeysSuccess).revisionToken) {
```
johnschoeman, line 125: this casting jumps out as smell, should we introduce an intermediate representation for the possible responses?

src/AffectedUserFlow/exposureNotificationAPI.ts
```diff
@@ -40,7 +57,7 @@ export const postDiagnosisKeys = async (
   hmacKey: string,
   appPackageName: string,
   revisionToken: string,
-): Promise<NetworkResponse<PostKeysSuccess, PostKeysError>> => {
```
johnschoeman, line 43: Why are we moving away from the network response either type? do we not what to make a distinction between a success or error case?

src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
aledustet: What kind of smell? HMW avoid this typecasting on a union type?

src/AffectedUserFlow/exposureNotificationAPI.ts
aledustet: My reasoning is behind that a `NetworkResponse` implies that the component is interacting with the network layer, which we are abstracting here using the helper. Having it respond with the result of a network call hits me of as the component interacting with the network which shouldn't be? There is still a distinction between error and success, the error nature of the response includes that.

src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
aledustet: Do you mean like having a distinction amongst the kinds of responses like the type argument on the original Network Response?

aledustet: @johnschoeman I added a kind attribute to avoid type casting, IMO these are pretty similar, but if it improves readability I'm all for it

PR #302, opened 2020-08-28
Reference specific commit from pathcheck-mobile-resources

juancruzgs:
#### Description:

We need a way to check which configuration we are using for each build.
Instead of always getting the latest changes from `master` this PR allows us to target a specific commit.

Each time there's a configuration change in https://github.com/Path-Check/pathcheck-mobile-resources we will have to update the file `/bin/mobile-resources-commit` to reflect those changes in the app.

I also allows us to define a different configuration per branch (for `mn-release` branch for example)

#### Linked issues:

https://trello.com/c/RZKjDdRu/404-as-a-dev-team-we-have-a-way-to-version-a-build-to-the-correct-commit-to-the-pathcheck-mobile-resources-repo

#### How to test:

Run `./bin/set_ha.sh mn` to check that it reads the configuration from the commit specified in `/bin/mobile-resources-commit`


bin/helpers.rb
```diff
@@ -62,3 +62,12 @@ def validate_ha_label!(label, script_name)
 def valid_token(token)
   token.length == 40
 end
+
+def mobile_resources_commit()
+  _output, error, status = Open3.capture3("cat", "./bin/mobile_resources_commit")
```
juancruzgs, line 67: Maybe there's a better way to read the content of the file.
I used `cat` just because I'm not familiar with Ruby

bin/helpers.rb
```diff
@@ -62,3 +62,12 @@ def validate_ha_label!(label, script_name)
 def valid_token(token)
   token.length == 40
 end
+
+def mobile_resources_commit()
```
aledustet, line 66: I don't think you need the `()` for a method that doesn't have arguments

bin/helpers.rb
```diff
@@ -62,3 +62,12 @@ def validate_ha_label!(label, script_name)
...
+  _output, error, status = Open3.capture3("cat", "./bin/mobile_resources_commit")
+  return _output if status.success?
+
+  puts error
+  # Return master branch as default
+  "master"
+end
```
aledustet, line 73: How do you feel about adding a new line here for consistency's sake?

bin/helpers.rb
aledustet: I think it's perfectly fine, maybe the location of it since `bin` symbols executables, but definitely not a big deal.

bin/fetch_ha_env.sh
```diff
@@ -33,14 +33,17 @@ def fetch_env
   staging_env_source = ".env.bt.staging"
   release_env_source = ".env.bt.release"
 
+  resources_commit = mobile_resources_commit()
+  puts "Fetching configuration from pathcheck-mobile-resources commit #{resources_commit}"
+
   dev_env_url =
-  "https://#{token}@raw.githubusercontent.com/Path-Check/pathcheck-mobile-resources/master/environment/#{HA_LABEL}/.env.bt"
+  "https://#{token}@raw.githubusercontent.com/Path-Check/pathcheck-mobile-resources/#{resources_commit}/environment/#{HA_LABEL}/.env.bt"
```
aledustet, line 40: you could get rid of that variable if you inline the method, you don't have to add the `()` for a method call, if the definition of the method has no arguments

aledustet, line 40: Maybe we could extract the front of the URL into a variable and then use it? interpolated in the three variables? that way if we change it again we only change one spot?

bin/helpers.rb
juancruzgs: Thanks!
I'm using the `File` module now

juancruzgs:
This PR tries to fix 2 issues we are having
- We didn't know which config is used for each build
- Flags are removed/renamed from the config and they break builds for everyone until someone creates a PR to fix it in this repository

I think it's important to make it a **manual** process where someone updates the commit reference and creates a PR that everyone can review and make sure it compiles.
We could add a new script to manually update the commit reference and get the latest one instead of having to copy-paste it.

I'll add the commit reference to the release notes in AppCenter so everyone can read it.

I'm open to suggestions, we can change it if we feel we will usually forget to update the reference.
Are you ok with this implementation for now?

bin/helpers.rb
```diff
@@ -62,3 +62,11 @@ def validate_ha_label!(label, script_name)
 def valid_token(token)
   token.length == 40
 end
+
+def mobile_resources_commit
+  file = File.open("mobile_resources_commit")
+  commit = file.read.chomp # the chomp is to drop the `/n`
```
johnschoeman, line 68:
```suggestion
  commit = file.read.chomp
```

johnschoeman:
> I think it's important to make it a **manual** process where someone updates the commit reference and creates a PR that everyone can review and make sure it compiles.

Not 100% sold on this **needs** to be manual and im worried about people forgetting to update the commit hash, or miss copying it. That said, this is a significant improvement and solves the immediate problems you listed, so we should move forward with it and automate if need be later.

> Are you ok with this implementation for now?
100% !

PR #303, opened 2020-08-28
Only log errors on staging and dev builds

johnschoeman:
Why:
We would only log errors for staging and dev builds

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -2,7 +2,6 @@ import React, { FunctionComponent, useState, useEffect } from "react"
 import SplashScreen from "react-native-splash-screen"
 import "array-flat-polyfill"
 import env from "react-native-config"
-import Bugsnag from "@bugsnag/react-native"
 
 import MainNavigator from "./src/navigation/MainNavigator"
 import { ErrorBoundary } from "./src/ErrorBoundaries"
@@ -14,8 +13,9 @@ import {
 import { ConfigurationProvider } from "./src/ConfigurationContext"
 import { PermissionsProvider } from "./src/PermissionsContext"
 import { initializei18next, loadUserLocale } from "./src/locales/languages"
+import Logger from "./src/logger"
 
-Bugsnag.start()
+Logger.start()
 
 const App: FunctionComponent = () => {
   const [isLoading, setIsLoading] = useState(true)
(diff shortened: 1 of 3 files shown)


App.tsx
```diff
@@ -15,7 +15,9 @@ import { ConfigurationProvider } from "./src/ConfigurationContext"
 import { PermissionsProvider } from "./src/PermissionsContext"
 import { initializei18next, loadUserLocale } from "./src/locales/languages"
 
-Bugsnag.start()
+if (env.STAGING === "true" || __DEV__) {
```
aledustet, line 18: I believe that this check will also be needed then on the logger file since we could try to call `Bugsnag` on the wrong environment, maybe we could move this to an initializer on that class that sets a logger instance that calls to `Bugsnag` on staging and dev and uses the console otherwise.

johnschoeman: I think we are also missing the android side to not instantiate the bugsnag native library

juancruzgs: Android disables Bugsnag on release builds, so only staging and dev should have it enabled

johnschoeman: awesome, thanks @juancruzgs

ios/AppDelegate.m
```diff
@@ -43,7 +44,15 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   [[ExposureManager shared] registerBackgroundTask];
 
   [RNSplashScreen showSplash:@"LaunchScreen" inRootView:rootView];
+  
+  // Read dev environment variable
+#if DEBUG
```
johnschoeman, line 49: @mattThousand or @emilianojuan , can i get a quick gut check on this bit of code?

mattThousand: @johnschoeman yes this is what we want I believe - bugsnag will be enabled only for builds either running locally using the `debug` build configuration, or any build using `.env.staging`

emilianojuan: I think its ok maybe add @"STAGING" as a static string but you may have issues with swift/objc so its ok to me, also you can do [ReactNativeConfig envFor:@"STAGING"].boolValue if you dont want to use the Boolean literal

johnschoeman: awesome, thanks!

App.tsx
johnschoeman: @aledustet , went for a quick fix, i think we can revisit the initialization pattern in a future commit.

PR #308, opened 2020-09-01
Fix for Exposure History Screen State Bug

mattThousand:
### Why
Exposures stored in the native database were not visible in the UI after force quitting and re-opening the app

### This Commit
This commit accommodates the function signature `getCurrentExposures = async (): Promise<ExposureInfo>` in `nativeModule.ts` (had previously used a callback some weeks ago, the JS-layer functions calling the method were still using the callback although it had been changed) and also fetches exposures from the database on AppState change.

### DIFF
--- a/src/ExposureContext.tsx
+++ b/src/ExposureContext.tsx
@@ -10,6 +10,7 @@ import React, {
 import gaenStrategy from "./gaen"
 import { ExposureKey } from "./exposureKey"
 import { ExposureInfo } from "./exposure"
+import { AppState } from "react-native"
 
 type Posix = number
 const { exposureEventsStrategy } = gaenStrategy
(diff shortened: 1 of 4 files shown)


src/ExposureContext.tsx
```diff
@@ -73,12 +77,20 @@ const ExposureProvider: FunctionComponent = ({ children }) => {
     })
   }, [getLastDetectionDate])
 
-  const getCurrentExposures = useCallback(() => {
-    const cb = (exposureInfo: ExposureInfo) => {
-      setExposureInfo(exposureInfo)
+  useEffect(() => {
+    const handleAppStateChange = async () => {
+      const exposureInfos = await getCurrentExposures()
```
aledustet, line 82: I think the name tripped me because I forgot that `ExposureInfo` is a collection, maybe `exposureInfoData` would be ok?

PR #309, opened 2020-09-01
Wrap fetch call for publish keys on a timeout

aledustet:
Why:
----
The publish keys request sometimes can timeout, the request is hanging and it it failing to complete. We need a way to detect when the request took too long and show the failure to the user.

This Commit:
----
- Add the `fetchWithTimeout` utility
- Wrap the fetch on the `postDiagnosisKeys` with a timeout of 5 seconds
- Display proper timeout error message to users when the request times out

How to test:
----
Submit keys through the flow and it should work as usual, and if you can send a first-time request(takes at least a second) and disconnect your internet in the middle you should see a timeout error

### DIFF
--- a/src/AffectedUserFlow/PublishConsent/PublishConsentForm.spec.tsx
+++ b/src/AffectedUserFlow/PublishConsent/PublishConsentForm.spec.tsx
@@ -209,6 +209,48 @@ describe("PublishConsentScreen", () => {
       })
     })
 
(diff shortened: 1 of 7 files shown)


aledustet:
> Nice work!
> 
> Would a future iteration be allowing users to retry the request after a timeout failure?

I'm starting to work on doing the retries on certain responses now, should this retry happen automatically or when the user chooses to?

PR #310, opened 2020-09-01
Add ActivateLocation screen to onboarding for required Android devices

devinjameson:
Why: some Android devices need Location to be on in Settings in order for Exposure Notifications to work.

This commit:
- Adds an `ActivateLocation` screen to the `ActivationStack` if the device needs location on for exposure notifications to work
- Navigates to the `ActivateLocation` screen from the `ActivateProximityTracing` screen only if location is off and needed for exposure notifications to work

Screen Shot 2020-09-01 at 1 58 56 PM


src/navigation/ActivationStack.tsx
```diff
@@ -104,7 +119,7 @@ const ActivationStack: FunctionComponent = () => {
 
   return (
     <Stack.Navigator screenOptions={screenOptions}>
-      {activationSteps.map((step, idx) => {
+      {activationSteps?.map((step, idx) => {
```
johnschoeman, line 123: Whats the need for the optional operator here? wouldnt activationSteps always be defined as an array?

src/locales/ar.json
```diff
@@ -22,6 +22,9 @@
     "launch_get_started": "ابدأ",
     "launch_screen1_header": "طريق العودة إلى الوضع الطبيعي يبدأ من هنا"
   },
+  "onboarding": {
```
johnschoeman, line 25: why are all these translations included in this pr? should they be in a different commit/pr?

src/Activation/ActivateProximityTracing.tsx
```diff
@@ -59,6 +62,10 @@ const ActivateProximityTracing: FunctionComponent = () => {
     )
   }
 
+  const handleOnPressActivateProximityTracing = () => {
+    showProximityTracingAlert()
```
johnschoeman, line 66: 💯

src/navigation/ActivationStack.tsx
devinjameson: For some reason Typescript thinks it could be undefined. Is there a workaround for that?

johnschoeman: I would look into the type signature of Platform.select and see if theres a thread to follow there.

devinjameson: Good call, it needed a default.

src/locales/ar.json
devinjameson: Removed thank!

PR #311, opened 2020-09-01
Fix homepage copy issue

devinjameson:
Why: prior to this commit, we were not informing the relevant users that
they need Location enabled in the home page subheader copy.

Screen Shot 2020-09-01 at 2 20 04 PM

### DIFF
--- a/src/Home/index.tsx
+++ b/src/Home/index.tsx
@@ -41,7 +41,10 @@ const Home: FunctionComponent = () => {
   const style = createStyle(insets)
 
   const isBluetoothOn = useBluetoothStatus()
-  const { isLocationOffAndNeeded } = useHasLocationRequirements()
+  const {
+    isLocationNeeded,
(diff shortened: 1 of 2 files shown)


src/Home/index.tsx
```diff
@@ -62,9 +65,18 @@ const Home: FunctionComponent = () => {
   const headerText = appIsActive
     ? t("home.bluetooth.tracing_on_header")
     : t("home.bluetooth.tracing_off_header")
-  const subheaderText = appIsActive
-    ? t("home.bluetooth.all_services_on_subheader", { applicationName })
-    : t("home.bluetooth.tracing_off_subheader")
+  const subheaderText = () => {
+    switch (appIsActive) {
```
johnschoeman, line 69: why the switch here? could and if-else block work?

devinjameson: Good call.

PR #314, opened 2020-09-01
Introduce StatusBar component

devinjameson:
Why: prior to this commit, there were places within the app where screen
content would overlap with status bar content.

Using `SafeAreaView` does not result in the correct UI or functionality because
it results in challenging and non-intuitive cross-platform behavior.

The cross-platform behavior is more intuitive and predictable using a
plain `View`, which wraps the `StatusBar` component from React Native.
This `View`'s height is set using `insets.top` (provided by the `useSafeAreaInsets`
hook from `react-native-safe-area-context`). The background color on this `View`
effectively becomes the status bar's background color.

Notes:
- This should be used on all screens without a header bar because all screens will
be scrollable to support large accessibility text sizes. Most screens will use this
with the `primaryLightBackground` color. 
- A future commit may remove `useStatusBarEffect` and wrap all the status bar
styling in the new `StatusBar` component

This commit:
- Introduces a `StatusBar` component
- Uses `StatusBar` on `Home/index` and `ExposureHistory/History/index`
- Removes translucent status bars on android

**Before**
before

**After**
statusbar

### DIFF
--- a/__mocks__/react-native-safe-area-context.js
+++ b/__mocks__/react-native-safe-area-context.js
@@ -1,0 +1,1 @@
+export const useSafeAreaInsets = jest.fn().mockReturnValue({insets: {bottom: 0}})
--- a/__mocks__/react-native-safe-area-context.js
+++ b/__mocks__/react-native-safe-area-context.js
@@ -1,1 +1,1 @@
-export const useSafeAreaInsets = jest.fn().mockReturnValue({insets: {bottom: 0}})
+export const useSafeAreaInsets = jest.fn().mockReturnValue({ insets: { bottom: 0 } })
--- a/src/Home/index.spec.tsx
+++ b/src/Home/index.spec.tsx
@@ -2,7 +2,6 @@ import React from "react"
 import { Alert, Share } from "react-native"
 import { render, fireEvent, within } from "@testing-library/react-native"
 import { useNavigation } from "@react-navigation/native"
-import { useSafeAreaInsets } from "react-native-safe-area-context"
 import "@testing-library/jest-native/extend-expect"
 
 import Home from "."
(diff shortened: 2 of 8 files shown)


src/Home/index.spec.tsx
```diff
@@ -16,9 +15,6 @@ import { ConfigurationContext } from "../ConfigurationContext"
 
 jest.mock("@react-navigation/native")
 
-jest.mock("react-native-safe-area-context")
-;(useSafeAreaInsets as jest.Mock).mockReturnValue({ insets: { bottom: 0 } })
```
aledustet, line 20: nice!! mocks that are not relevant to most of the tests go great under the `mocks` folder 👏

__mocks__/react-native-safe-area-context.js
```diff
@@ -0,0 +1 @@
+export const useSafeAreaInsets = jest.fn().mockReturnValue({insets: {bottom: 0}})
```
amandabeiner: 💯

src/Home/index.spec.tsx
devinjameson: 😋

src/components/StatusBar.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Colors } from "../styles"
+
+interface StatusBarProps {
+  backgroundColor?: string
+}
+
+const StatusBar: FunctionComponent<StatusBarProps> = ({
+  backgroundColor = Colors.primaryLightBackground,
+}) => {
+  const insets = useSafeAreaInsets()
```
amandabeiner, line 14: Really like moving this logic into this component! So much neater :)

src/components/StatusBar.tsx
```diff
@@ -0,0 +1,39 @@
...
+  )
+}
+
+type StyleProps = {
+  insets: EdgeInsets
+  backgroundColor: string
+}
+
+const createStyle = ({ insets, backgroundColor }: StyleProps) => {
+  /* eslint-disable react-native/no-unused-styles */
```
johnschoeman, line 30: do we still need this eslint-disable?

src/components/StatusBar.tsx
devinjameson: 💯

src/components/StatusBar.tsx
devinjameson: Yes we do unfortunately…

PR #315, opened 2020-09-01
Retry post diagnosis request when necessary

aledustet:
Why:
----
Some responses from the post-diagnosis keys endpoint return with a retry HTTP status code. We should honor these and retry the request when necessary.

How:
----
A document was shared about which error values deemed a retry response. This was used to determine the responses, in the case of the post-diagnosis endpoint, only `internal_error` and the responses with redirect headers `429` and `503`

This Commit:
----
- Adds retry logic to postDiagnosisKeys request
- Add post keys error for an internal error message

Reviewers:
----
Not sure about the delay after getting the retry to be confirmed, there is an [header] in these requests according to the HTTP standard, could not see any of those on my testing perhaps is worth checking? 

[header]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

### DIFF
--- a/src/AffectedUserFlow/exposureNotificationAPI.spec.ts
+++ b/src/AffectedUserFlow/exposureNotificationAPI.spec.ts
@@ -9,6 +9,10 @@ import { fetchWithTimeout, TIMEOUT_ERROR } from "./fetchWithTimeout"
 jest.mock("./fetchWithTimeout")
 
 describe("postDiagnosisKeys", () => {
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
+
   it("executes requests with default headers and serialized data", async () => {
     const exposureKeys: ExposureKey[] = []
     const regionCodes: string[] = []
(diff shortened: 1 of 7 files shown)


src/AffectedUserFlow/exposureNotificationAPI.spec.ts
```diff
@@ -109,6 +113,81 @@ describe("postDiagnosisKeys", () => {
     })
   })
 
+  describe("on a retry response", () => {
+    it("retries the request and returns the response of the next not retry", async () => {
```
devinjameson, line 117: Is this correct or is there a typo? ➡️ `not retry`

aledustet: I think i missed the response at the end 😄

src/AffectedUserFlow/exposureNotificationAPI.ts
```diff
@@ -54,7 +53,125 @@ type PostKeysResponse = PostKeysSuccess | PostKeysNoOp
...
+  temporaryExposureKeys: ExposureKey[]
+  regions: RegionCode[]
+  verificationPayload: Token
+  hmackey: string
+  padding: string
+  appPackageName: string
+  revisionToken: string
+}
+
+class PostDiagnosisKeysRequest {
```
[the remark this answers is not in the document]
aledustet: Also to encapsulate the request running logic without repeating the arguments across the different methods, and to keep the local-only constants inside a container

PR #316, opened 2020-09-01
Use wrench icon instead of fix text on home screen

devinjameson:
Why: using text doesn't work because `fix` is much longer than three characters in many languages.

Screen Shot 2020-09-01 at 5 30 04 PM


src/Home/ActivationStatus.tsx
```diff
@@ -26,19 +26,26 @@ export const ActivationStatus: FunctionComponent<ActivationStatusProps> = ({
   const { t } = useTranslation()
 
   const bodyText = isActive ? t("common.enabled") : t("common.disabled")
-  const icon = isActive ? Icons.CheckInCircle : Icons.XInCircle
-  const iconFill = isActive ? Colors.success100 : Colors.danger75
+  const leftIcon = isActive ? Icons.CheckInCircle : Icons.XInCircle
```
devinjameson, line 29: Looking for thoughts: is this big enough that it should become a config object or is that overkill?

aledustet: Maybe it could become a constant? you could declare the `activeAssets` and `inactiveAssets`?  And then use it? but you would have to put thems somewhere and this component is not too big? Maybe it's worth separating? It's also not symmetrical(both left and right) so it might be ok to leave as is. Your call

src/Home/ActivationStatus.tsx
```diff
@@ -25,38 +25,59 @@ export const ActivationStatus: FunctionComponent<ActivationStatusProps> = ({
 }) => {
   const { t } = useTranslation()
 
-  const bodyText = isActive ? t("common.enabled") : t("common.disabled")
-  const icon = isActive ? Icons.CheckInCircle : Icons.XInCircle
-  const iconFill = isActive ? Colors.success100 : Colors.danger75
+  type Content = {
```
aledustet, line 28: maybe the type definition could live outside the component body since this will be transpiled away when we turn it into javascript?

src/Home/ActivationStatus.tsx
```diff
@@ -25,38 +25,59 @@ export const ActivationStatus: FunctionComponent<ActivationStatusProps> = ({
...
+    bodyText: string
+    leftIcon: string
+    leftIconFill: string
+    rightIcon: string
+    onPress: () => void
+    accessibilityLabel: string
+  }
+
+  const activeContent: Content = {
```
aledustet, line 37: nice

src/Home/ActivationStatus.tsx
devinjameson: Good call!

PR #317, opened 2020-09-01
Introduce SystemServicesContext

devinjameson:
Why: instead of using various hooks, such as `useBluetoothStatus` and `useHasLocationRequirements`, to determine the state of system services, we would like to have a `SystemServicesContext` from which we can pull the relevant system-related information. This will reduce the amount of imports and top-level hooks, thereby making the use of system-related data more ergonomic and reducing the number of mocks in our specs.

This commit:
- Introduces an `SystemServicesContext`
- Removes `useBluetoothStatus` and `useHasLocationRequirements`, instead using the new context
- Updates the relevant specs


src/ActivationContext.tsx
```diff
@@ -0,0 +1,120 @@
...
+    return isBluetoothEnabled()
+  }
+
+  useEffect(() => {
+    const determineBTStatus = async () => {
+      const status = await fetchBTStatus()
+      setIsBluetoothOn(status)
+    }
+    determineBTStatus()
+    AppState.addEventListener(determineOSListener(), () => determineBTStatus())
```
devinjameson, line 46: Reviewers: thoughts on wrapping this logic in a reusable function?

devinjameson:
@amandabeiner Kindly requesting your help tomorrow in validating that the event listener for location still functions properly. I would do it on my own phone, but I can't trigger a state change in Bluetooth because the EN system doesn't use CoreBluetooth.

src/ActivationContext.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const initialState: ActivationState = {
+  isBluetoothOn: false,
+  isLocationOn: false,
+  isLocationNeeded: true,
+}
+
+export const ActivationContext = createContext<ActivationState>(initialState)
+
+const ActivationProvider: FunctionComponent = ({ children }) => {
```
aledustet, line 28: Could we have a name like `SystemServicesProvider` probably not a big deal but I thought activation could be brought into another context. Food for thought, I think is a perfectly used name but maybe worth considering a variant.

src/Home/index.spec.tsx
```diff
@@ -336,16 +355,19 @@ describe("Home", () => {
...
-        ;(useHasLocationRequirements as jest.Mock).mockReturnValue({
-          isLocationOn,
-          isLocationNeeded,
-        })
-
-        const { getByTestId } = render(<Home />)
+    describe("and the location is off", () => {
+      it("location is shown as disabled", () => {
+        const { getByTestId } = render(
+          <ActivationContext.Provider
```
aledustet, line 361: 🔥

src/Activation/ActivateProximityTracing.tsx
```diff
@@ -24,7 +24,8 @@ const ActivateProximityTracing: FunctionComponent = () => {
   const navigation = useNavigation()
   const { applicationName } = useApplicationName()
   const { exposureNotifications } = usePermissionsContext()
-  const { isLocationOffAndNeeded } = useHasLocationRequirements()
+  const { isLocationOn, isLocationNeeded } = useActivationContext()
+  const isLocationOffAndNeeded = !isLocationOn && isLocationNeeded
```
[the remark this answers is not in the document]
devinjameson: I had it that way at first, but it introduced the possibility of impossible shenanigans like…

```
{
  isLocationOn: false,
  isLocationNeeded: true,
  isLocationOffAndNeeded: false,
}

… when mocking the provider. Is there a way around that?

File: src/ActivationContext.tsx (hunk 2 above)
Comment (devinjameson) [reply src/ActivationContext.tsx:28]: I like that name a lot more, will update!

File: src/Activation/ActivateProximityTracing.tsx (hunk 4 above)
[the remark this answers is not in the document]
Comment (aledustet) [reply src/Activation/ActivateProximityTracing.tsx:28]: Sadly, I don't think there is unless we change the `isLocationOffAndNeeded` from a property to a function that returns:

``` typescript
const isLocationOffAndNeeded = (): boolean => {
 return !isLocationOn && isLocationNeeded;
}
```
Even though the nature of how context providers are defined you could still override the function when creating the provider and this will make uses of that property a bit weird since we would have to do `isLocationOffAndNeeded()` to use it, probably not worth it.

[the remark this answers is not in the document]
devinjameson: Thanks for the explanation. I wish there were a better way to handle stuff like this, as we've run into this exact dilemma on a few occasions now.

App.tsx
```diff
@@ -46,7 +47,9 @@ const App: FunctionComponent = () => {
             >
               <PermissionsProvider>
                 <ExposureProvider>
-                  <MainNavigator />
+                  <SystemServicesProvider>
```
johnschoeman, line 50: I might put SystemServices around Exposures as it feels like the exposure logic is dependent on knowing about the system

src/SystemServicesContext.tsx
```diff
@@ -0,0 +1,135 @@
...
+} from "react"
+import { AppState, Platform } from "react-native"
+
+import { isBluetoothEnabled } from "./gaen/nativeModule"
+import { isLocationEnabled } from "./gaen/nativeModule"
+import { doesDeviceSupportLocationlessScanning } from "./gaen/nativeModule"
+
+export interface SystemServicesState {
+  isBluetoothOn: boolean
+  isLocationOn: boolean
```
johnschoeman, line 16: for iOS would this be not known?

so should we have this be more like
```typescript
isLocationOn: boolean | "unknown"
```

src/SystemServicesContext.tsx
```diff
@@ -0,0 +1,135 @@
...
+      determineIsBluetoothOn()
+    }
+    addListener(handleAppStateChange)
+    return removeListener(handleAppStateChange)
+  }, [])
+
+  const fetchIsLocationOn = async (): Promise<boolean> => {
+    return Platform.select({
+      android: isLocationEnabled(),
+      ios: new Promise(() => true),
```
johnschoeman, line 57: Yeah this strikes me as a bit odd.  We are saying location is on, but we in fact do not know one way or the other.

Not sure its worth updating at this point, but I would lean towards trying to encode what we know about the world more directly

src/SystemServicesContext.tsx
devinjameson: The context is set up such that this returns `true` for iOS which seems correct to me.

src/SystemServicesContext.tsx
devinjameson: My thinking here is that it doesn't location is, by default, always enabled on iOS, and it doesn't really add anything to encode the fact that we don't know. It also may make working with this value on the view layer more challenging because we'd have to check if this is false, true, or "unknown", which will behave the same as it being true. Is there something off with my thinking there?

devinjameson: I'll take a look at how this would change if we encode this as unknown and circle back. Curious…

devinjameson: Well, turns out we do know whether location is on on iOS, so we can just use `isLocationEnabled()` directly. I won't change any behavior, but we'll have the actual value.

App.tsx
devinjameson: Good call!

src/SystemServicesContext.tsx
devinjameson: Turns out we do know on iOS because `isLocationEnabled()` works cross-platform. So this can just be a `boolean`.

PR #318, opened 2020-09-02
Fix status bar rendering issues

devinjameson:
Why: in order for the status bar to render correctly on all platforms, we need
to make use of both the `StatusBar` component from React Native as well
as a custom `useStatusBarEffect` hook.

This commit:
- Updates the `StatusBar` component
- Updates `useStatusBarEffect` to take a background color to render the
status bar correctly on Android
- Uses `StatusBar` in all the appropriate places
- Uses `useStatusBarEffect` in all the appropriate places

Notes:
- `useStatusBarEffect` is used to set the `barStyle` on both platforms
- `useStatusBarEffect` is used to set the status bar background color on
Android
- `StatusBar` is used to set the status bar background color on iOS
- `StatusBar` should not be used if a header exists on the screen via
React Navigation
- `StatusBar` should be used on any screen that does not have a header
because text _could_ overlap the status bar if the text size is made large
via accessibility settings

status

### DIFF
--- a/src/Activation/AcceptEula.tsx
+++ b/src/Activation/AcceptEula.tsx
@@ -12,9 +12,8 @@ import { useNavigation } from "@react-navigation/native"
 import { SvgXml } from "react-native-svg"
 
 import { Icons } from "../assets"
-import { GradientBackground, StatusBar } from "../components"
-import { GlobalText, Button } from "../components"
-import { ActivationScreens } from "../navigation"
+import { GlobalText, Button, GradientBackground } from "../components"
+import { ActivationScreens, useStatusBarEffect } from "../navigation"
 
 import {
   Forms,
@@ -27,6 +26,7 @@ import {
 import { useConfigurationContext } from "../ConfigurationContext"
 
 const AcceptEula: FunctionComponent = () => {
+  useStatusBarEffect("dark-content", Colors.primaryLightBackground)
   const configuration = useConfigurationContext()
   const [boxChecked, toggleCheckbox] = useState(false)
   const { t } = useTranslation()
@@ -54,7 +54,6 @@ const AcceptEula: FunctionComponent = () => {
 
   return (
     <>
-      <StatusBar backgroundColor={Colors.primaryLightBackground} />
       <GradientBackground gradient={Colors.gradientPrimary10}>
         <SafeAreaView style={style.container}>
           <GlobalText style={style.headerText}>
--- a/src/Activation/AcceptEula.tsx
+++ b/src/Activation/AcceptEula.tsx
@@ -12,7 +12,7 @@ import { useNavigation } from "@react-navigation/native"
 import { SvgXml } from "react-native-svg"
 
 import { Icons } from "../assets"
-import { GradientBackground } from "../components"
+import { GradientBackground, StatusBar } from "../components"
 import { GlobalText, Button } from "../components"
 import { ActivationScreens } from "../navigation"
 
@@ -24,11 +24,9 @@ import {
   Outlines,
   Typography,
 } from "../styles"
-import { useStatusBarEffect } from "../navigation"
 import { useConfigurationContext } from "../ConfigurationContext"
 
 const AcceptEula: FunctionComponent = () => {
-  useStatusBarEffect("dark-content")
   const configuration = useConfigurationContext()
   const [boxChecked, toggleCheckbox] = useState(false)
   const { t } = useTranslation()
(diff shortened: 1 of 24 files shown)


src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
```diff
@@ -147,103 +146,91 @@ const PublishConsentForm: FunctionComponent<PublishConsentFormProps> = ({
...
-              <View style={style.cancelButtonInnerContainer}>
-                <SvgXml
-                  xml={Icons.X}
-                  fill={Colors.black}
-                  width={Iconography.xSmall}
-                  height={Iconography.xSmall}
-                />
-              </View>
-            </TouchableOpacity>
+    <View style={style.outerContainer}>
```
aledustet, line 149: Should we have a `StatusBar` here since we are using the `useStatusBarEffect` hook?

src/More/LanguageSelection.tsx
```diff
@@ -58,33 +58,35 @@ const LanguageSelection: FunctionComponent = () => {
...
-          onPress={navigation.goBack}
-        >
-          <SvgXml
-            xml={Icons.XInCircle}
-            fill={Colors.neutral30}
-            width={Iconography.small}
-            height={Iconography.small}
-          />
-        </TouchableOpacity>
+    <>
```
aledustet, line 61: Should we have `StatusBar` here?

src/More/Menu.tsx
```diff
@@ -7,69 +7,72 @@ import env from "react-native-config"
...
-          onPress={() => navigation.navigate(MoreStackScreens.About)}
-        />
-        <SettingsListItem
-          label={t("screen_titles.legal")}
-          onPress={() => navigation.navigate(MoreStackScreens.Legal)}
-          lastItem
-        />
-      </View>
-      {showDebugMenu ? (
+    <>
```
aledustet, line 31: Should this component have a `StatusBar`?

src/Activation/AcceptEula.tsx
```diff
@@ -55,43 +53,46 @@ const AcceptEula: FunctionComponent = () => {
...
-            <GlobalText style={style.checkboxText}>
-              {t("onboarding.eula_agree_terms_of_use")}
-            </GlobalText>
-          </TouchableOpacity>
-          <Button
-            onPress={handleOnPressNext}
-            disabled={!boxChecked}
-            label={t("common.continue")}
+    <>
+      <StatusBar backgroundColor={Colors.primaryLightBackground} />
```
amandabeiner, line 57: Just asking, does this mean that we will want to manually render a StatusBar on every screen where we want one, rather than having it be the default and then trying to hide it via react-native?

amandabeiner (approved):
LGTM!

aledustet (approved):
Seems correct, I'm confused on some components that do render a `StatusBar` and use the hook for the status bar effect and some don't. Maybe worth checking those since it seemed like a fragment was added but no `StatusBar`

src/Activation/AcceptEula.tsx
```diff
@@ -55,43 +53,46 @@ const AcceptEula: FunctionComponent = () => {
...
-            />
-            <GlobalText style={style.checkboxText}>
-              {t("onboarding.eula_agree_terms_of_use")}
-            </GlobalText>
-          </TouchableOpacity>
-          <Button
-            onPress={handleOnPressNext}
-            disabled={!boxChecked}
-            label={t("common.continue")}
+    <>
```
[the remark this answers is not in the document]
devinjameson: Dammit Git!

src/More/LanguageSelection.tsx
devinjameson: Nope because this screen pops up as a card-style modal -- but I will remove the fragment, good catch!

src/More/Menu.tsx
devinjameson: Forgot to remove the fragment! No `StatusBar` because this screen has a header.

src/Activation/AcceptEula.tsx
devinjameson: I believe the status bar will show up anyways, but on most screens we will still want to to use `useStatusBarEffect` and on many we will want to use `StatusBar` to achieve the correct styling.

devinjameson: If that doesn't answer your question lmk and I can explain more clearly!

src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
devinjameson: Nope because this screen has a header!

devinjameson:
`Seems correct, I'm confused on some components that do render a StatusBar and use the hook for the status bar effect and some don't. Maybe worth checking those since it seemed like a fragment was added but no StatusBar`

I had missed a few things so if you go back and look now it should all make sense.

PR #321, opened 2020-09-03
Display authority links on about and legal screen

aledustet:
Why:
----

Authorities will need to display a set of links under the about and legal screens information.

How:
----

By using the values on the new overrides file `configuration/links.json`, we will go through the contents of each file and display the links under the required screen.

This Commit:
----
- Add useAuthorityLinks helper
- Add tests and a default locale to the useAuthorityCopy hook
- Display authority links on legal and about screens
- Add download links methods for CI

Caveat:
----

This depends on the PR's for adding these necessary configuration files

Screen Shot 2020-09-03 at 10 18 49 AMScreen Shot 2020-09-03 at 10 18 54 AM


src/More/ExternalLink.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { FunctionComponent } from "react"
+import { StyleSheet, TouchableOpacity, Linking } from "react-native"
+
+import { GlobalText } from "../components"
+import { Typography } from "../styles"
+
+type ExternalLinkProps = {
```
devinjameson, line 7: What are your thoughts on using `interface` here for consistency? Based on what I see elsewhere in the app, when we type the props object of a `FunctionComponent` we use `interface`. Totally a stylistic preference but curious to get your thoughts!

devinjameson (approved):
Design looks 👍 !

src/More/ExternalLink.tsx
aledustet: The idea for using types vs interfaces is that types don't merge declarations, for example if you have a type `A` defined here and a type `A` defined lower in the file, you will see an error from the compiler and if you do it with interfaces typescript will merge the declarations if it's possible. I don't see a strong reason to use type vs interfaces on any of these files, especially since none of those types are exported but since types are the stronger declaration I tend to reach for them. Also types can be used as "values" in some scenarios. No problem to change it back to an interface if we have been using interfaces.

devinjameson: Makes sense! Yeah totally a personal preference, thanks for sharing, very helpful/educational :)

src/configuration/useAuthorityLinks.spec.ts
```diff
@@ -0,0 +1,51 @@
+import useAuthorityLinks from "./useAuthorityLinks"
+
+jest.mock("./links.json", () => {
+  return {
+    about: [
+      {
+        url: "firstUrl",
```
aledustet, line 7: thoughts on this structure? Should we accommodate that links can be localized also? Maybe there's a query string on the endpoints? And the website where the url is pointing at supports internationalization? @devinjameson @johnschoeman ?

johnschoeman: All good questions, we dont know the answers to what we need to support in that detail yet.

my guess is that we will start with just a single url, so just localization of the labels.  so we might want to update the structure to accommodate this, but tbh i would start with just defaulting the en and pick up the localization in a follow up commit.

so this structure will likely want to change, but only after we know more so i might start with a simpler one for now such as:
```
about: [
  {
    url: "firstUrl",
    label: "firstLabel"
  },
  {
    url: "secondUrl",
    label: "secondLabel"
  }
]
```
but what you have here looks fine to me.

src/configuration/useAuthorityLinks.ts
```diff
@@ -0,0 +1,39 @@
...
+        const localizedLabel = label[localeCode] || label[DEFAULT_LOCALE]
+        return { url, label: localizedLabel }
+      })
+    return linksOverride
+  } catch {
+    return []
+  }
+}
+
+export default useAuthorityLinks
```
johnschoeman, line 39: should we be using the `useThing` convention if this isnt using any react hooks?

src/configuration/useAuthorityLinks.ts
```diff
@@ -0,0 +1,39 @@
...
+  label: TranslationValue
+}
+type LinkKeys = "about" | "legal"
+type AuthorityLinksContent = Record<LinkKeys, LinkData[]>
+type DisplayableLink = {
+  url: string
+  label: string
+}
+
+const useAuthorityLinks = (
```
johnschoeman, line 16: This code looks great! but i wonder if we can simply a bit. for example, do we need to check the label here? could we just pass in an array of linkData? and do the check of if the config provides a sensible list in the component?

src/configuration/useAuthorityLinks.ts
```diff
@@ -0,0 +1,39 @@
...
+  localeCode: string,
+): DisplayableLink[] => {
+  try {
+    const authorityLinks = links as AuthorityLinksContent
+    const linksForKey = authorityLinks[key] || []
+
+    const linksOverride = linksForKey
+      .filter(({ label }) => {
+        const localizedLabel = label[localeCode] || label[DEFAULT_LOCALE]
+        return localizedLabel.length > 0
```
johnschoeman, line 27: Do we need to be filtering for this? we have ultimate control over the data, so we could just not include any entries that dont have labels.

src/More/About.tsx
```diff
@@ -23,12 +25,17 @@ export const AboutScreen: FunctionComponent = () => {
     t("about.description", { applicationName, healthAuthorityName }),
   )
 
+  const authorityLinks = useAuthorityLinks("about", localeCode)
```
johnschoeman, line 28: What do you think about a flow more like:

```typescript
import links from "./links.json"

...

const aboutLinks = links.about
const displayableLinks = applyTranslations(aboutLinks, localeCode)
```

Having the importing of the data in the same spot as the apply the translation seems a bit overly abstracted to me.

johnschoeman (approved):
Code looks great!

Had a thought about treating this less like a hook, if youre interested.

src/More/About.tsx
johnschoeman: the link data could live in the config context as well. which might be a better place to map the data to apply the label translation.

src/configuration/useAuthorityLinks.ts
aledustet: We do have control over the data, but there is zero validation on the input. It's just a file that gets downloaded. Not sure where can we gain more control and shift the complexity away from here. Still seems like an external dependency that we don't control the constraints off. I thought that this could make the part out of our control less disruptive of the functionality. But I also see how it is introducing complexity to it.

src/configuration/useAuthorityLinks.ts
aledustet: Wouldn't this makes us have the same behavior scattered across multiple components? Would it be our responsibility to exercise that behavior in all those places instead of having the encapsulation here?

aledustet:
@johnschoeman I took a hack at separating the load and the translation logic, didn't move it to the configuration context because it looked like a lot to load the file at the beginning especially since those are local to the project. Let me know if it is better now. Since the method that makes the translation accepts the input I think we can refactor it in a future.

PR #324, opened 2020-09-03
Add eula link on the activation screen

aledustet:
Why:
----

The accept eula/accept terms screen on the activation stack has been removed from it. We need a way to display it based solely on the configuration variable DISPLAY_ACCEPT_TERMS_OF_SERVICE. Also the eula link is not getting any values to navigate to

This Commit:
----
- Add the configuration to the configuration context
- Point skip and get started from onboarding to `root` screen on
activation stack
- Determine to add or not accept terms of service screen to stack based
on configuration on the activation stack
- Consume eula url value from the configuration
- Show the eula link only if the eula url is available

Reviewers:
----
This depends of a PR on the resources repo. The `default` screen on the activation stack will always be the one on top of the list if the name is not the same one on `initialRoute`, we could choose to have it explicit

### DIFF
--- a/src/Activation/AcceptEula.tsx
+++ b/src/Activation/AcceptEula.tsx
@@ -36,14 +36,6 @@ const AcceptEula: FunctionComponent = () => {
     navigation.navigate(ActivationScreens.ActivateProximityTracing)
   }
 
-  const linkToPrivacyPolicy = async () => {
-    await Linking.openURL(configuration.healthAuthorityPrivacyPolicyUrl)
-  }
-
-  const linkToEula = async () => {
-    await Promise.resolve()
-  }
-
   const checkboxIcon = boxChecked
     ? Icons.CheckboxChecked
     : Icons.CheckboxUnchecked
@@ -59,11 +51,14 @@ const AcceptEula: FunctionComponent = () => {
           <GlobalText style={style.headerText}>
             {t("onboarding.terms_header_title")}
           </GlobalText>
-          <EulaLink
+          <DocumentLink
             docName={"onboarding.privacy_policy"}
-            onPress={linkToPrivacyPolicy}
+            url={configuration.healthAuthorityPrivacyPolicyUrl}
           />
-          <EulaLink docName={"onboarding.eula"} onPress={linkToEula} />
+          <DocumentLink
+            docName={"onboarding.eula"}
+            url={configuration.healthAuthorityEulaUrl}
+          />
           <View style={style.footerContainer}>
             <TouchableOpacity
               style={style.checkboxContainer}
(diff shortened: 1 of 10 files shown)


johnschoeman (approved):
LGMT

devinjameson:
> LGMT

Looks good me to. @johnschoeman

src/Activation/AcceptEula.tsx
```diff
@@ -41,7 +41,8 @@ const AcceptEula: FunctionComponent = () => {
   }
 
   const linkToEula = async () => {
```
devinjameson, line 43: Thoughts on naming this `handleOnPressLinkToEula` to stay consistent?

src/Activation/AcceptEula.tsx
```diff
@@ -41,7 +41,8 @@ const AcceptEula: FunctionComponent = () => {
...
-    await Promise.resolve()
+    configuration.healthAuthorityEulaUrl &&
```
devinjameson, line 44: Just curious why the double check on `configuation.healthAuthorityEulaUrl`?

src/navigation/ActivationStack.tsx
```diff
@@ -121,6 +122,13 @@ const ActivationStack: FunctionComponent = () => {
 
   return (
     <Stack.Navigator screenOptions={screenOptions}>
+      <Stack.Screen
```
devinjameson, line 125: Should this screen be part of the activation steps or am I missing something?

devinjameson: E.g.

```
const acceptEula: ActivationStep = {
  screenName: ActivationScreens.AcceptEula,
  component: AcceptEula,
}

const activationStepsAndroid: ActivationStep[] = isLocationNeeded
  ? [acceptEula, activateProximityTracing, activateLocation]
  : [acceptEula, activateProximityTracing]

const activationStepsIOS: ActivationStep[] = [
  acceptEula,
  activateProximityTracing,
  notificationPermissions,
]
```

src/Activation/AcceptEula.tsx
```diff
@@ -63,7 +64,9 @@ const AcceptEula: FunctionComponent = () => {
             docName={"onboarding.privacy_policy"}
```
devinjameson, line 64: Should this component be named `EulaLink` or just something like `Link` because it handles the Eula and Privacy Policy links?

src/Activation/AcceptEula.tsx
aledustet: That's the thing with encapsulating the method outside, Typescript has no idea that the only place this is being called it's surrounded by a null check 😢

src/navigation/ActivationStack.tsx
aledustet: Good catch, It probably is. This will update the right side with step x out of y right? In this PR I'm gonna remove this part since we are not activating the accept eula screen just yet, but I'll open a new one with this next. 👍

src/Activation/AcceptEula.tsx
devinjameson: Ah I see, dammit Typescript!

src/navigation/ActivationStack.tsx
devinjameson: Yes! Cool cool.

aledustet:
This has been updated to accomodate for a configuration flag to display the accept terms of service screen on the flow or not. @johnschoeman @devinjameson

src/Activation/AcceptTermsOfService.tsx
```diff
@@ -59,11 +51,14 @@ const AcceptEula: FunctionComponent = () => {
           <GlobalText style={style.headerText}>
             {t("onboarding.terms_header_title")}
           </GlobalText>
-          <EulaLink
+          <DocumentLink
```
johnschoeman, line 54: 💯

johnschoeman (approved):
continues to LGMT!

PR #326, opened 2020-09-03
Fix unresolved promise issue in SystemServicesContext

devinjameson:
`doesDeviceSupportLocationlessScanning()` is not present in the iOS version of the native module and was the root cause of the warning. This is technically still an issue, but it doesn't matter as the iOS builds shouldn't be invoking the function at all.

### DIFF
--- a/src/SystemServicesContext.tsx
+++ b/src/SystemServicesContext.tsx
@@ -59,7 +59,7 @@ const SystemServicesProvider: FunctionComponent = ({ children }) => {
     }
   }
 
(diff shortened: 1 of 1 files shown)


johnschoeman (approved):
LGTM, I would just add a description to the commit message that mentions the fact that `doesDeviceSupportLocationlessScanning()` is not present in the ios version of the native module and was the root cause of the warning and that it is still technically an issue, but doesnt matter as the ios builds shouldnt be invoking that function at all.

PR #327, opened 2020-09-03
Enable user to activate proximity tracing from home screen

devinjameson:
## Why

Prior to this commit, on iOS, if the user did not enable exposure notifications in the onboarding flow, there was no way for them to enable exposure notifications at all.

This state was possible because, in the onboarding flow, if the user did not enable exposure notifications, exposure notification permissions were _never requested in the native layer_. This put the app in a state where it was not possible for the user to _authorize_ exposure notifications in settings.

## This Commit

- Removes the "Enable Exposure Notifications" alert on the javascript layer because the native enable exposure notifications alert makes it redundant
- Makes it so exposure notification permissions are always requested when the user tries to fix an issue with proximity tracing not working. We need to do this because it is possible for the user to not request exposure notifications during the onboarding flow. In this case, when the user tries to enable proximity tracing on the home screen, we need to first request exposure notifications, _then_ inform the user that the Pathcheck app needs to be authorized in settings in order for proximity tracing to work. If we do not first request exposure notifications, the user will find a blank screen in Settings > Pathcheck > Exposure Notifications and thus not be able to authorize the app.
- Updates various specs to test this updated functionality
- Updates the alert copy we present to the user to better inform them on how to authorize sharing of exposure information in settings
- Refactors the exposure notifications request function to use a promise instead of a callback

## Screenshots

Screen Shot 2020-09-03 at 3 21 23 PM
Screen Shot 2020-09-03 at 3 21 43 PM
Screen Shot 2020-09-03 at 3 21 57 PM
Screen Shot 2020-09-03 at 3 22 04 PM

### DIFF
--- a/src/Activation/ActivateProximityTracing.tsx
+++ b/src/Activation/ActivateProximityTracing.tsx
@@ -25,8 +25,6 @@ const ActivateProximityTracing: FunctionComponent = () => {
   const isLocationOffAndNeeded = !isLocationOn && isLocationNeeded
 
   const { exposureNotifications } = usePermissionsContext()
-  const { status } = exposureNotifications
-  console.log(status)
 
   const navigateToNextScreen = () => {
     if (Platform.OS === "ios") {
--- a/src/Activation/ActivateProximityTracing.tsx
+++ b/src/Activation/ActivateProximityTracing.tsx
@@ -1,7 +1,6 @@
 import React, { FunctionComponent } from "react"
 import {
   Platform,
-  Alert,
   ScrollView,
   SafeAreaView,
   View,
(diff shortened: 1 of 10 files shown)


src/Activation/ActivateProximityTracing.tsx
```diff
@@ -46,25 +36,13 @@ const ActivateProximityTracing: FunctionComponent = () => {
...
-        },
-      ],
-    )
+  const handleOnPressActivateProximityTracing = async () => {
+    await exposureNotifications.request()
+    navigateToNextScreen()
   }
 
-  const handleOnPressActivateProximityTracing = () => {
-    showProximityTracingAlert()
```
johnschoeman, line 67: So much better

johnschoeman (approved):
LGTM, no comments!

PR #330, opened 2020-09-03
Fix line height on code input

amandabeiner:
#### Description:
Why:
There is currently a bug on android where you cannot see the code you have inputted.

This commit:
Removes a `lineHeight: 0`, on the code input, which was causing the number code to not display.

#### Screenshots:
Before:
not working

After:
working


aledustet (approved):
🚢 :shipit:

johnschoeman:
have you pulled recently? I believe that this one does the same: https://github.com/Path-Check/gaen-mobile/commit/fe205b77d99aaa29dffa2468cc5d59b4b084f03d ?

devinjameson:
I believe this is fixed on develop, did I miss something?

aledustet:
I was gonna say that this line is not on my branch

aledustet:
Deja vu

PR #332, opened 2020-09-03
Remove title from the activation stack

aledustet:
Why:
----
The navigation title on the activation stack is all the way to the left and truncated.

This Commit:
----
- Remove the title from the activation screens

Reviewers:
----
Not sure if the solution is to remove it or to center it and make it fit, but it seemed like the affected screens all had an inline title.

Screen Shot 2020-09-03 at 4 39 44 PMScreen Shot 2020-09-03 at 4 40 28 PM

PR #333, opened 2020-09-04
Fix App Setup header display regression

devinjameson:


aledustet (approved):
Should I change/remove #332 since you updated this?

PR #334, opened 2020-09-04
Ensure platform is iOS before showing unauthorized alert

devinjameson:
## Why
Prior to this commit, we were not ensuring the platform was iOS before showing an unauthorized alert on the home screen.

## This Commit
- Checks to see if the platform is iOS before showing an unauthorized alert.
- Updates the home screen spec to test this behavior
- Uses `isPlatformiOS()` to avoid having to mock `Platform`

### DIFF
--- a/src/Home/ProximityTracingActivationStatus.tsx
+++ b/src/Home/ProximityTracingActivationStatus.tsx
@@ -7,6 +7,7 @@ import { usePermissionsContext, ENStatus } from "../PermissionsContext"
 import { HomeScreens } from "../navigation"
 import { ActivationStatus } from "./ActivationStatus"
 import { useApplicationName } from "../hooks/useApplicationInfo"
+import { isPlatformiOS } from "../utils"
 
 export const ProximityTracingActivationStatus: FunctionComponent = () => {
   const navigation = useNavigation()
@@ -37,7 +38,7 @@ export const ProximityTracingActivationStatus: FunctionComponent = () => {
 
   const handleOnPressFix = () => {
     exposureNotifications.request()
-    if (isUnauthorized) {
+    if (isUnauthorized && isPlatformiOS()) {
       showUnauthorizedAlert()
     }
   }
--- a/src/Home/ProximityTracingActivationStatus.tsx
+++ b/src/Home/ProximityTracingActivationStatus.tsx
@@ -49,10 +49,12 @@ export const ProximityTracingActivationStatus: FunctionComponent = () => {
     navigation.navigate(HomeScreens.ProximityTracingInfo)
   }
 
+  const isENEnabled = status === ENPermissionStatus.ENABLED
+
   return (
     <ActivationStatus
       headerText={t("home.bluetooth.proximity_tracing_header")}
-      isActive={status === ENPermissionStatus.ENABLED}
+      isActive={isENEnabled}
       infoAction={handleOnPressShowInfo}
       fixAction={handleOnPressFix}
       testID={"home-proximity-tracing-status-container"}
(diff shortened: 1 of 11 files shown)


src/Home/ProximityTracingActivationStatus.tsx
```diff
@@ -37,7 +38,7 @@ export const ProximityTracingActivationStatus: FunctionComponent = () => {
 
   const handleOnPressFix = () => {
     exposureNotifications.request()
-    if (isUnauthorized) {
+    if (isUnauthorized && isPlatformiOS()) {
```
aledustet, line 41: Weird question here, but do we have a scenario where the `isUnauthorized` will be used not in conjunction with the platform check? Should we move this check closer to that provider?

aledustet (approved):
🚦 => 🟢

src/Home/ProximityTracingActivationStatus.tsx
johnschoeman: plus one to seeing if we can hoist it higher up the control flow, but if thats too much of a refactor what do you think about colocating the logic? i.e.

```typescript
const isUnauthorized = isPlatformiOS() && status === ENStatus.UNAUTHORIZED_DISABLED
```

johnschoeman (approved):

src/PermissionsContext.tsx
```diff
@@ -34,7 +35,9 @@ const toENStatus = (enPermissionStatus: ENPermissionStatus): ENStatus => {
   const isEnabled = enPermissionStatus[1] === "ENABLED"
 
   if (!isAuthorized && !isEnabled) {
-    return ENStatus.UNAUTHORIZED_DISABLED
+    return isPlatformiOS()
```
devinjameson, line 38: @johnschoeman @aledustet What are your thoughts on this? This means on Android the possible states for `exposureNotifications.status` are `ENStatus.AUTHORIZED_DISABLED` and `ENStatus.AUTHORIZED_ENABLED`. So we only expose an unauthorized state from the provider on iOS.

src/Home/ProximityTracingActivationStatus.tsx
devinjameson: Good idea, tagged you both in a comment on a new commit.

src/PermissionsContext.tsx
johnschoeman: That logically makes sense to me, but i find the naming a bit confusing or at least the motivation behind this is hard to glean from reading the code.  So I wonder if the ENStatus enum values are too iOS focused?

devinjameson: I agree @johnschoeman, I think https://github.com/Path-Check/gaen-mobile/pull/334/commits/b746bfa292d769abd38e68c90bbc5cc53cfe631a is a step in the right direction, but I'm confused as to why we can end up in a state where exposure notification permissions are unauthorized on Android. Shouldn't we always be authorized on Android, in which case we only care if we are enabled or disabled?

src/PermissionsContext.tsx
```diff
@@ -18,40 +18,38 @@ import { isPlatformiOS } from "./utils"
...
+): ENPermissionStatus => {
   const isAuthorized = enPermissionStatus[0] === "AUTHORIZED"
   const isEnabled = enPermissionStatus[1] === "ENABLED"
 
-  if (!isAuthorized && !isEnabled) {
+  if (!isAuthorized) {
     return isPlatformiOS()
-      ? ENStatus.UNAUTHORIZED_DISABLED
-      : ENStatus.AUTHORIZED_DISABLED
+      ? ENPermissionStatus.NOT_AUTHORIZED
```
devinjameson, line 41: @juancruzgs @johnschoeman Looks like my last comment was deleted due to a rebase. Is there a way to always return an authorized state on Android to make this check unnecessary?

devinjameson: Merging this because it works but would be good to sort this out sometime.

PR #335, opened 2020-09-04
Handle rejected promise on request exposure notification permissions

devinjameson:
Why: prior to this commit, a warning showed on iOS when the promise was rejected when requesting exposure notification permissions.

This commit:
- Handles the promise rejection when requesting exposure notification permissions
- Updates the `requestAuthorization` method in the native module to return `Promise<void>` instead of 'Promise<string>`

### DIFF
--- a/src/PermissionsContext.tsx
+++ b/src/PermissionsContext.tsx
@@ -83,7 +83,7 @@ export interface PermissionStrategy {
     cb: (status: ENPermissionStatus) => void,
   ) => { remove: () => void }
   check: (cb: (status: ENPermissionStatus) => void) => void
-  request: () => Promise<string>
+  request: () => Promise<void>
 }
 
 const PermissionsProvider: FunctionComponent = ({ children }) => {
@@ -139,7 +139,7 @@ const PermissionsProvider: FunctionComponent = ({ children }) => {
   }
 
   const requestENPermission = async () => {
-    permissionStrategy.request()
+    permissionStrategy.request().catch(() => {})
   }
 
   const requestNotificationPermission = async () => {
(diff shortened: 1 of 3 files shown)


src/PermissionsContext.tsx
```diff
@@ -139,7 +139,7 @@ const PermissionsProvider: FunctionComponent = ({ children }) => {
   }
 
   const requestENPermission = async () => {
-    permissionStrategy.request()
+    permissionStrategy.request().catch(() => {})
```
aledustet, line 142: does the failure means something? Should we update any values?

aledustet (approved):
This looks ready to 🚢 . I have one question, seems to me that this is fixing a warning on develop of an unhandled promise rejection right? Should we look at who's calling in case there's a double request? I'm mostly coming from a place of no background here 😅

src/PermissionsContext.tsx
johnschoeman: Plus one, swallowing errors is a code smell. do we know why this is a failed promise?

devinjameson: My understanding is that this only happens when exposure notifications are requested on iOS, but the app is not authorized to share exposure information. So the attempt to enable ENs fails. I don't think we need to update anything if that is the case, but I may be missing something.

johnschoeman: So it might make sense to check if the app is iOS and authorized before making the request to enable app for en.

The problem with the generic catch(() => {}) is that if there ever is an unexpected error in the request, (like an api change or a typo), then we would have a very hard time debuging it or even finding out that it happened.

devinjameson: Got it thanks @johnschoeman, this should fix it: https://github.com/Path-Check/gaen-mobile/pull/335/commits/b4f82c69008af7d2af97e367780c7a7cc79ca620

PR #336, opened 2020-09-04
Add link to configuration in release notes

juancruzgs:
#### Description:

This PR adds a link to the configuration used for each build in the release notes.
It also removes some fastlane commands we are no longer using.

#### Screenshots:

image


johnschoeman (approved):
Great idea!

PR #337, opened 2020-09-04
Add Accept Eula to activation flow

aledustet:
Why:
----

The accept terms of service screen is not shown on the flow because we
don't have links in place for the eula for all authorities. We need to
be ready to turn it on when we need to.

This Commit:
----

- Add the AcceptEula screen to the ActivationStack

### DIFF
--- a/src/navigation/ActivationStack.tsx
+++ b/src/navigation/ActivationStack.tsx
@@ -19,6 +19,7 @@ import ActivationSummary from "../Activation/ActivationSummary"
 
 import { Icons } from "../assets"
 import { Spacing, Colors, Typography } from "../styles"
+import AcceptEula from "../Activation/AcceptEula"
 
 type ActivationStackParams = {
   [key in ActivationScreen]: undefined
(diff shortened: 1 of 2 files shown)


src/navigation/ActivationStack.tsx
```diff
@@ -51,14 +57,16 @@ const ActivationStack: FunctionComponent = () => {
     component: NotificationPermissions,
   }
 
+  const defaultSteps = [acceptEulaStep, activateProximityTracing]
```
devinjameson, line 60: 🏅

johnschoeman (commented):
LGTM,
Ive cut a card that extends this to basically having a config var to toggle this, as a heads up.  We might want to do this instead of having a draft pr ready to go.

aledustet:
> LGTM,
> Ive cut a card that extends this to basically having a config var to toggle this, as a heads up. We might want to do this instead of having a draft pr ready to go.

Sounds good to me, wasn't sure if this granted a config. Closing this

PR #338, opened 2020-09-04
Refactor method of re-rendering Exposure History screen on focus

devinjameson:
Why: prior to this commit, we were using `navigation.addListener` to check if the Exposure History screen was focused and updating state to trigger a re-render.

This commit:
- Uses `useIsFocused()` to trigger a re-render of the Exposure History screen on focus
- Removes the call to `observeExposures()` on the Exposure History Stack because it doesn't seem to be doing anything at the moment, given that we are not rendering a badge on the exposure history tab icon

### DIFF
--- a/src/navigation/ExposureHistoryStack.tsx
+++ b/src/navigation/ExposureHistoryStack.tsx
@@ -1,8 +1,6 @@
-import React, { FunctionComponent, useState, useEffect } from "react"
+import React, { FunctionComponent } from "react"
 import { createStackNavigator } from "@react-navigation/stack"
-import { useNavigation } from "@react-navigation/native"
 
-import { useExposureContext } from "../ExposureContext"
 import ExposureHistoryScreen from "../ExposureHistory/index"
 
 import {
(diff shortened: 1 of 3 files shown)


aledustet (approved):
🚢  :shipit:

src/navigation/ExposureHistoryStack.tsx
```diff
@@ -20,24 +18,6 @@ const SCREEN_OPTIONS = {
 }
 
 const ExposureHistoryStack: FunctionComponent = () => {
-  const navigation = useNavigation()
-  const { observeExposures } = useExposureContext()
```
johnschoeman, line 24: should we also remove this from the context?

johnschoeman (approved):
LGTM

src/navigation/ExposureHistoryStack.tsx
devinjameson: I think there is a lot of cleanup that can be done here to remove unused code, but I'll open another PR for that.

PR #339, opened 2020-09-04
Center header title on more info screen

devinjameson:


src/navigation/MainNavigator.tsx
```diff
@@ -39,6 +39,7 @@ const headerScreenOptions: StackNavigationOptions = {
   },
   headerBackTitleVisible: false,
   headerTintColor: Colors.headerText,
+  headerTitleAlign: "center",
```
aledustet, line 42: Will this make the setting on nested navigators obsolete?

aledustet (approved):
:shipit:

src/navigation/MainNavigator.tsx
devinjameson: I believe it serves as the default for any directly nested screens.

johnschoeman (approved):

PR #341, opened 2020-09-05
Remove unused code from ExposureContext

devinjameson:
## Why

We have a lot of unused code in the `ExposureContext`, likely a result of half-finished features that got dropped when priorities changed.

## This Commit

Removes this unused code.

### DIFF
--- a/src/navigation/MainTabNavigator.tsx
+++ b/src/navigation/MainTabNavigator.tsx
@@ -1,5 +1,4 @@
 import React, { FunctionComponent } from "react"
-import { View, StyleSheet } from "react-native"
 import { createBottomTabNavigator } from "@react-navigation/bottom-tabs"
 import { useTranslation } from "react-i18next"
 import { useSafeAreaInsets } from "react-native-safe-area-context"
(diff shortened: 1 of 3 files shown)


src/navigation/MainTabNavigator.tsx
```diff
@@ -95,7 +77,7 @@ const MainTabNavigator: FunctionComponent = () => {
         size={size}
       />
     )
-    return userHasNewExposure ? applyBadge(tabIcon) : tabIcon
```
aledustet, line 98: I thought this was for another feature we were waiting for? Is it going away?

aledustet (approved):
Seems right to me, just wanted to make sure we have confirmation this is getting nixed

src/navigation/MainTabNavigator.tsx
devinjameson: Perhaps @johnschoeman can comment — I think the development of that feature is sufficiently far out that it doesn’t make sense to have half-developed code lying around.

johnschoeman: Havent seen much of this feature recently, so i would lean towards considering this dead code, and therefore we should remove it.

johnschoeman (approved):

PR #343, opened 2020-09-05
Refactor SystemServicesContext

devinjameson:
**Note: this is not high priority work, so I've been working on it for fun in my free time. 🙂**

Why: I was unsatisfied with the organization of the `SystemServicesContext` and wanted to try my hand at cleaning it up and making it more readable.

I'm curious to hear folks' thoughts on this pattern.

This commit:
- Breaks the logic within the `SystemServicesContext` into `useIsBluetoothOn`, `useIsLocationOn` and `useIsLocationNeeded` hooks
- Creates a `useOnAppStateChange` hook to abstract the logic of creating and removing listeners


src/SystemServicesContext/useIsLocationOn.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useState, useCallback } from "react"
+import { Platform } from "react-native"
+
+import useOnAppStateChange from "./useOnAppStateChange"
+import { isLocationEnabled } from "../gaen/nativeModule"
+
+const useIsLocationOn = (): boolean => {
```
aledustet, line 7: Could this be a bit more specific, like determine the location for android or something like that? Would that be relevant or make the structure easier to understand? Naming with much detail is deceiving some times but maybe it makes sense here? Would that contribute to the reason why we split it or would help the split?

src/SystemServicesContext/useIsLocationNeeded.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useState, useCallback } from "react"
+import { Platform } from "react-native"
+
+import useOnAppStateChange from "./useOnAppStateChange"
+import { doesDeviceSupportLocationlessScanning } from "../gaen/nativeModule"
+
+const useIsLocationNeeded = (): boolean => {
```
aledustet, line 7: would this hook be used outside the context of the `useIsLocationOn` hook? Could this become a two-way enum value like the exposure notifications for iOS?

aledustet (approved):
Seems like a great idea to me, just opening up the conversation but I would merge it since you have been dealing with the more down to earth use of it I feel like you have the context

src/SystemServicesContext/useIsLocationOn.ts
devinjameson: That's an interesting point. In this case, I think it makes sense to leave the hook name platform agnostic because the platform check is handled within the hook as opposed to being handled outside in the context. What do you think of that?

src/SystemServicesContext/useIsLocationNeeded.ts
devinjameson: Oo I like that, what do you think of this @aledustet? https://github.com/Path-Check/gaen-mobile/pull/343/commits/6f7da2555cfa4afc39dc66276320b5e7f9312493

src/SystemServicesContext/index.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { createContext, FunctionComponent, useContext } from "react"
+
+import useIsBluetoothOn from "./useIsBluetoothOn"
+import useIsLocationOn from "./useIsLocationOn"
+import useIsLocationRequired from "./useIsLocationRequired"
+
+export enum LocationPermissions {
```
johnschoeman, line 7: This is totally cosmetic, but i find union types easier to read:

```suggestion
export type LocationPermissions = "NotRequired" | "RequiredOff" | "RequiredOn"
```

src/SystemServicesContext/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  isBluetoothOn: false,
+  locationPermissions: LocationPermissions.REQUIRED_OFF,
+}
+
+export const SystemServicesContext = createContext<SystemServicesState>(
+  initialState,
+)
+
+const SystemServicesProvider: FunctionComponent = ({ children }) => {
+  const isBluetoothOn = useIsBluetoothOn()
```
johnschoeman, line 28: nice and clean! I like how we have a clean separation between the Device State and the AppUI State and a function to convert from one to the other.

src/SystemServicesContext/useIsLocationOn.ts
```diff
@@ -0,0 +1,27 @@
...
+  const [isLocationOn, setIsLocationOn] = useState(false)
+
+  const determineIsLocationOn = useCallback(async () => {
+    if (Platform.OS === "android") {
+      isLocationEnabled().then((result) => setIsLocationOn(result))
+    } else {
+      setIsLocationOn(true)
```
johnschoeman, line 14: this `if the platform is ios, then set the locationOn to be true` jumps out as odd to me as we dont actually know the location permission status for ios (and we know that we _cant_ be asking for location permission on ios anyways).  Further, we know that we dont need to check if location is on if we know that the location status is not required, so these are pretty tightly linked. what do you think about combining these hooks into one and having them return a tuple or object?

let me know if you would like to discuss irl!

johnschoeman (approved):
Nice refactor!

src/SystemServicesContext/index.tsx
devinjameson: Yeah agreed!

src/SystemServicesContext/index.tsx
devinjameson: Yeah actually I agree, updated.

PR #344, opened 2020-09-05
Lokalise Updates from Sept 05- Hmong and Somali

alpita-masurkar:
#### Description:
Latest Lokalise updates. Adding two new languages Somali and Hmong (we have a fair bit of translation now but there's translation missing for several keys. Will push updates as soon as I have them).
Also can you confirm once again for me if hmn is the right code for Hmong? I think it is based on what I read, but I would be happy if someone (@aledustet maybe?) can confirm it too.
Thanks.
Alpita


aledustet (approved):
:shipit:

PR #345, opened 2020-09-06
Set up scaffolding for new tabs and settings structure

devinjameson:
## Why
We want the app to have a layout with three tabs: Home, Exposure History, and Connect.

## This Commit
- Updates the tab bar to have Home, Exposure History, and Connect tabs
- Scaffolds out the ConnectStack
- Adds a settings button to the top left of the Home screen that links to the Settings
menu, previously the More menu


src/navigation/MainTabNavigator.tsx
```diff
@@ -165,32 +132,12 @@ const MainTabNavigator: FunctionComponent = () => {
...
-        />
-      )}
       <Tab.Screen
-        name={Stacks.More}
-        component={MoreStack}
+        name={Stacks.Connect}
+        component={ConnectStack}
         options={{
-          tabBarLabel: t("navigation.more"),
-          tabBarIcon: MoreIcon,
```
aledustet, line 193: Are we ready to merge this? Wouldn't it leave the app on a non-functional state? Should we make the additions in a way that it could be there and development wise we could put it on and keep working or is this something that will wait until all elements are present to be released?

devinjameson: I think we would merge it into the feature branch and then once everything is functional merge to develop.

aledustet: That might prove tricky on a long-running branch, it will take discipline to keep this branch as the main one and not using rebase on this one regularly if we have other branches spawn from this one.

src/ReportIssue/ReportIssueForm.tsx
```diff
@@ -16,7 +16,7 @@ import { SvgXml } from "react-native-svg"
 
 import { useVersionInfo } from "../hooks/useApplicationInfo"
 import { GlobalText, Button } from "../components"
-import { reportAnIssue, ReportIssueError } from "../More/zendeskAPI"
+import { reportAnIssue, ReportIssueError } from "../Settings/zendeskAPI"
```
aledustet, line 19: Should this api file be moved to this folder? Is it used anywhere else?

aledustet (approved):
Seems ready to go

src/ReportIssue/ReportIssueForm.tsx
devinjameson: Basically just renaming the More folder to Settings so I think this is ok for now!

PR #346, opened 2020-09-07
Refactor exposure detection

emilianojuan:
#### Description:

Exposure detection method was long and entangled, it performs many actions and had a lot of nested callbacks, by using promises we make that async code readable as if it were sync. We can keep refactoring the finish method but I think this is a good start


ios/BT/API/Result.swift
```diff
@@ -31,17 +31,15 @@ public enum ExposureError: LocalizedError {
   case dailyFileProcessingLimitExceeded
   case cancelled
 
+  /// TODO localize this
```
mattThousand, line 34: Is this comment still applicable? It looks like you've localized the strings below 😃

ios/BT/ExposureManager.swift
```diff
@@ -239,153 +242,58 @@ final class ExposureManager: NSObject {
...
-                    processedFileCount: processedFileCount,
-                    lastProcessedUrlPath: lastProcessedUrlPath,
-                    progress: progress,
-                    completionHandler: completionHandler)
-        return
+    Promise<[Exposure]>(on: .global()) { () -> [Exposure] in
+      if self.isDetectingExposures {
+        // Disallow concurrent exposure detection,
+        // because if allowed we might try to detect the same diagnosis keys more than once
+        throw ExposureError.default("Detection Already in Progress")
```
mattThousand, line 257: general question, will these errors be user-facing? Ideally this would fail silently and the redundant detection cycle would be simply be aborted 🤔

ios/BT/ExposureManager.swift
```diff
@@ -536,10 +440,103 @@ private extension ExposureManager {
     }
   }
 
-  func cleanup() {
-    // Delete downloaded files from file system
-    localUncompressedURLs.cleanup()
-    localUncompressedURLs = []
-    downloadedPackages = []
+  // MARK: == Exposure Detection Private Promisses ==
```
mattThousand, line 443: ~Promisses~ Promises

mattThousand: what about naming this function `fetchIndexFile` (because it is fetching the index file from the server)?

ios/BT/ExposureManager.swift
```diff
@@ -536,10 +440,103 @@ private extension ExposureManager {
...
+
+  func fetchFilesIndexes() -> Promise<String> {
+    return Promise<String> { fullfill, reject in
+      self.apiClient.requestString(IndexFileRequest.get,
+                              requestType: .downloadKeys) { result in
+        switch result {
+        case .success(let index):
```
mattThousand, line 450: `index` might be an odd name for this. What about something such as `keyArchiveFilePathsString`? That is also quite a long variable name, but I'd like to communicate the meaning of the variable here somehow 🤔

ios/BT/ExposureManager.swift
```diff
@@ -536,10 +440,103 @@ private extension ExposureManager {
...
+          fullfill(index)
+        case .failure(let error):
+          reject(error)
+        }
+      }
+    }
+  }
+
+  func downloadPackedKeys(targetUrls: [String]) -> Promise<[DownloadedPackage]> {
```
mattThousand, line 459: ~downloadPackedKeys~ `downloadKeyArchives`

ios/BT/ExposureManager.swift
```diff
@@ -536,10 +440,103 @@ private extension ExposureManager {
...
+          dispatchGroup.leave()
+        }
+      }
+      dispatchGroup.notify(queue: .main) {
+        fullfill(downloadedPackages)
+      }
+    }
+  }
+
+  func unpackPackages(packages: [DownloadedPackage]) -> Promise<[URL]> {
```
mattThousand, line 481: ~unpackPackages~ `unpackedKeyArchives`

ios/BT/ExposureManager.swift
```diff
@@ -536,10 +440,103 @@ private extension ExposureManager {
...
+        try packages.unpack({ (urls) in
+          fullfill(urls)
+        })
+      } catch(let error) {
+        reject(error)
+      }
+    }
+  }
+
+  func getExposureConfiguration() -> Promise<ExposureConfiguration> {
```
mattThousand, line 493: 👍

ios/BT/ExposureManager.swift
```diff
@@ -239,153 +242,58 @@ final class ExposureManager: NSObject {
...
-  /// Local urls of the bin/sig files from each archive
-  private var localUncompressedURLs = [URL]()
 
   @discardableResult func detectExposures(completionHandler: @escaping ((ExposureResult) -> Void)) -> Progress {
     
     let progress = Progress()
-    
     var lastProcessedUrlPath: String = .default
     var processedFileCount: Int = 0
+    var unpackedKeysURLs: [URL] = []
```
mattThousand, line 251: ~unpackedKeysURLs~ `unpackedArchiveURLs`

mattThousand (changes requested):
Thank you for doing this Emi!  In general looks great. Just a few small changes and a question

ios/BT/API/Result.swift
emilianojuan: I dont know how localization and react native works, I added this comment so we add the proper translations

ios/BT/ExposureManager.swift
emilianojuan: on my understanding we are only showing error notifications in debug mode
check postExposureDetectionErrorNotification

ios/BT/ExposureManager.swift
emilianojuan: yes, I'll change it

ios/BT/ExposureManager.swift
emilianojuan: you are right, I'll change it

mattThousand (approved):

PR #352, opened 2020-09-08
Replace welcome image with specific image on build

aledustet:
Why:
----
We need to offer the option to override the welcome screen graphic for branding purposes per authority. The override should be optional and we should be able to build the app without requiring it.
This should only be updated if the file is present.

This Commit:
----
- Download welcome image from the resources repo if the file exists
- Rename welcome screen image with a more generic name to be replaced

Reviewers:
----
There is a PR pending on resources for this


johnschoeman (approved):
LGTM!

PR #354, opened 2020-09-09
Show error when there is no internet connection

juancruzgs:
#### Description:

When the user tries to submit a verification code and there is no internet connection we are showing the error "Try a different code".
This PR shows an error saying "There is no internet connection" when that happens.

#### Linked issues:

https://pathcheck.atlassian.net/browse/GAEN-260

#### Screenshots:

device-2020-09-08-220158

#### How to test:

Disable WiFi and mobile data and try to submit a verification code

### DIFF
--- a/src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
+++ b/src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
@@ -131,8 +131,8 @@ const CodeInputForm: FunctionComponent = () => {
       case "VerificationCodeUsed": {
         return t("export.error.verification_code_used")
       }
-      case "InvalidVerificationUrl": {
-        return "Invalid Verification Url"
+      case "NetworkConnection": {
+        return t("export.error.network_connection_error")
       }
       default: {
         return t("export.error.unknown_code_verification_error")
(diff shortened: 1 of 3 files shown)


mattThousand (approved):

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -131,8 +131,8 @@ const CodeInputForm: FunctionComponent = () => {
       case "VerificationCodeUsed": {
         return t("export.error.verification_code_used")
       }
-      case "InvalidVerificationUrl": {
-        return "Invalid Verification Url"
+      case "NetworkConnection": {
```
aledustet, line 134: Good call removing this error, we haven't seen it in a while.

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -131,8 +131,8 @@ const CodeInputForm: FunctionComponent = () => {
...
+        return t("export.error.network_connection_error")
```
aledustet, line 135: Good call removing this error, we haven't seen it in a while.

aledustet (approved):
Looks ready to me, I was wondering if we should create some future work for shielding this flow from starting if there's no internet connection, I know we have a hook that does this. Probably worth adding the ticket to check if we have internet even before we start this flow. Food for thought 😄

PR #357, opened 2020-09-09
Add call back form shown per configuration

aledustet:
Why:
----
There is a need for one of the authorities to submit a call back flow from inside the application. The form should be available when requested on build time and it should submit, name and phone number from a customer to that endpoint using provided credentials from the authority. The form will have a final spot inside the exposure details flow, but for starters we will have it on the settings screen.

Reviewers:
----
The form is currently complete but the data is not being pulled from the right spot(namely the date). There are still some flow missing, like from where to show the form and limits on who/how can it be used. This body of work is the form to gather the data and the communication with the backend service to submit such data.

This Commit:
----
- Add the callback form screen
- Add the helper to submit the form data to the provided endpoint via environment
- Use configuration to hide/show the form
- Move displayCallbackForm to configuration context

Co-authored-by: Matt Buckley <matt.buckley@rocketinsights.com>

### DIFF
--- a/src/More/CallbackForm.tsx
+++ b/src/More/CallbackForm.tsx
@@ -1,0 +1,237 @@
+import React, { useState, FunctionComponent } from "react"
+import { useTranslation } from "react-i18next"
+import {
+  ActivityIndicator,
+  Alert,
+  KeyboardAvoidingView,
+  StyleSheet,
+  Platform,
+  TextInput,
+  View,
+  ScrollView,
+  Keyboard,
+} from "react-native"
+
+import { useStatusBarEffect } from "../navigation"
+import { useConfigurationContext } from "../ConfigurationContext"
+import { Button, GlobalText } from "../components"
+import * as API from "./callbackAPI"
+
+import { Spacing, Layout, Forms, Colors, Outlines, Typography } from "../styles"
+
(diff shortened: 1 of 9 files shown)


src/navigation/SettingsStack.tsx
```diff
@@ -46,6 +47,10 @@ const SettingsStack: FunctionComponent = () => {
         }}
```
aledustet, line 47: The header shown detail might be useful for the rest of the views under this stack but we will most likely have to set the options with the hook once the screen loads to update the title and maybe the back action?

src/More/CallbackForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+          onPress={handleOnPressSubmit}
+          label={t("common.submit")}
+          loading={isLoading}
+        />
+      </KeyboardAvoidingView>
+    </ScrollView>
+  )
+}
+
+const LoadingIndicator = () => {
```
johnschoeman, line 168: @devinjameson noticing that we have 3 or so of these now, what do you think about introducing a component for them to to be shared. seems _pretty_ self contained to me and not likely to want to diverge based on call site?

johnschoeman (approved):
LGTM

PR #359, opened 2020-09-09
Update content of Settings screen and make some navigational refactors

devinjameson:
## Why
The content on the Settings and Home screen has been updated.

## This Commit
- Adds a new ModalStack to house Modal screens that show up over the bottom tab bar
- Removes the language select button from the home screen
- Moves the LanguageSelection screen to the Modal folder and puts it in the ModalStack
- Removes the "card style" animations from the app as they we were running into strange bugs while using them. Using standard modals styles solves this issue.
- Removes the SafeAreaView from the OnboardingScreen and uses manually set insets instead
- Moves the ReportIssue screen from its own stack to the Settings stack
- Moves the ProtectPrivacy screen and AffectedUserFlow both to the ModalStack
- Updates the SettingsStack to include the correct screens

gif
gif2


aledustet (approved):
Not 100% sure I understand the nuances of the values for the styles but it looks great 😅 
My one grain of salt might be on the naming. Having a `modal` stack sounds a bit odd since the concept of modals is already a thing. I can see some friction with this in the future if we start using modals more often in the app. Or perhaps this is a smell that these should be modals 🤷 Could a name like `FloatingStack` or something in that verse work?

devinjameson:
> My one grain of salt might be on the naming. Having a `modal` stack sounds a bit odd since the concept of modals is already a thing. I can see some friction with this in the future if we start using modals more often in the app. Or perhaps this is a smell that these should be modals 🤷 Could a name like `FloatingStack` or something in that verse work?

@aledustet My thinking here is that this Stack would contain any screens that show up as modals, i.e. screens that show over the tab bar and slide up from the bottom. Does that make sense? Otherwise these screens would just be floating around in the `MainNavigator` which feels odd.

src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
```diff
@@ -233,67 +239,69 @@ const PublishConsentForm: FunctionComponent<PublishConsentFormProps> = ({
...
-    borderTopColor: Colors.neutral25,
-    borderTopWidth: Outlines.hairline,
-  },
-  bottomButtonText: {
-    ...Typography.header5,
-    color: Colors.primary100,
-    marginRight: Spacing.xSmall,
-  },
-})
+const createStyle = (insets: EdgeInsets) => {
```
johnschoeman, line 242: would it make sense to have this function be more specific to the particular component that is dependent on the parameter? something like
```typescript
const createButtonContainerStyle = (insets: EdgeInsets) => {
....

const style = StyleSheet.create({
...
```

src/Onboarding/OnboardingScreen.tsx
```diff
@@ -62,92 +57,174 @@ const OnboardingScreen: FunctionComponent<OnboardingScreenProps> = ({
...
+    },
+    bottomButtonText: {
+      ...Typography.header5,
+      color: Colors.primary100,
+      marginRight: Spacing.xSmall,
+    },
+  })
+}
+
+export default OnboardingScreen
```
johnschoeman, line 226: Nbd on this one, but its convention to have the exports at the bottom of the file.

src/Settings/index.tsx
```diff
@@ -69,16 +70,21 @@ const MenuScreen: FunctionComponent = () => {
             lastItem
           />
         )}
+        <SettingsListItem
```
johnschoeman, line 73: Should this be conditionally rendered?
my understanding is that we dont want this on most of the builds.

src/navigation/ModalStack.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { FunctionComponent } from "react"
+import { createStackNavigator } from "@react-navigation/stack"
+
+import { Stacks, ModalScreens } from "./index"
+import LanguageSelection from "../Modal/LanguageSelection"
```
johnschoeman, line 5: What do you think about not having a `Modal` folder?  (and if you decide to keep it, what do you think about `modal/` as convention is CamelCase for specific components, and `Modal` here is not a specific component.)

johnschoeman (approved):
Code looks great! a few minor thoughts.

The one thing that should be address though is conditionally rendering the report an issue form as we dont want that in every build.

src/AffectedUserFlow/PublishConsent/PublishConsentForm.tsx
devinjameson: What do you think the advantage of that would be? I'm leaning toward keeping all the styles in one place for simplicity, and because naming could get weird if we have more than one component that uses the parameter.

src/navigation/ModalStack.tsx
devinjameson: Ah I see, using `modal` makes sense to me if that's cool.

src/Settings/index.tsx
```diff
@@ -0,0 +1,178 @@
+import React, { FunctionComponent } from "react"
```
devinjameson, line 1: @johnschoeman Can you give this a once over before merge pretty please?

src/Settings/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const baseListItems: SettingsListItem[] = [About, Legal]
+
+  const settingsListItems: (SettingsListItem | undefined)[] = [
+    ...baseListItems,
+    displayCallbackForm ? CallbackForm : undefined,
+    displayReportAnIssue ? ReportAnIssue : undefined,
+  ]
+
+  const settingsListItemsToDisplay: SettingsListItem[] = settingsListItems.filter(
```
johnschoeman, line 71: I dig the functional style! though i think you could get away without the filter if you hoist the conditionals up and allow for some mutation. e.g.
```typescript
const listItems = [About, Legal]
if (displayCallbackForm) {
  listItems.push(CallbackForm)
}
if (displayReportAnIssue) {
  listItems.push(ReportAnIssue)
}
```
which might be a bit easier to follow.

johnschoeman (approved):
Continues to look good! some nerdy programing things to think, if you are interested.

src/Settings/index.tsx
devinjameson: Ahhh this is much easier to follow, thanks!

devinjameson: Love the nerdy programming thingz!

PR #360, opened 2020-09-10
Add scrollview around welcome screen

aledustet:
Why:
----

On smaller phones the call to action on the welcome screen can not be tapped because the image and text push it out the screen. We need this button to be reachable by users when it is not reachable.

How:
----

Adding a scroll view container where the container can grow to its full height will allow users on smaller screens to scroll down as they need to.

This Commit:
----

- Add scroll view container to welcome screen

Reviewers:
----

Not sure this is a long term solution to some text size issues but it allows users to move past this step.

Large GIF (396x656) Screen Shot 2020-09-10 at 10 23 18 AM

### DIFF
--- a/src/Onboarding/Welcome.tsx
+++ b/src/Onboarding/Welcome.tsx
@@ -1,5 +1,11 @@
 import React, { FunctionComponent } from "react"
-import { Image, StyleSheet, View, TouchableOpacity } from "react-native"
+import {
+  Image,
+  StyleSheet,
+  View,
+  TouchableOpacity,
+  ScrollView,
+} from "react-native"
 import { useTranslation } from "react-i18next"
 import { useNavigation } from "@react-navigation/native"
 import LinearGradient from "react-native-linear-gradient"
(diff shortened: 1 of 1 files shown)


johnschoeman (approved):

src/Onboarding/Welcome.tsx
```diff
@@ -44,50 +50,60 @@ const Welcome: FunctionComponent = () => {
   return (
     <>
       <StatusBar backgroundColor={Colors.primaryLightBackground} />
-      <GradientBackground
-        gradient={Colors.gradientPrimary10}
-        angleCenterY={0.75}
+      <ScrollView
+        alwaysBounceVertical={false}
+        contentContainerStyle={style.scrollContentContainer}
```
devinjameson, line 55: We tend to call this `style.contentContainer` elsewhere in the app but totally a matter of preference.

devinjameson (approved):
:shipit:

PR #361, opened 2020-09-10
Update translations and remove unused keys

aledustet:
Why:
----

We need the latest translations from lokalise. Also, we need to periodically make sure that unused keys are trimmed from the lokalise project since the push script does not "force" the unused keys to be deleted properly on CI.

This Commit:
----

- Cleanup unused translations, running `yarn i18n:extract`
- Move translation keys usage to be with the `t` method so the extract script picks up those translations
- Update translations after pulling from lokalise
- Make sure all available languages are shown on the language selection

Screen Shot 2020-09-10 at 12 24 58 PM


johnschoeman (approved):

devinjameson (approved):
Not all heroes wear capes 🦸‍♂️

PR #366, opened 2020-09-10
Add button to restart onboarding and update settings screen design

devinjameson:
Why: as a user, I would like to be able to click a button on the Settings screen to
be brought back through the onboarding flow, but not the activation flow.

This commit:
- Adds a button to the settings screen that restarts Onboarding and brings the
user back to the Settings screen when they finish onboarding
- Refactors the `OnboardingStack`
- Adds icons to the settings list buttons

gif2


src/OnboardingContext.tsx
```diff
@@ -29,23 +32,37 @@ export const OnboardingProvider: FunctionComponent<OnboardingProviderProps> = ({
...
   const completeOnboarding = () => {
     StorageUtils.setIsOnboardingComplete()
-    setOnboardingIsComplete(true)
+    setIsOnboardingComplete(true)
   }
 
   const resetOnboarding = () => {
     StorageUtils.removeIsOnboardingComplete()
-    setOnboardingIsComplete(false)
+    setDestinationAfterComplete(Stacks.Activation)
```
aledustet: If this method will be called from the contact menu should this have defaulted to these activation screens? Should this argument be passed as an argument if we change this method somewhere else? We could leave the argument as the activation stack by default. Or is this just for debug purposes that we must have the complete flow there?

App.tsx
```diff
@@ -1,35 +1,37 @@
...
 
 Logger.start()
 
 const App: FunctionComponent = () => {
   const [isLoading, setIsLoading] = useState(true)
-  const [onboardingIsComplete, setOnboardingIsComplete] = useState(false)
+  const [isOnboardingComplete, setIsOnboardingComplete] = useState(true)
+
+  const determineIsOnboardingComplete = async (): Promise<boolean> => {
+    return await StorageUtils.getIsOnboardingComplete()
```
aledustet: Curious to learn why we moved this back into the component? I thought the intent was not to reach for helpers directly from the components if these get an accessor from a context. Don't think one way or the other is best but curious about why the reversal.

src/Settings/index.tsx
```diff
@@ -92,42 +115,27 @@ const Settings: FunctionComponent = () => {
   return (
     <ScrollView style={style.container}>
       <View style={style.section}>
-        <TouchableOpacity
+        <ListItem
```
aledustet, line 125: Should this be moved also to the external declarations like the rest of the items on the list?

aledustet (approved):
🚢 :shipit: 

One or two questions about some placements but this seems ready to ⛵

src/OnboardingContext.tsx
```diff
@@ -16,9 +17,11 @@ export const OnboardingContext = createContext<
 >(undefined)
 
 export interface OnboardingContextState {
-  onboardingIsComplete: boolean
+  isOnboardingComplete: boolean
   completeOnboarding: () => void
   resetOnboarding: () => void
+  destinationAfterComplete: Stack
```
johnschoeman: Could this be a prop on the Stack Component itself? Seems like if we are going to be making a distinction in the flows in which the onboarding screens can appear by mounting two separate instances of the screen we dont need to add to the state in the context, but parameterize the component?

I suppose im a little concerned about adding confusion to the notion of `Onboarding`. Is it the act of enabling all the features we need on first visit to the app? or is it the act of moving through a particular set of screens?  I would say its the former not the later and this conflates those two flows in a way that would be hard to parse as a new reader of the codebase.

I might even consider changing what we call the OnboardingStack now to something like `HowTheAppWorksStack` and consider onboarding to be the flow of `Welcome -> How The App Works -> Activation -> the Main App`

Something like this would also allow us to remove the first `Welcome` screen from the currently named `OnboardingReview` flow which seems a bit out of place.

johnschoeman (approved):
Code looks great! I think we may be overusing the context here and a few thoughts on how we logically structure the flows.

src/OnboardingContext.tsx
devinjameson: My understanding is that this function will be called from the debug menu and that we want to reset the entire onboarding state to be like it was on first download. So we want to set `isOnboardingComplete` to false and ensure that `destinationAfterComplete` is set to `Stacks.Activation`, because if they have tapped `How the app works` on the Settings screen, `destinationAfterComplete` would be set to `Stacks.Settings`. Does that make sense? I may be misunderstanding you.

App.tsx
devinjameson: Yeah makes sense to move this function back to the `OnboardingContext` and just import it!

src/Settings/index.tsx
devinjameson: Good call, we'll also move the `ENDebugMenu` there so they're all in one place.

src/OnboardingContext.tsx
devinjameson: Makes sense @johnschoeman, went down the path of parameterizing the `OnboardingStack`. I think it's more clear. Can you take a look? In the process, we ended up refactoring the `OnboardingStack` to map through onboarding screen data instead of rendering various screens e.g. `Introduction` that in turn render an `OnboardingScreen`. This reduced the amount of strange and cumbersome prop-drilling.

devinjameson: Thoughts on renaming in a future PR though? I agree but it seems like a lot for this PR.

src/navigation/OnboardingStack.tsx
```diff
@@ -23,32 +23,59 @@ const onboardingScreenOptions: StackNavigationOptions = {
...
-      <Stack.Screen
-        name={OnboardingScreens.GetNotified}
-        component={GetNotified}
-      />
-      <Stack.Screen
-        name={OnboardingScreens.ValueProposition}
-        component={ValueProposition}
-      />
+      <Stack.Screen name={Screens.Welcome} component={Welcome} />
+      {onboardingScreens.map(
```
johnschoeman:
```suggestion
      const toStackScreen = (data) => Component
      {onboardingScreenData.map(toStackScreen)
```

I like this refactor with the hook and fewer component files a lot! better separation of data from markup and its nice to have the data all in one spot

johnschoeman (approved):
Nice! well done.

src/OnboardingContext.tsx
johnschoeman: makes sense to me!

src/OnboardingContext.tsx
aledustet: Roger roger

src/navigation/OnboardingStack.tsx
devinjameson: Heck yaaaaaa

devinjameson: I like this `toStackScreen` function, very Elm-y.

PR #368, opened 2020-09-11
Resolve promises after the user denies/accepts the permissions dialog

juancruzgs:
#### Description:

There are some GAEN API methods that require permission from the user.
We were resolving JS promises before the user accepted/denied that permissions Dialog, and that wasn't reflecting the correct state.

This PR waits until the dialog is accepted/denied to return the result to the JS layer.
It mostly impacts some options in the debug menu and shouldn't have any impact in the main flow.

This PR also:
- Moves some code from `MainActivity` to `ExposureNotificationsClientWrapper`
-  Uses the `rollingPeriod` returned by the API instead of a default value

### DIFF
--- a/android/app/src/main/java/org/pathcheck/covidsafepaths/MainActivity.java
+++ b/android/app/src/main/java/org/pathcheck/covidsafepaths/MainActivity.java
@@ -1,46 +1,17 @@
 package org.pathcheck.covidsafepaths;
 
-import android.app.Activity;
 import android.content.Intent;
-import android.content.IntentSender;
 import android.os.Bundle;
 import androidx.annotation.Nullable;
 import com.facebook.react.ReactActivity;
 import com.facebook.react.ReactActivityDelegate;
 import com.facebook.react.ReactRootView;
-import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactContext;
-import com.google.android.gms.common.api.ApiException;
-import com.google.android.gms.nearby.exposurenotification.ExposureNotificationStatusCodes;
-import com.google.android.gms.nearby.exposurenotification.TemporaryExposureKey;
-import com.google.common.collect.ImmutableList;
-import com.google.common.io.BaseEncoding;
-import com.google.common.util.concurrent.FluentFuture;
-import com.google.common.util.concurrent.FutureCallback;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.ListenableFuture;
 import com.swmansion.gesturehandler.react.RNGestureHandlerEnabledRootView;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.TimeUnit;
 import org.devio.rn.splashscreen.SplashScreen;
-import org.jetbrains.annotations.NotNull;
 import org.pathcheck.covidsafepaths.exposurenotifications.ExposureNotificationClientWrapper;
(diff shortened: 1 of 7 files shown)


android/app/src/main/java/org/pathcheck/covidsafepaths/MainActivity.java
```diff
@@ -91,126 +62,11 @@ private void checkIfExposureNotificationsEnabled() {
...
-          public void onSuccess(ImmutableList<DiagnosisKey> shared) {
-            final List<RNExposureKey> exposureKeys = new ArrayList<>();
-
-            for (DiagnosisKey k : shared) {
-              final String key = BASE64.encode(k.getKeyBytes());
-              final int rollingStartNumber = k.getIntervalNumber();
-
-              exposureKeys.add(new RNExposureKey(
-                  key,
-                  DEFAULT_PERIOD,
```
juancruzgs, line 157: We were sending to JS this default rolling period (144) for all the keys returned by the API.
I'm now using the `rollingPeriod` that is present in the `TemporaryExposureKey` object returned by the API.

@aledustet @sherifkozman  Is this the right thing to do, or should I keep using the default value?

aledustet: Here it says it's 144 https://blog.google/documents/69/Exposure_Notification_-_Cryptography_Specification_v1.2.1.pdf not sure where else we could look for that value

android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/ExposureNotificationClientWrapper.java
```diff
@@ -57,26 +61,72 @@ public static ExposureNotificationClientWrapper get(Context context) {
...
-            return;
+          completer.setException(exception);
+        })
+    );
+  }
+
+  /**
+   * Gets recent (initially 14 days) Temporary Exposure Keys from Google Play Services.
+   */
+  public ListenableFuture<List<TemporaryExposureKey>> requestPermissionToGetExposureKeys(ReactContext context) {
```
aledustet, line 99: Just curious if this one could follow the same pattern that the `fetchDiagnosisKeys` having an `onSuccess` `onFailure` implementations.

aledustet (approved):
Don't see anything that pops up for me here, not 100% sure about the `rollingPeriod` but it seems like that 144 is the default across the board where i've been looking(docs from apple and google) :shipit:

android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/ExposureNotificationClientWrapper.java
juancruzgs: We have a mix of implementations because the GAEN API uses `Tasks` and the app uses `Futures`.
In a future refactor I'll change `ExposureNotificationClientWrapper` to only expose `Futures` (since Tasks can only be created by the API).

That's why in some places we have `addOnSuccessListener`/`addOnFailureListener` and some others have `onSuccess`/`onFailure`

juancruzgs:
`144` is the default `rollingPeriod`, we took it from Google's sample code.
But the keys can use different values, for example:
image

I confirmed with Matt that iOS is sending the value that is returned for each key instead of a hardcoded one

PR #369, opened 2020-09-11
Rename Onboarding to HowItWorks and make general navigational refactors

devinjameson:
Why:
- Our concept of onboarding was unclear. We were calling the entire Welcome > How it Works > Activation process "Onboarding", but also calling the "How it Works" part itself "Onboarding"
- Some of our navigation types had become outdated and unclear

This commit:
- Renames the onboarding stuff that has to do with how the app works "How it Works"
- Removes the `Screens` type which was a catch all for all screens in the app
- Moves the `ExposureDetail` and `MoreInfo` screens into the `ExposureHistoryStack`
- Only shows the "Next Steps" button on the `ExposureDetail` screen if we have a link to
send the user to, because the `SelfAssessment` flow should not be displayed


src/Welcome.spec.tsx
```diff
@@ -20,13 +20,13 @@ jest.mock("react-native-webview", () => {
     },
   }
 })
-jest.mock("../locales/eula/en.html", () => {
```
aledustet, line 23: Do we still need these? Maybe we can run a quick cleanup after since we are only using the links on that terms of service screen right?

src/navigation/MainNavigator.tsx
```diff
@@ -65,31 +47,20 @@ const MainNavigator: FunctionComponent = () => {
...
-                ...headerScreenOptions,
-                title: t("navigation.exposure"),
-              }}
-            />
           </>
         ) : (
           <>
             <Stack.Screen
-              name={Stacks.Onboarding}
+              name={OtherScreens.Welcome}
```
aledustet: Is there any other name we could add here to encapsulate this view? I'm worried about having a catch-all too soon and overuse it 😬

src/styles/headers.ts
```diff
@@ -13,3 +14,15 @@ export const headerTitleStyle: TextStyle = {
   letterSpacing: Typography.mediumLetterSpacing,
   textTransform: "uppercase",
 }
+
+export const headerScreenOptions: StackNavigationOptions = {
```
aledustet, line 18: 👏

src/HowItWorks/HowItWorksScreen.tsx
```diff
@@ -226,11 +227,11 @@ interface PositionDotsProps {
   screenNumber: number
 }
 
-const NUMBER_OF_ONBOARDING_SCREENS = 5
-
 const PositionDots: FunctionComponent<PositionDotsProps> = ({
   screenNumber,
 }) => {
+  const howItWorksData = useHowItWorksData()
```
johnschoeman: its a bit odd to have this hook in the helper component, what do you think about parameterizing this?
```suggestion
  const PositionDots: FunctionComponent<PositionDotsProps> = ({
    highlightedDotIdx,
    totalDotCount
}) => {
```

src/HowItWorks/HowItWorksScreen.tsx
```diff
@@ -272,6 +273,6 @@ const dotsStyle = StyleSheet.create({
   },
 })
 
-const MemoizedOnboardingScreen = React.memo(OnboardingScreen)
+const MemoizedHowItWorksScreen = React.memo(HowItWorksScreen)
```
aledustet, line 280: Could it be that the problem we were having with having to add 2 stacks for the onboarding had to do with not receiving the navigation as a prop here? Am I way off? Curious if accessing the navigation via the hook and doing the render based on the props had something to do with it. If you tried it sounds like a nice learning. 😄

src/navigation/HowItWorksStack.tsx
```diff
@@ -0,0 +1,53 @@
...
+const Stack = createStackNavigator()
+
+interface HowItWorksStackProps {
+  destinationOnSkip: DestinationStack
+}
+
+const HowItWorksStack: FunctionComponent<HowItWorksStackProps> = ({
+  destinationOnSkip,
+}) => {
+  const howItWorksData = useHowItWorksData(destinationOnSkip)
```
johnschoeman: Why are we using a hook here? is it mostly to group the data and get access to the use useTranslationHook?

Typically hooks are intended for sharing stateful logic, which we arent really needing to do here beyond wrapping the useTranslation hook. So it seems like we are adding a bit of complexity for no real gain.  What do you think about inlining the screen data?

aledustet (approved):
Feels like a step in the right direction, 🔥

src/navigation/HowItWorksStack.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    return (
+      <Stack.Screen
+        key={howItWorksScreenDisplayDatum.header}
+        name={howItWorksScreenDisplayDatum.name}
+      >
+        {(props) => (
+          <HowItWorksScreen
+            {...props}
+            howItWorksScreenContent={howItWorksScreenDisplayDatum}
```
johnschoeman, line 112:
```suggestion
            howItWorksScreenContent={howItWorksScreenDisplayDatum}
            totalScreenCount={...}
            destinationOnSkip={destinationOnSkip}
            
```

johnschoeman (approved):
Code looks good! one thought on the hook.

src/HowItWorks/HowItWorksScreen.tsx
devinjameson: Sounds good!

src/HowItWorks/HowItWorksScreen.tsx
devinjameson: That's an interesting thought, but I don't think so. The problem with having two stacks with the same name was a bit different. The onboarding stack that is shown when the app is first downloaded was getting registered as the initial route, and since another stack with that same name was in the main app stack, the name did not get unregistered as the initial route -- even though it wasn't the first screen in the navigator. Does that make sense?

src/navigation/HowItWorksStack.tsx
devinjameson: 👍

src/navigation/HowItWorksStack.tsx
devinjameson: Makes sense! 👍

src/navigation/MainNavigator.tsx
devinjameson: Yeah good idea. I'll just call it `WelcomeScreens`.

src/Welcome.spec.tsx
devinjameson: It appears we don't! Nice catch!

PR #370, opened 2020-09-11
add version 2 of the exposure detection algorithm

emilianojuang:
#### Description:

Implementation of the exposure detection algorithm V2 for iOS. A new method was added so when deprecating iOS13.5 we can remove it without having too many complications


ios/BT/API/Model/Exposure.swift
```diff
@@ -6,20 +6,11 @@ import RealmSwift
 class Exposure: Object, Codable {
   @objc dynamic var id: String = .default
   @objc dynamic var date: Int = 0
-  @objc dynamic var duration: Double = 0.0
```
lmartensson, line 9: Hm, are these not needed anymore?

ios/BT/ExposureManager.swift
```diff
@@ -244,7 +241,15 @@ final class ExposureManager: NSObject {
   private var isDetectingExposures = false
 
   @discardableResult func detectExposures(completionHandler: @escaping ((ExposureResult) -> Void)) -> Progress {
-    
+    if #available(iOS 13.7, *) {
```
lmartensson, line 253: I don't know anything about iOS development, but the docs here says you're supposed to set v2 in info.plist:
https://developer.apple.com/documentation/exposurenotification (under "Specify Exposure Notification API Version
")
Do you need to do this also, or should this just be done differently?
@mattThousand

ios/BT/ExposureManager.swift
```diff
@@ -297,7 +301,72 @@ final class ExposureManager: NSObject {
...
+      lastProcessedUrlPath = targetUrls.last ?? .default
+      processedFileCount = targetUrls.count
+      let downloadedKeyArchives = try await(self.downloadKeyArchives(targetUrls: targetUrls))
+      unpackedArchiveURLs = try await(self.unpackKeyArchives(packages: downloadedKeyArchives))
+      let exposureConfiguraton = try await(self.getExposureConfigurationV2())
+      var exposureSummary = try await(self.callDetectExposures(configuration: exposureConfiguraton.asENExposureConfiguration,
+                                                               diagnosisKeyURLs: unpackedArchiveURLs))
+      exposureSummary = try await(self.callAggregateDetectExposures(configuration: exposureConfiguraton.asENExposureConfiguration))
+      var newExposures: [Exposure] = []
+      if let summary = exposureSummary {
```
lmartensson: Note that a difference here is that the OS will now keep track of older exposures for you.
What happens if, yesterday, the app detected an exposure that happened a few days ago, and now you run this code again? We wouldn't want to notify again for an exposure we already knew about.

lmartensson (commented):
Looks good overall!

A couple of things to pay attention to, so we don't keep notifying on an exposure we already knew about, plus stuff about how to version things correctly that I'm not sure how they're supposed to work but the apple docs mention info.plist.

Thanks!

ios/BT/ExposureManager.swift
emilianojuang: we are setting ENAPIVersion in build time according to the .env file corresponding to each HA, to me this value should be always be set in 2 now, if the user is running <13.5, ENAPIVersion is ignored, and for those running iOS 13.7+ we always want to use V2

lmartensson:
Ok, sounds good!

mattThousand (changes requested):
Looks great overall! It appears that one test case is failing at the moment however: `testDetectExposuresSuccessScoreBellow`

ios/BT/API/Model/Exposure.swift
mattThousand: the UI only makes use of the exposure's `date` property (the use of `duration` etc is an artifact of an outdated design spec for the UI). This is the current UI for the exposure history screen:
IMG_497B415B1FD4-1

ios/BT/ExposureManager.swift
emilianojuang: uhmmm yes, should would filter exposures by date before scoring?

emilianojuang:
> Looks great overall! It appears that one test case is failing at the moment however: `testDetectExposuresSuccessScoreBellow`

yes, Matt, im working on it, and I will add more tests to v2 algorithm

ios/BT/ExposureManager.swift
lmartensson: I'd want to hear what @mattThousand has to say here as well, but here's what I would think:
- store exposures that led to notifications. this should be at most one per day, since we now have daily summaries.
  I think we do this already right, for the calendar? With the code in Exposure.swift above?
- for each newExposures, that happened on a day that we did *not* already notify for, do notify again
- save new exposures that led to notifications
- It looks to me as if we currently only save the date as meta data for the exposure? If so, we're done. If we care about duration as well, or other meta data, we should make sure to potentially update existing exposures as well, in case we added more exposure to a day we already notified for.

mattThousand: If a day should have at most one exposure, it sounds like you will want to store any exposures that result in notifications in realm (or another option would be to store the exposure's date), and each time this codepath is run, filter out resulting exposures for dates that already have exposures in realm. Duration etc is not used, only the exposure's date

mattThousand (approved):

johnschoeman (approved):
Went over together irl, looks great!

PR #371, opened 2020-09-14
Check for Already Stored Exposures on a Given Date

mattThousand:
Per @lmartensson a user should not be notified for exposures on days for which they have already been notified. This PR checks the database on iOS for existing exposures on a given date before notifying the user and storing the exposure.

### DIFF
--- a/ios/BT/ExposureManager.swift
+++ b/ios/BT/ExposureManager.swift
@@ -337,9 +337,10 @@ final class ExposureManager: NSObject {
       var newExposures: [Exposure] = []
       if let summary = exposureSummary {
         summary.daySummaries.forEach { (daySummary) in
-          if daySummary.isAboveScoreThreshold(with: exposureConfiguraton) {
+          if daySummary.isAboveScoreThreshold(with: exposureConfiguraton) &&
+              self.btSecureStorage.canStoreExposure(for: daySummary.date) {
             let exposure = Exposure(id: UUID().uuidString,
-                                    date: daySummary.date.posixRepresentation)
+                                    date: daySummary.date.toMidnight.posixRepresentation)
             newExposures.append(exposure)
           }
         }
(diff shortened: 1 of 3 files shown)


johnschoeman (approved):
LGTM

ios/BT/ExposureManager.swift
```diff
@@ -337,9 +337,10 @@ final class ExposureManager: NSObject {
       var newExposures: [Exposure] = []
       if let summary = exposureSummary {
         summary.daySummaries.forEach { (daySummary) in
-          if daySummary.isAboveScoreThreshold(with: exposureConfiguraton) {
+          if daySummary.isAboveScoreThreshold(with: exposureConfiguraton) &&
+              self.btSecureStorage.canStoreExposure(for: daySummary.date) {
             let exposure = Exposure(id: UUID().uuidString,
-                                    date: daySummary.date.posixRepresentation)
+                                    date: daySummary.date.toMidnight.posixRepresentation)
```
lmartensson, line 343: I don't think we should do toMidnight here? I believe the dates are already normalized to be by day?

mattThousand: then wouldn't the time of day have no effect? The goal is for each exposure to be saved to the database at the same time of day

mattThousand: let me know if there's a different way to accommodate this: https://github.com/Path-Check/gaen-mobile/pull/370#discussion_r487446174

lmartensson: The documentation doesn't specify with sharp clarity, but just like the previous date fields were rounded by day, I assume this one is too. Let me double check but I'm fairly certain that this already is normalized by day. Rounding the date to start of day here (possibly in a different time zone from how it was rounded in the first place) is what leads to bugs like the one we had before where we couldn't tell what exact day an exposure happened on.

PR #372, opened 2020-09-14
Add learn more URL when it is present

amandabeiner:
## Description:
We want health authorities to be able to specify a "learn more" url that is different than the general authority advice url. This PR reads the LEARN_MORE_URL value from the environment, and sets it on the configuration context. If a learn more url is not provided, it default to the authority advice url.

#### Linked issues:

[GAEN-263](https://pathcheck.atlassian.net/jira/software/projects/GAEN/boards/33/?selectedIssue=GAEN-263)

### TODO:
Update `mobile_resources_commit` after [update](https://github.com/Path-Check/pathcheck-mobile-resources/pull/55) merges.

### DIFF
--- a/src/ConfigurationContext.tsx
+++ b/src/ConfigurationContext.tsx
@@ -11,6 +11,7 @@ export interface Configuration {
   displaySelfAssessment: boolean
   displaySymptomChecker: boolean
   healthAuthorityAdviceUrl: string
+  healthAuthorityLearnMoreUrl: string
   healthAuthorityEulaUrl: string | null
   healthAuthorityName: string
   healthAuthorityPrivacyPolicyUrl: string
@@ -27,6 +28,7 @@ const initialState = {
   displaySelfAssessment: false,
   displaySymptomChecker: false,
   healthAuthorityAdviceUrl: "",
+  healthAuthorityLearnMoreUrl: "",
   healthAuthorityEulaUrl: null,
   healthAuthorityName: "",
   healthAuthorityPrivacyPolicyUrl: "",
@@ -39,6 +41,7 @@ const ConfigurationContext = createContext<Configuration>(initialState)
 const ConfigurationProvider: FunctionComponent = ({ children }) => {
   const {
     AUTHORITY_ADVICE_URL: healthAuthorityAdviceUrl,
+    LEARN_MORE_URL: healthAuthorityLearnMoreUrl,
     GAEN_AUTHORITY_NAME: healthAuthorityName,
     PRIVACY_POLICY_URL: healthAuthorityPrivacyPolicyUrl,
     LEGAL_PRIVACY_POLICY_URL: legalPrivacyPolicyUrl,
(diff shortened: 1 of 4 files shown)


src/ExposureHistory/History/NoExposures.spec.tsx
```diff
@@ -24,29 +24,54 @@ describe("NoExposures", () => {
...
+        expect(
+          queryByText(`Review guidance from ${healthAuthorityName}`),
+        ).not.toBeNull()
+        fireEvent.press(getByText("Learn More"))
+        expect(openURLSpy).toHaveBeenCalledWith(healthAuthorityLearnMoreUrl)
+      })
+    })
+
+    describe("when the health authority has not provided a link", () => {
+      it("", () => {
```
johnschoeman:
```suggestion
      it("should not show a learn more link", () => {
```

src/ExposureHistory/History/NoExposures.spec.tsx
```diff
@@ -24,29 +24,54 @@ describe("NoExposures", () => {
...
-      expect(
-        queryByText(`Review guidance from ${healthAuthorityName}`),
-      ).not.toBeNull()
-      fireEvent.press(getByText("Learn More"))
-      expect(openURLSpy).toHaveBeenCalledWith(healthAuthorityAdviceUrl)
+  describe("clicking the Learn more link", () => {
+    describe("when the health authority has provided a learn more url", () => {
+      it("prompts the user to see HA guidance", () => {
+        expect.assertions(2)
+        const healthAuthorityLearnMoreUrl = "https://www.health.state.mn.us/"
```
johnschoeman:
```suggestion
        const healthAuthorityLearnMoreUrl = "https://www.health.example.com/"
```

src/ConfigurationContext.tsx
```diff
@@ -61,6 +64,7 @@ const ConfigurationProvider: FunctionComponent = ({ children }) => {
         displayReportAnIssue,
         displaySelfAssessment,
         healthAuthorityAdviceUrl,
+        healthAuthorityLearnMoreUrl: learnMoreUrl || healthAuthorityAdviceUrl,
```
johnschoeman: This jumps out as a bit subtle and prone to error/confusion.  Should we default to the advice url? or would it be better to enforce providing a url and otherwise not see a url at all? We could always duplicate the url in the config file itself if we want both to be the same.

johnschoeman (approved):
LGTM!

src/ConfigurationContext.tsx
aledustet: Not sure if we should default to the value from the advice URL, maybe this learn more link should not be rendered if there's not a `LEARN_MORE_URL` I'm not sure we got clarity on this when we discussed it, might be worth circling back.

src/ExposureHistory/History/NoExposures.spec.tsx
aledustet:
```suggestion
      it("does not display the learn more link", () => {
```

aledustet (approved):
Other than consulting the default value for the configuration, I think this one is ready to ⛵

src/ConfigurationContext.tsx
amandabeiner: @johnschoeman @aledustet Would you all recommend that I update the mobile resources repo to set the LEARN_MORE _URL to the AUTHORITY_ADVICE_URL in order to keep the current behavior for other HAs?

johnschoeman: Yes, i think that makes sense to do.

PR #374, opened 2020-09-14
Add content to Connect screen

devinjameson:
Why: we don't have any content on the Connect screen right now.

This commit:
- Adds content to the Connect screen based on the Figma mockups


johnschoeman (approved):
LGTM

src/locales/en.json
```diff
@@ -1,9 +1,10 @@
 {
   "_display_name": "English",
-  "about": {
-    "description": "The {{applicationName}} app is made available by {{healthAuthorityName}}",
+  "connect": {
+    "description": "The {{applicationName}} app is made available by {{healthAuthorityName}}.\n\nFor more information, visit the ",
```
aledustet, line 4: I think that this text should not change if this is meant to be the same as the About screen text. This text will be replaced by configuration from each of the authorities so maybe we should leave the default as is. Probably should confirm with product.

johnschoeman: We should rollback all of these language file changes and use the exact same text, as we dont have time to wait for the translations. We can updated it in a future commit.

devinjameson: I see, thanks that's helpful. I'm going to merge for now just to get this through but will update in the next pass!

PR #376, opened 2020-09-14
Reduce APK size

juancruzgs:
#### Description:

This PR enables [Hermes](https://reactnative.dev/docs/hermes) and [Proguard](https://developer.android.com/studio/build/shrink-code) to reduce the final APK size.

The file size went from `105MB` to `65MB` for production builds.
It will be reduced a bit more when the AAB file is uploaded to Play Store and it generates individual APKs for each architecture.

The code is now obfuscated for `staging` and `release` builds

#### Linked issues:

https://pathcheck.atlassian.net/browse/GAEN-282

### DIFF
--- a/android/app/proguard-rules.pro
+++ b/android/app/proguard-rules.pro
@@ -9,8 +9,17 @@
 
 # Add any project specific keep options here:
 
-# react-native-config, prevents obfuscating of .env flags
--keep class org.pathcheck.covidsafepaths.BuildConfig { *; }
+-keepattributes LineNumberTable,SourceFile
+
+# This allows proguard to strip isLoggable() blocks containing only <=INFO log
+# code from release builds.
+-assumenosideeffects class android.util.Log {
+  public static *** d(...);
+  public static *** v(...);
+  public static *** isLoggable(...);
+}
+
+-dontwarn android.support.annotation.**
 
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -42,7 +42,7 @@
     ]
   },
   "dependencies": {
-    "@bugsnag/react-native": "^7.3.2",
+    "@bugsnag/react-native": "^7.3.4",
```
juancruzgs, line 45: Previous versions of Bugsnag have some issues with Proguard, so I had to update to the latest one.
https://github.com/bugsnag/bugsnag-js/blob/next/CHANGELOG.md

android/app/proguard-rules.pro
```diff
@@ -9,8 +9,17 @@
...
--keep class org.pathcheck.covidsafepaths.BuildConfig { *; }
+-keepattributes LineNumberTable,SourceFile
+
+# This allows proguard to strip isLoggable() blocks containing only <=INFO log
+# code from release builds.
+-assumenosideeffects class android.util.Log {
+  public static *** d(...);
+  public static *** v(...);
+  public static *** isLoggable(...);
+}
```
juancruzgs, line 20: Logs are removed from staging/release builds.
I can add them only for staging if we find it useful for QA in the future

johnschoeman (approved):
LGTM!

PR #378, opened 2020-09-14
Fix bug in how it works final button

johnschoeman:
Why:
Currently the final button in the how it works flow has a bug where it
tries to navigate to the "Home" Screen, which no longer exists. The
intended behavior is to navigate to the destinationOnSkip value.

Co-Authored-By: devinjameson <devin@thoughtbot.com>


devinjameson (approved):
🚢  :shipit:

PR #381, opened 2020-09-15
Handle complete call back requested flow

aledustet:
Why:
----

The call back request flow needs to redirect to the origin of the start of such flow. Since it can be from the exposure details screen or the settings/connect screens.

How:
----

Using a simple context surrounding the flow will allow us to pivot the strategy, with extra layers of logic if we need to change it in the future, right now is a simple function that uses the context where the stack is defined to just navigate back from that point.

This Commit:
----

- Add a CallbackFormContext with a completed request handler
- Use the context from the callback form flow to navigate back after completing the flow

Large GIF (428x886)
Large GIF (428x886)

### DIFF
--- a/src/Callback/CallbackFormContext.tsx
+++ b/src/Callback/CallbackFormContext.tsx
@@ -1,0 +1,17 @@
+import { createContext, useContext } from "react"
+
+type CallbackFormContextState = {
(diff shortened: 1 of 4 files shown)


johnschoeman (approved):
LGTM

PR #383, opened 2020-09-15
Use "Number" Type for ENAPIVersion

mattThousand:


mobile_resources_commit
```diff
@@ -1 +1 @@
-e9c1e9da272e37e63bc8e53b55d3206f42e1ae32
```
aledustet, line 1: Is this update intentional here?

aledustet (approved):
Just double-checking the mobile resources SHA update other than that :shipit:

mobile_resources_commit
mattThousand: good question - yes, I updated to use `ENAPIVersion`: `2`

PR #384, opened 2020-09-15
Move request call back to connect screen

aledustet:
Why:
----

The request call back action should be accessible from both the exposure details and the connect screen

This Commit:
----

- Import SVG for the headset element on the connect screen
- Move request call back link from settings to connect menu

Caveat:
----

The connect screen menu should include sections but since some work is being done in that front, the re-arrangement is left for when that job is added

Screen Shot 2020-09-15 at 12 50 56 PM


src/Connect/index.tsx
```diff
@@ -86,6 +87,15 @@ const ConnectScreen: FunctionComponent = () => {
     listItems.push(reportAnIssue)
   }
 
+  if (displayCallbackForm) {
```
aledustet, line 99: @devinjameson I punted on adding the sections since that seems to be closer to your plate, but let me know if that fits in this scope and I can take on that(or we can 🍐 ) 😄

devinjameson: Sounds good! We can do it in another pass :)

devinjameson (approved):
LGTM!

src/Connect/index.tsx
```diff
@@ -86,6 +87,15 @@ const ConnectScreen: FunctionComponent = () => {
...
+    const callbackForm: ConnectListItem = {
```
devinjameson, line 100: I like this pattern, where you only create the `callbackForm` if `displayCallbackForm` is true. Is this better in some way than creating `callbackForm` outside the if block and then only pushing it to the list items array if `displayCallbackForm` is true? I'm curious as to why you would do it this way vs. the other.

src/assets/svgs/headset.ts
```diff
@@ -0,0 +1,5 @@
...
...3.2396 26.2278C13.236 26.258 13.236 26.2885 13.2396 26.3187C13.2396 26.5198 13.3195 26.7126 13.4617 26.8548C13.6039 26.997 13.7967 27.0769 13.9978 27.0769C16.09 27.0743 18.1384 26.4776 19.9047 25.3562C21.6709 24.2348 23.0824 22.6349 23.9749 20.7426C24.9782 20.6623 25.9231 20.2383 26.6502 19.5424C27.3774 18.8464 27.8423 17.921 27.9665 16.9221C28.0908 15.9233 27.8668 14.9121 27.3323 14.0592C26.7978 13.2063 25.9855 12.5639 25.0325 12.2401V12.2477ZM4.59309 13.5403V19.2643H4.36944C3.6104 19.2643 2.88245 18.9627 2.34572 18.426C1.809 17.8893 1.50747 17.1613 1.50747 16.4023C1.50747 15.6433 1.809 14.9153 2.34572 14.3786C2.88245 13.8419 3.6104 13.5403 4.36944 13.5403H4.59309ZM14.7559 25.5303V24.7721C14.7569 24.4068 14.9027 24.0569 15.1614 23.7989C15.42 23.541 15.7704 23.3961 16.1357 23.3961H17.3639C17.6572 23.379 17.9478 23.4608 18.1891 23.6285C18.4304 23.7962 18.6083 24.0401 18.6944 24.321C17.4868 25.0094 16.1418 25.4224 14.7559 25.5303ZM23.6261 19.2529H23.4025V13.5403H23.6261C24.3852 13.5403 25.1131 13.8419 25.6498 14.3786C26.1866 14.9153 26.4881 15.6433 26.4881 16.4023C26.4881 17.1613 26.1866 17.8893 25.6498 18.426C25.1131 18.9627 24.3852 19.2643 23.6261 19.2643V19.2529Z" fill="#1F2C9B"/>
```
devinjameson: It's probably a good idea to remove the `fill`s from this and set the fill on the icon itself.

src/Connect/index.tsx
aledustet: It was mostly to keep the usage close to the declaration to make it easier to read. There's not a lot there other than avoiding creating these things in memory unless we need them 🤷  removing one will definitely not have any performance impact but it's not a bad idea to have that on the back of our minds for when it matters I guess 😄

johnschoeman (approved):
LGTM

PR #385, opened 2020-09-15
Change launch screen icon size

juancruzgs:
#### Description:

Update the icon size to always take 50% of the screen width and set 1:1 aspect ratio.
Use `ic_launcher_foreground` instead of `ic_launcher` because the former has transparency.

#### Screenshots:

image


aledustet (approved):
:shipit:

PR #386, opened 2020-09-15
Scaffold out symptom checker

devinjameson:
Why: as a user, I would like to be able to tell the app how I'm feeling and be given a recommendation on how to proceed.

This commit:
- Adds a symptom checker tab
- Adds basic content to the symptom checker tab
- Changes the calls to `require` to `imports` in `images/index`

Screen Shot 2020-09-15 at 3 33 43 PM


src/assets/images/index.ts
```diff
@@ -1,32 +1,34 @@
...
-const PersonGettingNotification = require("./personGettingNotification.png")
-const PersonOnComputer = require("./personOnComputer.png")
-const PersonOnPhoneWithCode = require("./personOnPhoneWithCode.png")
-const PersonWithCodesInBackground = require("./personWithCodesInBackground.png")
-const PersonWithLockedPhone = require("./personWithLockedPhone.png")
-const PersonWithPhoneAndCode = require("./personWithPhoneAndCode.png")
-const PhonesSharingCodes = require("./phonesSharingCodes.png")
-const SingleCrossPathBackground = require("./singleCrossPathBackground.png")
-const WelcomeImage = require("./welcomeImage.png")
+import BlueGradientBackground from "./blueGradientBackground.png"
```
johnschoeman, line 1: are we able to use the es6 import syntax here? irc, this cause some typescript errors?

src/navigation/MainTabNavigator.tsx
```diff
@@ -114,6 +130,14 @@ const MainTabNavigator: FunctionComponent = () => {
           tabBarIcon: ExposureHistoryIcon,
         }}
       />
+      <Tab.Screen
```
johnschoeman: We will need to be able to conditionally render this based off of a config var as not every HA will want/need this. We also want to be able to cut release builds potentially while this is under active development.  This could come in a followup pr though.

johnschoeman (approved):
LGTM! one comment.

src/assets/images/index.ts
devinjameson: Seems to work just fine!

src/navigation/MainTabNavigator.tsx
devinjameson: Sounds good will handle in this PR.

PR #388, opened 2020-09-15
Add support for new configuration for legal page

aledustet:
Why:
----

With the list of links being configurable on both legal and about screens, having the links currently being rendered there being optional is necessary to avoid having duplicated ones. Also, links should not default to a language to allow for specialized links per locale.

This Commit:
----

- Remove default locale for links on legal and about screens
- Make authority privacy policy url optional

Reviewers:
----

This depends on [this PR](https://github.com/Path-Check/pathcheck-mobile-resources/pull/65)


johnschoeman (approved):
LGTM

yonatanmk (approved):

PR #389, opened 2020-09-15
Log errors from callback form

aledustet:
Why:
----

Once we have the call back form online we will need some visibility on the behavior of the same while we are testing on staging.

This Commit:
----

- Add logging logic to the call back form
- Add tests for the error handling on the component

### DIFF
--- a/src/Callback/Form.tsx
+++ b/src/Callback/Form.tsx
@@ -17,6 +17,8 @@ import { useStatusBarEffect, CallbackStackScreens } from "../navigation"
 import { useConfigurationContext } from "../ConfigurationContext"
 import { Button, GlobalText } from "../components"
 import * as API from "./callbackAPI"
+import Logger from "../logger"
+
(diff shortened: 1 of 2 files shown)


src/Callback/Form.tsx
```diff
@@ -66,10 +68,19 @@ const CallbackForm: FunctionComponent = () => {
       if (response.kind === "success") {
         navigation.navigate(CallbackStackScreens.Success)
       } else {
+        Logger.addMetadata("requestCallbackError", {
+          errorMessage: response.message,
+        })
+        Logger.error(
+          `FailureToRequestCallback.${response.error}.${
+            response.message || defaultErrorMessage
```
johnschoeman: defaultErrorMessage is assigned to`" "` here, should we be using something different for the logger?

johnschoeman (approved):
LGTM!

PR #393, opened 2020-09-16
Add support for callback success configuration

aledustet:
Why:
----

The call back request success message should be configured independently per authority.

This Commit:
----

- Add support for a new key on the authority copy helper
- Use copy override on the call back success screen

Reviewers:
----

This depends on a new configuration value on this PR https://github.com/Path-Check/pathcheck-mobile-resources/pull/68 but there's no need to update the repo commit since we don't have any overrides at the moment


johnschoeman (approved):
LGTM

PR #397, opened 2020-09-16
Recent Exposures iOS

mattThousand:
- Only return most recent 2 weeks of exposures from the iOS layer

- Workaround for an undocumented EN framework limitation whereby the cached exposures call (https://developer.apple.com/documentation/exposurenotification/enmanager/3650384-detectexposures) is limited to 6 per day.


johnschoeman (approved):
LGTM!

PR #398, opened 2020-09-16
Store bitrise configuration in repository

juancruzgs:
#### Description:

We have reached the max file size allowed in Bitrise (200KB). That limit only applies to configuration files stored in bitrise.io.
We are moving the configuration to `pathcheck-mobile-resources` repository (as recommended [here](https://devcenter.bitrise.io/bitrise-cli/basics-of-bitrise-yml/#bitriseyml-size-limitation)) to avoid that issue.

This PR adds a script to download the configuration and copy it to the root folder.
I'm also making some changes in Bitrise to use this configuration.

Depends on https://github.com/Path-Check/pathcheck-mobile-resources/pull/73


mattThousand (approved):
Thanks @juancruzgs!

johnschoeman (approved):

PR #400, opened 2020-09-16
Add UI for opting into analytics sharing

amandabeiner:
### Description: 

Why:
As a product team, we want to collect useful analytics data from users
in a trustworthy and privacy-focused way.

This commit:
Adds UI to the settings screen to allow users to opt into anonymized
data sharing. Before they opt in, they will have a chance to review how
we will use their data and view their HA's privacy policy.

This commit introduces:
- An AnonymizedDataConsentScreen, which displays information about how
  we will use their analytics data
- A ShareAnonymizedDataSwitch, which allows users to turn data sharing
  on or off
- A SwitchListItem, which establishes a pattern for switch settings
  which we will use going forward.

#### Linked issues:

[GAEN-264](https://pathcheck.atlassian.net/jira/software/projects/GAEN/boards/33/?selectedIssue=GAEN-264)

#### Screenshots:

toggly boi

#### How to test:

- Visit the settings screen
- You should see an analytics consent toggle that is switched off by default
- Switch it on
- You should be brought to a consent screen
- Click to consent
- You should be brought back to the settings screen, with the toggle switched on


johnschoeman (approved):
LGTM! the only thing is that we likely need to have this consent switch be toggle-able per a config var as not every build will want to include this for now.

src/Settings/AnonymizedDataConsentScreen.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const handleOnPressConsent = async () => {
+    await updateUserConsent(true)
+    navigation.navigate(Stacks.Settings)
+  }
+
+  const handleOnPressClose = () => {
+    navigation.navigate(Stacks.Settings)
+  }
+
+  return (
```
devinjameson, line 40: I think we need a `<StatusBar backgroundColor={Colors.primaryLightBackground} />` here.

src/Settings/AnonymizedDataConsentScreen.tsx
```diff
@@ -0,0 +1,113 @@
...
+    <ScrollView style={style.container}>
```
devinjameson: Does this need a `contentContainerStyle` with some bottom padding?

src/Settings/ShareAnonymizedDataSwitch.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { useNavigation } from "@react-navigation/native"
+
+import { AnalyticsContext } from "../AnalyticsContext"
+import ShareAnonymizedDataSwitch from "./ShareAnonymizedDataSwitch"
+import { factories } from "../factories"
+import { SettingsScreens } from "../navigation"
+
+jest.mock("@react-navigation/native")
+describe("toggling anonymized data consent", () => {
+  describe("when the user has consented analytics", () => {
```
devinjameson:
```suggestion
  describe("when the user has consented to analytics", () => {
```

App.tsx
```diff
@@ -48,7 +49,9 @@ const App: FunctionComponent = () => {
               <PermissionsProvider>
                 <SystemServicesProvider>
                   <ExposureProvider>
-                    <MainNavigator />
+                    <AnalyticsProvider>
```
devinjameson, line 52: Should this live in the `SettingsStack` if that's the only place we use it?

src/components/SwitchListItem.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const SwitchListItem: FunctionComponent<SwitchListItemProps> = ({
+  label,
+  isActive,
+  onChange,
+  testID,
+}) => {
+  return (
+    <View style={style.toggleContainer}>
+      <Text style={style.toggleText}>{label}</Text>
```
devinjameson, line 20: Does this need `GlobalText`?

devinjameson (approved):
LGTM! Few comments.

App.tsx
amandabeiner: This will soon be used basically everywhere (it's where the function to track analytics events lives) so we'll need access to it throughout the app!

PR #401, opened 2020-09-16
Add symptom selection functionality to symptom checker

devinjameson:
Why: as a user, I would like to be able to select some symptoms and be given a recommendation on how to proceed.

This commit:
- Adds the ability for a user to select symptoms from a list
- Gives the user a recommendation based on their symptoms. Right now, the scoring is just based on whether the user has selected more than 3 symptoms. Of course, this is a placeholder.
- Adds a screen that the user is navigated to if their symptoms indicate they may be at risk
- Adds a `SymptomCheckerContext`

gif2

### DIFF
--- a/src/SymptomChecker/CheckIn.tsx
+++ b/src/SymptomChecker/CheckIn.tsx
@@ -224,9 +224,11 @@ const style = StyleSheet.create({
   },
   checkInEyebrowText: {
     ...Typography.body1,
+    marginBottom: Spacing.xxxSmall,
   },
   checkInHeaderText: {
     ...Typography.header3,
+    paddingRight: Spacing.xxLarge,
   },
   feelingButtonsContainer: {
     flexDirection: "row",
--- a/src/SymptomChecker/CheckIn.tsx
+++ b/src/SymptomChecker/CheckIn.tsx
@@ -11,10 +11,11 @@ import { useNavigation } from "@react-navigation/native"
 import { SvgXml } from "react-native-svg"
 
 import { GlobalText } from "../components"
+import { useSymptomCheckerContext } from "./SymptomCheckerContext"
+import { SymptomCheckerStackScreens } from "../navigation"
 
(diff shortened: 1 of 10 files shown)


src/SymptomChecker/CheckIn.tsx
```diff
@@ -121,13 +138,44 @@ const FeelingGoodContent = () => {
...
+
+const FeelingNotWellContent: FunctionComponent<FeelingNotWellContentProps> = ({
+  symptoms,
+  healthAssessment,
+}) => {
   const { t } = useTranslation()
 
+  const determineHealthAssessmentText = () => {
+    switch (healthAssessment) {
+      case HealthAssessment.GetTested:
```
johnschoeman:
```suggestion
      case HealthAssessment.AtRisk:
```

src/SymptomChecker/CheckIn.tsx
```diff
@@ -121,13 +138,44 @@ const FeelingGoodContent = () => {
...
+        return t("symptom_checker.get_tested")
+      case HealthAssessment.FollowHAGuidance:
```
johnschoeman:
```suggestion
      case HealthAssessment.NotAtRisk:
```

This naming will leave us open to keep the notion of the assessment flow separate from the recommendation flow and allow them to grow/change independently. for example we might have a future world where we want something like

```typescript
type HealthAssessment = NotStarted | InProgress | AtRisk | NotAtRisk

type HealthRecommendation = GetTested | CallDoctor | DoNothing
```

its a bit premature to finalize what exactly these states should be but we probably want to be moving towards something like this to support the self assessment feature.

johnschoeman (approved):
LGTM!

src/SymptomChecker/CheckIn.tsx
devinjameson: Sounds good!

PR #402, opened 2020-09-16
Handle authorization failure from call back api

aledustet:
Why:
----

Errors from parts of the operation requesting a call back should be captured and logged on a staging environment so we can gain insight into its behavior.

This Commit:
----

- Handle OAuth response failures separately
- Add special case for when an error comes from an exception-handling the request


johnschoeman (approved):
Nice!

PR #404, opened 2020-09-17
Update node-fetch

johnschoeman:


package.json
```diff
@@ -55,6 +55,7 @@
     "dayjs": "^1.8.24",
     "i18next": "^19.3.3",
     "lodash.isequal": "^4.5.0",
+    "node-fetch": "^2.6.1",
```
aledustet, line 58: I'm a little curious about why we are pulling this library but not using it, did we forget a commit on this PR?

johnschoeman: trying to get the security vulnerability patch in without having to update all the other upstream libraries that depend on this one.  so being a bit lazy and you're right, we should do this the proper way. Ill close this and cut a card for the work.

drunkle-rocket: I believe this is just forcing node-fetch to update in response to the recent vulnerability in it, without updating react-native entirely.

si1k:
Closing the loop on this.  After merging https://github.com/Path-Check/gaen-mobile/pull/562 the node-fetch dependency will update.  This should clean up the outdated node-fetch in most, if not all, nested dependencies.

PR #406, opened 2020-09-17
Fix display issues on AcceptTermsOfService

devinjameson:
Why: the `AcceptTermsOfService` screen broke when using large text sizes.


johnschoeman (approved):
LGTM

PR #411, opened 2020-09-17
Add custom gradient style for buttons

aledustet:
Why:
----

On some android devices, there is a background that shows on some of the buttons in the application. We need the buttons to be displayed consistently among all the platforms.

Screen Shot 2020-09-17 at 2 45 23 PM

How:
----

The custom styles that are being applied to the buttons are being added both for the button and for the inner gradient used to display the color pattern on the buttons. On android devices, this style gets added twice resulting on the white bottom lines we can see in some of the buttons. We need as a first fix to separate those customizations. Meaning that style overrides for our buttons will have to take into consideration both the button and the inner gradient on it.

This Commit:
----

- Add a new argument for the button component to receive custom gradient
styles
- Update buttons on the how it works, activation summary and home
screens to use the new API

Co-authored-By: devinjameson <devin@thoughtbot.com>

Screen Shot 2020-09-17 at 2 45 23 PM Screen Shot 2020-09-17 at 2 45 41 PM 
Screen Shot 2020-09-17 at 2 45 51 PM Screen Shot 2020-09-17 at 2 45 54 PM

### DIFF
--- a/src/Activation/ActivationSummary.tsx
+++ b/src/Activation/ActivationSummary.tsx
@@ -50,6 +50,7 @@ const ActivationSummary: FunctionComponent = () => {
           label={t("common.settings")}
           onPress={handleOnPressOpenSettings}
           customButtonStyle={style.primaryButton}
+          customButtonInnerStyle={style.primaryButtonGradient}
         />
         <TouchableOpacity
           onPress={handleOnPressGoToHome}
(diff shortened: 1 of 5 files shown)


devinjameson (approved):
LGTM! One thought, but I think this is good to go as is.

Would it be more clear if we called what is currently `buttonContainerStyle` => `buttonStyle`? Then we could call the `buttonStyle` => `buttonInnerStyle` and `customGradientStyle` => `customButtonInnerStyle`.

Food for thought! 🤔

johnschoeman (approved):
Nice!

PR #412, opened 2020-09-17
Add expanding circle animation to home screen

devinjameson:
Why: the initial designs for the home screen had an animated expanding circle, but we punted on implementing that feature due to time constraints.

This commit:
- Adds an expanding circle animation to the home screen when the app is active
- Does not show the animation if the user has turned on reduce motion in their phone settings

gif


johnschoeman (approved):
Neat

aledustet (approved):
:shipit:

PR #414, opened 2020-09-17
Validate exposure keys from native module

johnschoeman:
Why:
Currently we have a bug in our android builds that was caused by the
gaen native module getExposureKeys() function returning empty strings,
this lead to an error in key submission, but we would like to be able to
raise this error and log it sooner in the process.

This commit:
Adds a validate step to NativeModule.getExposureKeys and logs an error
with meta data if the validation fails.


aledustet (approved):
🚢  :shipit: 

We should keep an eye on staging after merging it because of the javascript runtime switch on a release build. But it should be ok.

PR #418, opened 2020-09-18
(Android) Store check in status in Realm

juancruzgs:
#### Description:

This PR exposes native methods to store `Check In` status in Realm.


android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/storage/Migration.kt
```diff
@@ -31,6 +31,14 @@ internal class Migration : RealmMigration {
...
         if (version == 3L) {
             schema.remove("PositiveDiagnosis")
+            version++
+        }
+
+        if (version == 4L) {
+            schema.create("CheckInStatus")
+                .addField("posixDate", Long::class.java, FieldAttribute.PRIMARY_KEY)
+                .addField("feeling", Boolean::class.java)
+                .addRealmListField("symptoms", String::class.java)
```
aledustet, line 42: Looking through some of the mocks, one of the requirements was to edit a particular entry of the symptoms log. Does Realm adds an id automatically that we can expose to identify each entry for edition?

aledustet (approved):
Looks solid to me, just one question about identifying each record for editing them.

android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/storage/Migration.kt
juancruzgs: Realm doesn't create IDs automatically, but we could generate random IDs ourselves:
> Autoincrementing IDs are not supported by Realm by design. This is primarily because it is impossible to generate such keys in a distributed environment, and compatibility between data stored in a local Realm and a synchronized Realm is a high priority. Note that Realm does not need primary keys in order to create relationships.

We are using the date as the primary key, is that enough?
We should only have one entry per date, so you could search by date (I can expose a method to do that).

aledustet: I believe that the interface supports multiple entries per day and we should group them by date on the UI so maybe using the date as a primary key might not solve our issue, it's hard that two entries happen on the same timestamp but an auto-generated id would take care of that issue. Thoughts?

juancruzgs: Ah okay, I didn't know that. I can add an ID as the primary key

juancruzgs:
@aledustet I updated the Realm fields, please review it again when you have some time

aledustet (approved):
🚢  :shipit:

PR #419, opened 2020-09-18
Update symptom checker flow

devinjameson:
Why: as a user, I would like to be able to tell the app my symptoms and be given a link
where I can find a place nearby to get tested for COVID-19.

This commit:
- Adds content to the `AtRiskRecommendation` screen
- Always navigates users to the `AtRiskRecommendation` screen regardless of their symptoms
- Updates the content of the `CheckIn` screen to also include a button to find a test center

gif


src/SymptomChecker/AtRiskRecommendation.tsx
```diff
@@ -19,6 +26,10 @@ const AtRiskRecommendationScreen: FunctionComponent = () => {
     navigation.navigate(SymptomCheckerStackScreens.SymptomChecker)
   }
 
+  const handleOnPressFindTestCenter = () => {
+    Linking.openURL("https://google.com")
```
devinjameson: Placeholder.

src/SymptomChecker/CheckIn.tsx
```diff
@@ -149,10 +150,23 @@ const FeelingNotWellContent: FunctionComponent<FeelingNotWellContentProps> = ({
 }) => {
   const { t } = useTranslation()
 
+  const handleOnPressFindTestCenter = () => {
+    Linking.openURL("https://google.com")
```
devinjameson: Placeholder.

src/ConfigurationContext.tsx
```diff
@@ -73,6 +76,7 @@ const ConfigurationProvider: FunctionComponent = ({ children }) => {
         displayReportAnIssue,
         displaySelfAssessment,
         displaySymptomChecker,
+        findATestCenterUrl,
```
aledustet, line 79: This means this URL is not being localized. This is perfectly fine, maybe worth asking if it is supported by the authority or if that's a desire.

aledustet (approved):
Looking goood

src/ConfigurationContext.tsx
johnschoeman: for now we dont need to support localization of the link here.  We can add it as an enhancement if turns out to be a requirement

src/SymptomChecker/AtRiskRecommendation.tsx
```diff
@@ -71,8 +112,32 @@ const style = StyleSheet.create({
   cancelButtonInnerContainer: {
     padding: Spacing.medium,
   },
+  image: {
+    resizeMode: "contain",
+    width: "97%",
+    height: 300,
```
johnschoeman: Does this fixed height force a scroll on small format screens?

src/SymptomChecker/symptoms.ts
```diff
@@ -6,10 +6,7 @@ export enum HealthAssessment {
...
 
 export const determineHealthAssessment = (
-  symptoms: Symptom[],
+  _symptoms: Symptom[],
 ): HealthAssessment => {
-  if (symptoms.length > 3) {
-    return HealthAssessment.AtRisk
-  }
-  return HealthAssessment.NotAtRisk
+  return HealthAssessment.AtRisk
```
johnschoeman: We probably want `if (symptoms.length > 0)` here.

johnschoeman (approved):
LGTM

src/SymptomChecker/symptoms.ts
devinjameson: Just curious why?

src/SymptomChecker/AtRiskRecommendation.tsx
devinjameson: Yeah it does, I'll make this smaller and reduce the text size to avoid scrolling shenanigans.

PR #423, opened 2020-09-18
Use flexWrap for recommended actions list to mitigate text overlap on small screen sizes

mattThousand:
JIRA issue: https://pathcheck.atlassian.net/jira/software/projects/GAEN/boards/33?selectedIssue=GAEN-300

BEFORE:
Screen Shot 2020-09-18 at 11 14 35 AM

AFTER:
Screenshot_1600464619

### DIFF
--- a/src/ExposureHistory/detail/ExposureActions.tsx
+++ b/src/ExposureHistory/detail/ExposureActions.tsx
@@ -11,6 +11,7 @@ import { useConnectionStatus } from "../../hooks/useConnectionStatus"
 import { Colors, Iconography, Spacing, Typography, Buttons } from "../../styles"
 import { Icons } from "../../assets"
 import { useConfigurationContext } from "../../ConfigurationContext"
+import { screenWidth } from "../../styles/layout"
 
 const ExposureActions: FunctionComponent = () => {
   const { t } = useTranslation()
@@ -41,7 +42,13 @@ const ExposureActions: FunctionComponent = () => {
               healthAuthorityName,
             })}
           </GlobalText>
-          <View style={style.recommendations}>
+          <View
+            style={
+              screenWidth > 350
+                ? [style.row, style.recommendations]
+                : [style.column, style.recommendations]
+            }
+          >
             <RecommendationBubble
               icon={Icons.IsolateBubbles}
               text={t("exposure_history.exposure_detail.isolate")}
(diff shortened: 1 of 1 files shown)


src/ExposureHistory/detail/ExposureActions.tsx
```diff
@@ -11,6 +11,7 @@ import { useConnectionStatus } from "../../hooks/useConnectionStatus"
 import { Colors, Iconography, Spacing, Typography, Buttons } from "../../styles"
 import { Icons } from "../../assets"
 import { useConfigurationContext } from "../../ConfigurationContext"
+import { screenWidth } from "../../styles/layout"
```
devinjameson:
```suggestion
import { Layout } from "../../styles"
```

devinjameson: And then further down:

`Layout.screenWidth > 350`

devinjameson: This is the pattern we've been using elsewhere :)

src/ExposureHistory/detail/ExposureActions.tsx
```diff
@@ -41,7 +42,13 @@ const ExposureActions: FunctionComponent = () => {
               healthAuthorityName,
             })}
           </GlobalText>
-          <View style={style.recommendations}>
+          <View
+            style={
+              screenWidth > 350
```
devinjameson: Thoughts on wrapping this up in something like `recommendationsStyle`?

```
const recommendationsStyle = screenWidth > 350
  ? [style.row, style.recommendations]
  : [style.column, style.recommendations]
…
<View style={recommendationsStyle} />
```

src/ExposureHistory/detail/ExposureActions.tsx
```diff
@@ -41,7 +42,13 @@ const ExposureActions: FunctionComponent = () => {
...
+                ? [style.row, style.recommendations]
```
devinjameson: Thoughts on making `row` the default? Something like:

```
? { ...style.recommendations }
: { ...style.recommendations, ...style.recommendationsColumn }
…
recommendations: {
  display: "flex",
  flexDirection: "row",
  …
recommendationsColumn: {
  flexDirection: "column"
}
```

I also prefer the object spread style but totally up to personal preference!

devinjameson (approved):
Looks good! Few comments.

It may be worth consulting with someone on design to see if the column style works.

Also, is it possible to solve this problem by adding some horizontal margin between each `RecommendationBubble` instead of conditionally changing the layout? 🤔

johnschoeman (approved):
Plus one with checking with design if this is okay.  I think an alternative would be to use something like flex wrap: https://reactnative.dev/docs/layout-props#flexwrap; which is probably similar enough to the current to move forward with without design review.

PR #425, opened 2020-09-18
(iOS) Store check in status in Realm

mattThousand:
Description:

This PR exposes native methods to store Check In status in Realm.

### DIFF
--- a/ios/BT/API/Model/CheckInStatus.swift
+++ b/ios/BT/API/Model/CheckInStatus.swift
@@ -1,0 +1,39 @@
+import ExposureNotification
+import Foundation
+import RealmSwift
+
+@objcMembers
+class CheckInStatus: Object, Codable {
+  @objc dynamic var id: String = .default
+  @objc dynamic var date: Int = 0
+  var feelingGood: Bool = false
+  var symptoms: List<String> = List<String>()
+
+  init(date: Int,
+       feelingGood: Bool,
+       symptoms: [String]) {
+    self.date = date
+    self.feelingGood = feelingGood
+    let list = List<String>()
+    list.append(objectsIn: symptoms)
+    self.symptoms = list
+    super.init()
+  }
+
+  required init() {
+    super.init()
+  }
+
+  override class func primaryKey() -> String? {
+    "date"
+  }
+
+  var asDictionary : [String: Any] {
+    return [
+      "posixDate": date,
+      "feelingGood": feelingGood,
+      "symptoms": Array(symptoms)
+    ]
+  }
+
+}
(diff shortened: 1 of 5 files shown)


devinjameson (approved):
This is at least 50 meters over my head but very cool and I want to learn how to do it 👺

ios/BT/API/Model/CheckInStatus.swift
```diff
@@ -0,0 +1,39 @@
+import ExposureNotification
+import Foundation
+import RealmSwift
+
+@objcMembers
+class CheckInStatus: Object, Codable {
+  @objc dynamic var id: String = .default
+  @objc dynamic var date: Int = 0
+  var feelingGood: Bool = false
```
aledustet: At the risk of jumping the gun here. Is there a world where this becomes more than a flag? Maybe the `feeling` can hold 3 states? Should we store an enum/integer here? cc @juancruzgs we do have an enum already on the typescript side. Definitely an improvement to make down the road but if we are just starting maybe a 5 min chat is worth it?

ios/BT/API/Model/CheckInStatus.swift
```diff
@@ -0,0 +1,39 @@
...
+    super.init()
+  }
+
+  override class func primaryKey() -> String? {
+    "date"
+  }
+
+  var asDictionary : [String: Any] {
+    return [
+      "posixDate": date,
```
aledustet, line 35: Is this the serialized structure that would be exposed? How do you feel about exposing the id? We will probably needed to render collections.

aledustet (commented):
Looks good to me a couple of questions about the data model but I think it matches the android side so feel free to move forward and we can regroup afterwards.

aledustet (approved):

ios/BT/API/Model/CheckInStatus.swift
juancruzgs: If there are more than 2 states to store then yes, we can change this to a Integer

ios/BT/API/Model/CheckInStatus.swift
mattThousand: hm, our thinking was that the `posixDate` would be the record's id, because no two records could exist for the same exact date, so we have a unique identifier at hand without needing to add an additional `id` property

aledustet: The requirements seem to indicate that we could have multiple symptom entries per day, so we might make use of this unique identifier more often than not.

mattThousand: wouldn't the posixDate be unique because it is expressed in seconds? 🤔

mattThousand: we were imagining that the date sent down from the JS layer would be a precise timestamp (i.e. not rounded to the day)

juancruzgs:
@mattThousand I made some changes here https://github.com/Path-Check/gaen-mobile/pull/418/commits/b1a48491af04d20750bda6fb88003f34a4765aea to support what's needed by the UI.

Changes:
- `feelingGood` from Boolean to Int
- Add `id` as the primary key
- Expose method `getCheckInStatus(id: String)`

PR #427, opened 2020-09-21
Add tests for onboarding context

amandabeiner:
Why:
As a team, we want to be confident that the app's onboarding behavior is
acting as expected.

This commit:
Adds a spec for the OnboardingContext

### DIFF
--- a/src/OnboardingContext.spec.tsx
+++ b/src/OnboardingContext.spec.tsx
@@ -1,0 +1,78 @@
+import React, { FunctionComponent, useEffect } from "react"
+import { render } from "@testing-library/react-native"
+
+import { OnboardingProvider, useOnboardingContext } from "./OnboardingContext"
+import { StorageUtils } from "./utils"
+import { Text } from "react-native"
+
+describe("OnboardingContext", () => {
+  describe("getting the current onboarding completion status", () => {
+    it("passes down the correct onboarding status to its children", async () => {
+      expect.assertions(1)
+
+      const { getByText } = render(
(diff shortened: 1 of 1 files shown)


src/OnboardingContext.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+      expect(storageSpy).toHaveBeenCalled()
+      expect(getByText(/onboarding complete/)).toHaveTextContent("false")
+    })
+  })
+})
+
+const OnboardingStatus: FunctionComponent = () => {
+  const { isOnboardingComplete } = useOnboardingContext()
+  console.log("in spec component", { isOnboardingComplete })
```
johnschoeman: extra console.log?

johnschoeman (approved):
I dig the test style. I usually hook up dummy button and exercise them with a fireEvent call, the useEffect is a neat alternative.

src/OnboardingContext.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+        </OnboardingProvider>,
+      )
+
+      expect(storageSpy).toHaveBeenCalled()
+      expect(getByText(/onboarding complete/)).toHaveTextContent("true")
+    })
+  })
+
+  describe("resetting onboarding", () => {
+    it("sets onboarding incomplete in storage and context", () => {
```
devinjameson:
```suggestion
    it("sets onboarding to incomplete in storage and context", () => {
```

devinjameson (approved):
Very interesting, haven't seen any tests like this before but I dig it!

PR #428, opened 2020-09-21
Parse dates in UTC

juancruzgs:
#### Description:

We were converting posix dates to the local timezone and then trying to calculate the start of day.
That resulted in exposures from **Friday 18 Sep** being shown as if they happened on **Thursday 17 Sep**.
That only happened on timezones with values less than UTC (I'm at GMT-3).

This PR fixes that and now displays the correct dates in Exposure History and Debug Menu

#### Linked issues:

https://pathcheck.atlassian.net/browse/GAEN-309

#### Screenshots:

Before:

device-2020-09-21-103258

After:

device-2020-09-21-150813


johnschoeman (approved):
LGTM!

PR #429, opened 2020-09-21
Add segmented control for symptoms log

devinjameson:
Why:
----

As a user, I would like to be able to see my symptom log in addition to
the symptoms I logged today.

This commit:
----

- Add react-native-segmented-control library
- Add first draft of SymptomLogContext
- Add OverTime screen to show symptom log
- Add spec for OverTime screen
- Rename SymptomChecker to MyHealth
- Rename CheckIn to Today

Reviewers:
----

- The SymptomLogContext is not hooked up to the native layer, so we are
  just using placeholders for now for that data
- Next steps involve creating the rest of the actions for the symptom
  log and implementing all of the UI

gif


src/MyHealth/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { GlobalText, StatusBar } from "../components"
+import Today from "./Today"
+import OverTime from "./OverTime"
+
+import { Typography, Colors, Spacing } from "../styles"
+
+const MyHealthScreen: FunctionComponent = () => {
+  useStatusBarEffect("dark-content", Colors.primaryLightBackground)
+  const { t } = useTranslation()
+  const [selectedIndex, setSelectedIndex] = useState<number>(0)
```
amandabeiner, line 16: This is really minor, but what do you think about naming this in terms of the UI, rather than the underlying implementation `SegmentedControl`? Perhaps with an enum, which is a `number` under the hood? Could be something like
```
enum MyHealthScreen {
   Today,
    Overtime
}

const [activeTab, setActiveTab] = useState<MyHealthScreen>(MyHealthScreen.Today)

const determineTodayStyles = () => {
   return activeTab === MyHealthScreen.Today ? {} : style.hidden
}
```

amandabeiner (approved):
LGTM! 👍  Left one comment for your consideration

ios/Podfile.lock
```diff
@@ -1,509 +0,0 @@
-PODS:
```
johnschoeman, line 1: probably want to keep this file

src/MyHealth/OverTime.tsx
```diff
@@ -0,0 +1,27 @@
...
+const OverTime: FunctionComponent = () => {
+  const { t } = useTranslation()
+  const { logEntries } = useSymptomLogContext()
+
+  return (
+    <View>
+      {logEntries.map((logEntry) => {
+        const healthAssessmentText =
+          logEntry.healthAssessment === HealthAssessment.AtRisk
+            ? t("my_health.symptom_log.feeling_not_well")
```
johnschoeman, line 18: If this is an intermediate bit of text, that is not intended for the final release, it would probably be better to not have it go through the translation process, so we could consider inline the string here.

src/MyHealth/SymptomLogContext.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+import { SymptomLogEntry, HealthAssessment } from "./symptoms"
+
+const fetchLogEntries = (): Promise<SymptomLogEntry[]> => {
+  return Promise.resolve([
+    {
+      id: 1,
+      symptoms: [],
+      healthAssessment: HealthAssessment.NotAtRisk,
+      date: new Date().getTime() / 1000,
```
johnschoeman, line 17: is there a reason we are doing seconds rather than ms here?

src/MyHealth/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const determineDisplayToday = () => {
```
johnschoeman, line 18: Minor thought building off of Amanda's comment of using a specific type for the tabs.

```suggestion
  const determineTabStyle = (myHealthTab: MyHealthTab) => {
     if (activeTab === myHealthTab) {
        return {}
     } else {
       return style.hidden
     }
     
     ...
             <View style={determineTabStyle(TodayTab)}>
          <Today />
        </View>
        <View style={determineTabStyle(OverTimeTab)}>
          <OverTime />
        </View>
```

johnschoeman (approved):
LGTM, a few minor thoughts

src/MyHealth/SymptomLogContext.tsx
aledustet: Not at all, this was just some dumb data we added for displaying something purposes, happy to remove as soon as we connect the data provider.

src/MyHealth/index.tsx
aledustet: Happy to do this, maybe in a second pass since we do `need` an actual index for the segmented control, and even if the types agree with an enum, it feels like a stretch to derive anevent argument into an enum. Thoughts?

PR #430, opened 2020-09-21
Track screen views for consenting HAs and users

amandabeiner:
Why:
As a health authority, I want to be able to gain insights into how
community members are using my app so that I can make better decisions
about contact tracing needs going forward.

This commit:
Updates the MainNavigator to track screen views if the health
authority has provided us analytics credentials, and the user has consented
to anonymous data sharing.

Changes include:
- Updating the ConfigurationContext to use the HA's
 analytics platform
- Updating the AnalyticsContext to explicitly opt-out of analytics if
  the HA has not provided credentials. The AnalyticsContext previously
worked this way in practice, but this makes it explicit.
- Upgrade react-navigation/native to 5.7.3, which supports an `onReady`
  prop on the NavigationContainer, which we use to set the initial route
if analytics tracking is enabled
- Update AnalyticsContext tests to test all cases and ensure we are
only collecting data if both the HA and user have consented.

#### Linked issues:

Fixes [GAEN-292](https://pathcheck.atlassian.net/browse/GAEN-292)

#### Screenshots:

Screen Shot 2020-09-21 at 3 03 08 PM


src/AnalyticsContext.tsx
```diff
@@ -35,18 +47,39 @@ const AnalyticsProvider: FunctionComponent = ({ children }) => {
...
     setUserConsentedToAnalytics(consent)
   }
 
   const trackEvent = async (event: string) => {
-    return userConsentedToAnalytics && actions.trackEvent(event)
+    return supportAnalyticsTracking && actions.trackEvent(event)
+  }
+
+  const trackScreenView = async (screen: string) => {
+    supportAnalyticsTracking && actions.trackScreenView(screen)
```
aledustet: How do you feel about creating a `voidState` of sorts and then on the initial mount we provide that value to the provider unless analytics tracking is supported? Does it feel like having the short circuit in both these places might make it hard to follow? Or is it better to have the checks right in front of the tracking event?

src/navigation/MainNavigator.tsx
```diff
@@ -29,9 +33,31 @@ const settingsStackTransitionPreset = Platform.select({
 
 const MainNavigator: FunctionComponent = () => {
   const { isOnboardingComplete } = useOnboardingContext()
+  const { trackScreenView } = useAnalyticsContext()
+  const navigationRef = React.useRef<NavigationContainerRef>(null)
```
aledustet: Just a small thing but how do you feel about importing the hook directly from react and avoid the `React` prepend?

src/navigation/MainNavigator.tsx
```diff
@@ -29,9 +33,31 @@ const settingsStackTransitionPreset = Platform.select({
...
+    if (currentRouteName && previousRouteName !== currentRouteName) {
+      trackScreenView(currentRouteName)
+    }
+
+    routeNameRef.current = currentRouteName
+  }
 
   return (
-    <NavigationContainer>
+    <NavigationContainer
```
aledustet, line 63: Nice single point for tracking routes 👍

aledustet (approved):
🚢  :shipit:

src/navigation/MainNavigator.tsx
johnschoeman: Yeah! nice and clean! 💯

src/AnalyticsContext.tsx
```diff
@@ -35,18 +55,47 @@ const AnalyticsProvider: FunctionComponent = ({ children }) => {
     checkAnalyticsConsent()
   }, [userConsentedToAnalytics])
 
+  useEffect(() => {
+    const supportAnalyticsTracking =
+      healthAuthoritySupportsAnalytics && userConsentedToAnalytics
+
+    const initializeAnalyticsTracking = () => {
```
johnschoeman, line 62: say a user updates their consent multiple times would we initialize this multiple times?   Should this be behind a subscription?

johnschoeman (approved):
LGTM! a few questions.

PR #431, opened 2020-09-22
GAEN-301

mattThousand:
Navigate to `Exposure History` screen when opening local notification informing user of possible exposure

Sep-22-2020 12-35-39


johnschoeman (approved):
LGTM!  should we cut a card for the android version of this?

and could you also make sure to have the final commit message to be something like:
`Navigate to exposure history screen when tapping native exposure notification`

juancruzgs:
I crashes on Android with the error
```
Error: Looks like you have configured linking in multiple places. This is likely an error since deep links should only be handled in one place to avoid conflicts. Make sure that:
    - You are not using both 'linking' prop and 'useLinking'
    - You don't have 'useLinking' in multiple components
    - You have set 'android:launchMode=singleTask' in the '<activity />' section of the 'AndroidManifest.xml' file to avoid launching multiple instances
```    

I'm still trying to figure out what the error is, please don't merge until it's fixed

juancruzgs:
I also added the code to open Exposure History on Android.
The only issue I found is that exposure history won't be opened from the notification if the application has been removed from recent tasks. I assume that happens because the `NavigationContainer` is not created yet and it has no way to handle the deep link.

Does it also happen on iOS?
We can investigate if there's any way to fix it in a future PR

juancruzgs (approved):

devinjameson (approved):
🆒

PR #432, opened 2020-09-22
Serialize daily check ins and symptom logs

aledustet:
Why:
----

There are going to be two sets of data captured as part of the My Health initiative. The first aspect will be a daily check-in. Shown on the  `Today` portion of the My Health screen. The other portion is multiple entries per day of logging symptoms. These two separate sets of data should compose what is presented to the users on the `Over Time` tab of the my health screen.

How:
----

We learned that the data captured on this screen should not be treated as one. There are two distinct sets of data:

- the check in: daily, and it only captures the status, feeling well or not
- the symptoms: multiple times a day. It captures a set of symptoms logged by the user at any given time

The two data sets need to be presented to the user in a way that:
- if any of the two is missing. The entry for the day should still appear
- if both are present, a single entry for the day should be added, with multiple instances for each entry of symptoms
- if none are present for the day, this day should not appear on the UI

This Commit:
----

- Add a grouping logic for symptom log entries
- Change data structure to be displayed top respond to needs of merging two different sets of data
- Hooked the log context to the new flow of data serialization
- Update the Over Time screen to display the transformed and joined daily check in and symptom log entries

Co-authored-by: Matt Buckley <matt.buckley@rocketinsights.com>

Screen Shot 2020-09-22 at 5 16 34 PM

### DIFF
--- a/src/MyHealth/symptoms.ts
+++ b/src/MyHealth/symptoms.ts
@@ -1,4 +1,4 @@
-import { Posix } from "../utils/dateTime"
+import { Posix, beginningOfDay } from "../utils/dateTime"
 
 export enum HealthAssessment {
   AtRisk,
@@ -7,10 +7,33 @@ export enum HealthAssessment {
 
 export type Symptom = string
 
+export enum CheckInStatus {
+  NotCheckedIn,
+  FeelingGood,
+  FeelingNotWell,
+}
+
+// Raw Daily Checkin
+export type DailyCheckIn = {
+  date: Posix
+  status: CheckInStatus
+}
+
+// Raw Log Entry
 export type SymptomLogEntry = {
-  id: number
+  id: string
+  date: Posix
   symptoms: Symptom[]
-  healthAssessment: HealthAssessment
+}
+
+type LogData = {
+  checkIn: DailyCheckIn | null
+  logEntries: SymptomLogEntry[]
+}
+
+type LogDataByDay = Record<Posix, LogData>
+
+export type DayLogData = LogData & {
   date: Posix
 }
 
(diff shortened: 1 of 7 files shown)


src/MyHealth/symptoms.ts
```diff
@@ -22,3 +45,81 @@ export const determineHealthAssessment = (
...
+
+const groupLogEntriesDaily = (allEntries: SymptomLogEntry[]): LogDataByDay => {
+  const groupedEntries: LogDataByDay = {}
+
+  allEntries.forEach((entry) => {
+    const { date } = entry
+    const entryDateBeginningOfDay = beginningOfDay(date)
+
+    if (groupedEntries[entryDateBeginningOfDay]) {
+      const newLogEntries = groupedEntries[entryDateBeginningOfDay].logEntries
```
devinjameson, line 71: Just curious why doesn't this need to be spread?

`const newLogEntries = [...groupedEntries[entryDateBeginningOfDay].logEntries]`

src/MyHealth/symptoms.ts
```diff
@@ -22,3 +45,81 @@ export const determineHealthAssessment = (
...
+        checkIn: null,
+      }
+    }
+  })
+
+  return groupedEntries
+}
+
+const joinLogDataAndDailyCheckIns = (
+  logData: LogDataByDay,
```
devinjameson, line 87: Thoughts on calling this `logDataByDay`? Maybe it's just me but I'm getting a bit turned around with the types.

```suggestion
  logDataByDay: LogDataByDay,
```

src/MyHealth/symptoms.ts
```diff
@@ -22,3 +45,81 @@ export const determineHealthAssessment = (
...
+  return Object.keys(dailyLogData)
+    .map((date: string) => {
+      const posixDate = Number(date)
+      return {
+        date: posixDate,
+        ...dailyLogData[posixDate],
+      }
+    })
+    .sort(compareDatesDescending)
+}
```
devinjameson, line 125: This is intense 😮  🔥

src/MyHealth/symptoms.ts
```diff
@@ -22,3 +45,81 @@ export const determineHealthAssessment = (
...
+  symptomLogEntries: SymptomLogEntry[],
+  dailyCheckIns: DailyCheckIn[],
+): DayLogData[] => {
+  const symptomEntriesLogData = groupLogEntriesDaily(symptomLogEntries)
+  const dailyLogData = joinLogDataAndDailyCheckIns(
+    symptomEntriesLogData,
+    dailyCheckIns,
+  )
+
+  return Object.keys(dailyLogData)
```
devinjameson, line 116: Is it worth naming this object we return?

devinjameson (approved):
🥇

src/MyHealth/symptoms.ts
```diff
@@ -7,10 +7,33 @@ export enum HealthAssessment {
...
 export type Symptom = string
 
+export enum CheckInStatus {
+  NotCheckedIn,
+  FeelingGood,
+  FeelingNotWell,
+}
+
+// Raw Daily Checkin
+export type DailyCheckIn = {
```
mattThousand, line 17: ~DailyCheckIn~ `CheckInEntry`
~SymptomLogEntry~ `SymptomEntry`
~LogData~ `DailyEntry`
~LogDataByDay~ `DateDailyEntryMap` 
🤔 🧐

src/MyHealth/symptoms.ts
aledustet: Since we are reading a value from an object, and it is coming from an argument it doesn't get passed as a reference, this is an actual copy of it. I think it's called call by sharing or something like that 🤔 if we were to modify the actual grouped entries then we would have to spread it, but we are accessing an attribute so we get a value back

src/MyHealth/symptoms.ts
aledustet: Sorry, I think I walked ourselves into a corner trying to simplify the data model we store so we can keep it separated as the feature asks of the data. Maybe we can re-shift how the two data models are stored and this can become a thing of the past?

src/MyHealth/symptoms.ts
devinjameson: Very cool! Thanks for the explanation.

src/MyHealth/symptoms.ts
devinjameson: I can't think of a better way to do it 🤷

Unless maybe we could keep the data structures separate, not combine them in the data model, and then handle the "serialization" at the view layer. We could map over an array of the Posix times for the beginnings of the last 14 days and check if we have a daily check in or symptom log for that day and display it…

src/MyHealth/symptoms.ts
```diff
@@ -22,3 +45,81 @@ export const determineHealthAssessment = (
...
+  return dateLeft < dateRight ? -1 : 1
+}
+const compareDatesDescending = (
+  { date: dateLeft }: WithADate,
+  { date: dateRight }: WithADate,
+) => {
+  return dateLeft < dateRight ? 1 : -1
+}
+
+const groupLogEntriesDaily = (allEntries: SymptomLogEntry[]): LogDataByDay => {
```
johnschoeman, line 63:
```suggestion
const groupLogEntriesByDay = (allEntries: SymptomLogEntry[]): LogDataByDay => {
```

src/MyHealth/symptoms.ts
```diff
@@ -22,3 +45,81 @@ export const determineHealthAssessment = (
...
+      }
+    }
+  })
+  return logData
+}
+
+export const serializeDailyLogData = (
+  symptomLogEntries: SymptomLogEntry[],
+  dailyCheckIns: DailyCheckIn[],
+): DayLogData[] => {
```
johnschoeman, line 109: Having a mix of Log / Entry / Data is a bit confusing
What do you think about

```suggestion
export const combineEntryData = (
  symptomLogEntries: SymptomLogEntry[],
  checkInEntries: CheckInEntry[],
): DailyEntry[] => {
```

or even

```suggestion
export const combineSymptomAndCheckInLogs = (
  symptomLogs: SymptomLog[],
  checkInLogs: CheckInLog[],
): DailyLog[] => {
```

?

src/MyHealth/symptoms.ts
johnschoeman: Plus one to combining the `CheckInLogs` and the `SymptomLogs` in the view layer, this seems like a view specific concern.

johnschoeman (approved):
This looks great! Well done on making something very complicated very easy to follow in code.

src/MyHealth/symptoms.ts
aledustet: I agree that we could make the serialization closer to the view but in this case then we would need a context that holds logic for:
- updating log entries
- adding log entries

another one for:
- creating a check in
- creating a log entry

and have them both consume a context with all the log entries and check ins. Could this be second round improvement while we complete the functionality?

src/MyHealth/symptoms.ts
aledustet: 💯 behind the renaming, gonna merge this and start making those changes

aledustet:
Thanks all for the feedback, we're adding a new step from this and will merge here and take on the other one.

PR #433, opened 2020-09-23
Dismissible Symptom Input Form

mattThousand:
### Why
The designs (below) do not specify a way to close the symptom input form if the user decides that they do not wish to submit the form after opening the modal.
Screen Shot 2020-09-22 at 9 26 12 PM

### This Commit
This commit adds a "dismiss" button to the symptom entry form:
Sep-22-2020 21-26-49


devinjameson (approved):
Code looks good! What do you think of using the same style as we do on the `PersonalPrivacy` and `ProximityTracingInfo` screens, where the header is in the same visual container as the back button?

https://github.com/Path-Check/gaen-mobile/blob/develop/src/modals/ProtectPrivacy.tsx

devinjameson:
You may be able to merge this into `develop` @mattThousand.

johnschoeman (approved):
Code looks good! would also plus one merging this to develop rather than another branch.

PR #434, opened 2020-09-23
Fix bug on fix proximity tracing event

johnschoeman:
Why:
Currently there is a bug in iOS 13.7 when a user tries to tap the 'Fix
Proximity Tracing' link on the home screen, where nothing will happen.
This was caused by the NatvieModule.requestAuthorization function
throwing and not being handled.

We would like to try to update the authorization immediately via the
gaen api, and if this fails, we would like to indicate to the user that
the need to update permissions manually via their settings app and link
them to the settings app.

This commit:
Wraps the call to request authorization in a try catch in the component
and shows the Alert dialog with settings link if the request throws.
Additionally we added a logger to the native module function to capture
any failure events in the future. We did not fix the underlying issue of
the `manager.setExposureNotificationEnabled` throwing in the iOS layer.

### DIFF
--- a/src/Home/ProximityTracingActivationStatus.tsx
+++ b/src/Home/ProximityTracingActivationStatus.tsx
@@ -1,5 +1,5 @@
 import React, { FunctionComponent } from "react"
-import { Alert } from "react-native"
+import { Alert, Platform } from "react-native"
 import { useNavigation } from "@react-navigation/native"
 import { useTranslation } from "react-i18next"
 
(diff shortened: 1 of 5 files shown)


src/Home/ProximityTracingActivationStatus.tsx
```diff
@@ -39,9 +46,16 @@ export const ProximityTracingActivationStatus: FunctionComponent = () => {
...
+  const handleOnPressFix = async () => {
+    try {
+      await exposureNotifications.request()
+
+      const isNotENEnabled = status !== ENPermissionStatus.ENABLED
+
+      if (isNotENEnabled) {
+        showNotAuthorizedAlert()
+      }
+    } catch (error) {
```
devinjameson: Why is it not sad that `error` is unused here?

src/Home/ProximityTracingActivationStatus.tsx
```diff
@@ -39,9 +46,16 @@ export const ProximityTracingActivationStatus: FunctionComponent = () => {
...
-  const handleOnPressFix = () => {
-    exposureNotifications.request()
-    if (isNotAuthorized) {
+  const handleOnPressFix = async () => {
+    try {
+      await exposureNotifications.request()
+
+      const isNotENEnabled = status !== ENPermissionStatus.ENABLED
+
+      if (isNotENEnabled) {
```
devinjameson: This is a rare case where I actually find not naming to be more clear:

```suggestion
      if (status !== ENPermissionStatus.ENABLED) {
```

Maybe it would be optimal if we renamed `status` to `enStatus` or something. 🤔

devinjameson (approved):
🚢

src/Home/ProximityTracingActivationStatus.tsx
johnschoeman: in short, it's not a language feature.

Here's any issue that touches a little on the why catches dont have type annotations: https://github.com/Microsoft/TypeScript/issues/20024

johnschoeman: removing the unused error parameter altogether though!

PR #437, opened 2020-09-23
Upgrade typescript 4.0

johnschoeman:
Why:
We would like to be using the current versions of our developer tooling
as a matter of best practice.

This commit:
Upgrades typescript to 4.0.3 and updates all of the related tooling for
eslint-typescript which caused a need to update various globals
defined in the types for react native and react.


.eslintrc.js
```diff
@@ -1,111 +1,116 @@
...
-      files: ['*.ts', '*.tsx'],
+      files: ["*.ts", "*.tsx"],
       rules: {
-        '@typescript-eslint/explicit-module-boundary-types': ['error'],
+        "@typescript-eslint/explicit-module-boundary-types": ["error"],
       },
     },
   ],
-};
+  globals: {
```
johnschoeman, line 111: This is the only added config in this file.

mattThousand (approved):
👏 🥇

src/SelfAssessment/index.js
```diff
@@ -76,6 +76,7 @@ const Assessment = ({ navigation }) => {
     // to support dynamic questions
 
     // memoize assessment question
+    /* eslint-disable react/display-name */
```
aledustet: I believe this comment could be avoided if we do not use an anonymous function, perhaps doing:
```
function assesmentQuestionMemoized({ navigation, route }) {
  return {...}
}
```
Or extracting the function into a named component might fix this?

src/components/GlobalText.tsx
```diff
@@ -7,15 +7,15 @@ interface GlobalTextProps {
   style?: TextStyle
   testID?: string
   onPress?: () => void
-  children: JSX.Element | string
+  children: React.ReactNode | string
```
aledustet: very low priority but how do we feel about importing node from `React` as we are with the `FunctionComponent` ?

aledustet (approved):
👏👏 👏 👏  🔥🔥🔥🔥🔥

src/SelfAssessment/index.js
johnschoeman: this file go removed anyways, so no longer an issue!

PR #435, opened 2020-09-23
Create-Retrieve check in entries from native layer

aledustet:
Why:
----

One of the two data models that will drive the my health screen are check-ins. These should be made once per day and will hold just a check-in status of `Feeling well` or `Feeling not well`. As a user I should be able to go to the My Health screen, add my check-in for the day and see it reflected both on th `Today` tab and the `OverTime` tab. This data should persist from app session to session.

This Commit:
----

- Display todays check in status on today view
- Add todays check in from Today screen
- Update ios native module for checkin/symptom entry changes.
- Store and retrieve check-in entries from the native side
- Rename methods on symptoms helper

Caveat:
----

The data collection and creation only works on iOS at the moment since the updates to the native layer on android are pending.

Co-authored-by: Matt Buckley <matt.buckley@rocketinsights.com>
Co-authored-by: Devin Jameson <devin@thoughtbot.com>

save-check-in

PR #439, opened 2020-09-24
(Android) Expose module to save symptom logs

juancruzgs:
#### Description:

Adds `SymptomLogEntryModule` and changes the field types of existing `CheckIns`.
It replicates the changes made on iOS in this PR https://github.com/Path-Check/gaen-mobile/pull/435

### DIFF
--- a/android/app/src/main/java/org/pathcheck/covidsafepaths/bridge/ExposureNotificationsPackage.java
+++ b/android/app/src/main/java/org/pathcheck/covidsafepaths/bridge/ExposureNotificationsPackage.java
@@ -14,6 +14,7 @@ import org.pathcheck.covidsafepaths.exposurenotifications.reactmodules.DeviceInf
 import org.pathcheck.covidsafepaths.exposurenotifications.reactmodules.ExposureHistoryModule;
 import org.pathcheck.covidsafepaths.exposurenotifications.reactmodules.ExposureKeyModule;
 import org.pathcheck.covidsafepaths.exposurenotifications.reactmodules.ExposureNotificationsModule;
+import org.pathcheck.covidsafepaths.exposurenotifications.reactmodules.SymptomLogEntryModule;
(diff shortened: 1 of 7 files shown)


aledustet (approved):
🚢 :shipit: 

We are gonna need a delete functionality for the SYmptom Log entries, if you think is better to add it here go for it but it's ok if you want to merge this and add a new PR. Thanks!

juancruzgs:
@aledustet I just added it here https://github.com/Path-Check/gaen-mobile/pull/439/commits/e3ef8007d21825736bddada6817bf1d1b9c29b20, let me know if that works for you

src/gaen/nativeModule.ts
```diff
@@ -272,3 +272,7 @@ export const modifyLogEntry = (entry: SymptomLogEntry): Promise<void> => {
 export const getLogEntries = (): Promise<SymptomLogEntry[]> => {
   return symptomLogEntryModule.getSymptomLogEntries()
 }
+
+export const deleteLogEntry = (entry: SymptomLogEntry): Promise<void> => {
```
aledustet: I think we made it receive only the id as a string 😬

aledustet (approved):
Looks good, but the signature of the native module and the methods is different, on one we are sending the object and on another one the id. We are using the id on the IOS implementation.

src/gaen/nativeModule.ts
```diff
@@ -272,3 +272,7 @@ export const modifyLogEntry = (entry: SymptomLogEntry): Promise<void> => {
...
+  return symptomLogEntryModule.deleteSymptomLogEntry(entry.id)
```
juancruzgs: @aledustet It sends the ID to the native layer, but I can also use the ID in the JS signature if that's easier for you

aledustet: No problem with that! I believe we added the method with the right signature so you might be even able to remove it altogether 😄

PR #441, opened 2020-09-24
Enable editing of symptom logs

devinjameson:
Why:
----
As a user, I would like to be able to edit symptom logs.

This commit:
----
- Enable editing of symptom logs in the `OverTime` view
- Add specs for editing symptom logs

gif


src/MyHealth/OverTime.spec.tsx
```diff
@@ -87,4 +89,46 @@ describe("OverTime", () => {
       expect(getByText(lossOfSmellSymptom)).toBeDefined()
     })
   })
+
+  describe("when the user has symptom logs", () => {
+    it("cant edit a symptom log", () => {
```
johnschoeman:
```suggestion
    it("allows the user to edit a symptom log", () => {
```

android/app/build.gradle
```diff
@@ -148,7 +148,7 @@ android {
     }
 
     defaultConfig {
-        applicationId "org.pathcheck.covidsafepaths.bt"
+        applicationId "org.pathcheck.la.bt"
```
johnschoeman: Lets not include these build specific changes.

src/MyHealth/SelectSymptoms.tsx
```diff
@@ -29,13 +29,22 @@ import {
...
   useStatusBarEffect("dark-content", Colors.primaryLightBackground)
   const { t } = useTranslation()
   const navigation = useNavigation()
-  const [selectedSymptoms, setSelectedSymptoms] = useState<string[]>([])
-  const { addLogEntry } = useSymptomLogContext()
+  const route = useRoute<RouteProp<MyHealthStackParams, "SelectSymptoms">>()
+  const { addLogEntry, updateLogEntry } = useSymptomLogContext()
+
+  const { logEntry } = route.params
+  const symptomLogEntryToEdit = logEntry ? JSON.parse(logEntry) : null
```
johnschoeman, line 42: There's a few nulls floating around in this component now. I wonder if it makes sense to separate the concepts of `new` and `edit` a bit here? for example should we have a two components, `CreateSymptomsLogForm` and `EditSymptomsForm` ?

or maybe the form want so accept props of `symptoms: Symptom[]` and and `onSubmit: (symptoms: Symptom[]) => void` where onSubmit is a createOrUpdateById equivalent or something and it defaults to an empty list?

Not 100% sure, but feels like finding a way to pull the nulls up the flow a bit will lead to less complexity in the form.

johnschoeman (approved):
A few thoughts on the form <> data handling, but might make sense to do in a follow up pr or to chat about irl.

src/MyHealth/SelectSymptoms.tsx
devinjameson: How do you feel about handling that in a refactor once this and the delete functionality are merged?

android/app/build.gradle
devinjameson: Not sure how those got there 👺

PR #443, opened 2020-09-24
Enable deletion of symptom logs

devinjameson:
Why:
----
As a user, I would like to be able to delete symptom log entries.

This commit:
----
- Enable user to delete symptom logs
- Updates some styling

gif


src/MyHealth/SelectSymptoms.spec.tsx
```diff
@@ -56,9 +59,61 @@ describe("SelectSymptomsScreen", () => {
         )
       })
     })
+    it("deletes the symptom log", async () => {
```
johnschoeman:
```suggestion
   describe("when the user taps 'Delete', () => {
    it("deletes the symptom log", async () => {
```

src/components/Button.tsx
```diff
@@ -25,6 +24,7 @@ interface ButtonProps {
   customTextStyle?: TextStyle
   testID?: string
   hasRightArrow?: boolean
+  hasPlusIcon?: boolean
```
johnschoeman, line 27: I see why you are doing this, but it makes me a bit sad.

johnschoeman (approved):
LGTM!

src/components/Button.tsx
devinjameson: What alternative would you suggest?

johnschoeman: dont use a common button component. But being a bit glib there, there is not a good general button solution that works in all cases in RN as far as i know and what you have here works!

this would belong in a future commit, but you could allow the button to accept children which could include an icon. Or perhaps you have `leftIcon` and `rightIcon` props that can be whatever. or continue with the pattern you are establishing here. I think all the patterns have their pros and cons, tbh. My main sadness is from the accumulated complexity that all projects tend to have.

devinjameson: Yeah I get that! Let's pair on cleaning this up soon.

PR #444, opened 2020-09-24
Update Matomo react native package

amandabeiner:
Why:
As a dev team, we want to use a Matomo package that works on later
versions of React Native and Swift.

This commit:
Changes our matomo package from 'react-native-matomo' to
'react-native-matomo-sdk' and uninstalls the former.

Co-Authored-By: aledustet <alejandro@thoughtbot.com>


aledustet (approved):
:shipit:

johnschoeman (approved):

PR #445, opened 2020-09-24
Improve analytics opt-in behavior

amandabeiner:
Why:
As a user visiting the settings screen, I want to be able to learn more
about sharing anonymized data.

Currently, if a user has _not_ yet opted into data sharing, clicking
the toggle redirects them to a new screen to turn it on. This is an
unexpected mobile UX pattern, since I would expect toggling a switch to
change the state immediately. When a user _has_ opted into analytics,
the toggle behaves as expected.

This commit:
Updates the Share Anonymized Data List Item to use the same UX pattern
we are currently using on the Home screen. We show the user's current
state with an X or checkmark icon. Clicking on the line item gives them
information about how we use their anonymized data and allows them to
change their status.

Co-Authored-By: johnschoeman <johnschoeman1617@gmail.com>

Sep-24-2020 17-28-33


src/Settings/ShareAnonymizedDataListItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { useNavigation } from "@react-navigation/native"
+import { useTranslation } from "react-i18next"
+
+import { useAnalyticsContext } from "../AnalyticsContext"
+import { GlobalText } from "../components"
+import { ModalScreens, Stacks } from "../navigation"
+import { Icons } from "../assets"
+import { Colors, Iconography, Typography, Spacing } from "../styles"
+
+const ShareAnonymizedDataSwitch: FunctionComponent = () => {
```
johnschoeman:
```suggestion
const ShareAnonymizedDataListItem: FunctionComponent = () => {
```

johnschoeman (approved):
LGTM

src/Settings/ShareAnonymizedDataListItem.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+        userConsentedToAnalytics: false,
+      })
+
+      const { getByTestId } = render(
+        <AnalyticsContext.Provider value={context}>
+          <ShareAnonymizedDataListItem />
+        </AnalyticsContext.Provider>,
+      )
+
+      const activeStateIcon = getByTestId("not-sharing-data")
```
devinjameson: Does this want to be `inactiveStateIcon`?

src/Settings/ShareAnonymizedDataListItem.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const { getByText } = render(
+      <AnalyticsContext.Provider value={context}>
+        <ShareAnonymizedDataListItem />
+      </AnalyticsContext.Provider>,
+    )
+
+    const shareDataListItem = getByText("Share Anonymized Data")
+    fireEvent.press(shareDataListItem)
+    expect(navigationSpy).toHaveBeenCalledWith(Stacks.Modal, {
+      screen: ModalScreens.AnonymizedDataConsent,
```
devinjameson, line 63: 💯

src/modals/AnonymizedDataConsentScreen.tsx
```diff
@@ -26,39 +20,44 @@ const AnonymizedDataConsentScreen: FunctionComponent = () => {
...
+    : t("settings.i_understand_and_consent")
+
   return (
-    <>
-      <StatusBar backgroundColor={Colors.primaryLightBackground} />
-      <ScrollView
-        style={style.container}
-        contentContainerStyle={style.mainContentContainer}
-      >
+    <View style={style.container}>
```
devinjameson, line 45: Does this need a `ScrollView`?

src/modals/AnonymizedDataConsentScreen.tsx
```diff
@@ -17,7 +11,7 @@ import { Stacks } from "../navigation"
 import { Button, GlobalText } from "../components"
 import { Colors, Typography, Spacing, Iconography } from "../styles"
 import { Icons } from "../assets"
-import ExternalLink from "./ExternalLink"
+import ExternalLink from "../Settings/ExternalLink"
 
 const AnonymizedDataConsentScreen: FunctionComponent = () => {
```
devinjameson, line 16: Does this need a `useStatusBarEffect`?

devinjameson (approved):
LGTM, few comments!

src/modals/AnonymizedDataConsentScreen.tsx
amandabeiner: I don't think so! Currently the outer View is not scrollable so that the header (that says "You are sharing data" and has the close button) is fixed. The view that contains the information about data sharing and the CTA button is scrollable

PR #446, opened 2020-09-25
Log publish request body on staging

aledustet:
Why:
----

When new servers are added, often the troubleshooting comes on the staging environment and one item often requested is the body of the request that was sent. We need a way to retrieve it for diagnosis purposes on the staging environment.

This Commit:
----

- Log body from the publish keys request


johnschoeman (approved):
LGTM

PR #449, opened 2020-09-25
Empty state for symptom checker

mattThousand:
### Why
We need an empty state for when the user does not have any symptom logs

### This Commit
This commit implements an empty state for when the user does not have any symptom logs


johnschoeman (approved):

devinjameson (approved):
🚢

PR #451, opened 2020-09-25
Handle failures when operating with symptom logs

aledustet:
Why:
----

We need to be able to recover from errors operating on symptom logs or check ins.

This Commit:
----
- Handle failures on log context operations
- Show flash error message when adding symptoms fails
- Show flash error message when deleting log symptoms fails
- Show flash error message when updating log symptoms fails

### DIFF
--- a/src/MyHealth/SymptomLogContext.tsx
+++ b/src/MyHealth/SymptomLogContext.tsx
@@ -23,30 +23,35 @@ import {
   deleteLogEntry as removeLogEntry,
 } from "../gaen/nativeModule"
 import { isToday } from "../utils/dateTime"
+import Logger from "../logger"
+
+type OperationResult = "success" | "failure"
+type OperationResponse = { kind: OperationResult }
+const SUCCESS_RESPONSE: OperationResponse = { kind: "success" }
 
 export type SymptomLogState = {
   dailyLogData: DayLogData[]
-  addLogEntry: (symptoms: Symptom[]) => Promise<void>
-  updateLogEntry: (entry: SymptomLogEntry) => Promise<void>
-  deleteLogEntry: (symptomLogEntryId: string) => Promise<void>
+  addLogEntry: (symptoms: Symptom[]) => Promise<OperationResponse>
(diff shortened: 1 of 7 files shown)


src/MyHealth/SymptomLogContext.tsx
```diff
@@ -84,31 +89,56 @@ export const SymptomLogProvider: FunctionComponent = ({ children }) => {
...
 
   const updateLogEntry = async (updatedLogEntry: SymptomLogEntry) => {
-    await modifyLogEntry(updatedLogEntry)
-    await fetchLogEntries()
+    try {
+      await modifyLogEntry(updatedLogEntry)
+      await fetchLogEntries()
+      return SUCCESS_RESPONSE
+    } catch (e) {
+      return failureResponse(e.message)
```
johnschoeman, line 117: What do you think about added `Logger.error()`'s for these operations?  Maybe it makes sense to do this in the native module?

johnschoeman (approved):
Nice!

src/MyHealth/SymptomLogContext.tsx
aledustet: I believe we are logging the errors on the `failureResponse` method. Would it be better to move it higher to the `nativeModule`?

johnschoeman: ah, no, just misread the code. imo, its fine where it is at.

src/MyHealth/Today.spec.tsx
```diff
@@ -55,6 +57,34 @@ describe("Today", () => {
...
+          <Today />
+        </SymptomLogContext.Provider>,
+      )
+
+      fireEvent.press(getByLabelText("Good"))
+
+      await waitFor(() => {
+        expect(showMessageSpy).toHaveBeenCalledWith(
+          expect.objectContaining({
+            message: "Sorry, we could not add todays check in",
```
devinjameson, line 82: @aledustet Does this need to be translated? Also I think it wants to be `add today's check in` with an apostrophe in `today`.

PR #453, opened 2020-09-25
Remove report an issue module

johnschoeman:
Why:
Currently we have an old feature in the code base for allowing a user to
report an issue. It has become clear after user testing that we will
never want to be directly collecting user issues from within the app.

This commit:
Removes the report an issue feature.


aledustet (approved):
🚢  :shipit:

PR #456, opened 2020-09-25
Place symptoms on a union type

aledustet:
Why:
----

We need to ensure type safety with the data on the UI side, in case a
new symptom is added or removed.

This Commit:
----

- Create symptoms union type
- Move translations to select symptoms form

Co-authored-by: John Schoeman<johnschoeman1617@gmail.com>


johnschoeman (approved):
💯

devinjameson (commented):
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #457, opened 2020-09-25
Refresh exposure history once it's focused

juancruzgs:
#### Description:

We were only refreshing the exposure history once the user minimized the app.
This PR updates the exposure history screen every time that tab gets focus.

#### Linked issues:

https://pathcheck.atlassian.net/browse/GAEN-252

### DIFF
--- a/src/ExposureContext.tsx
+++ b/src/ExposureContext.tsx
@@ -80,8 +80,8 @@ const ExposureProvider: FunctionComponent = ({ children }) => {
   }, [getLastDetectionDate])
 
   const refreshExposureInfo = async () => {
-    const exposureInfos = await getCurrentExposures()
-    setExposureInfo(exposureInfos)
+    const exposureInfo = await getCurrentExposures()
+    setExposureInfo(exposureInfo)
 
     const detectionDate = await getLastDetectionDate()
     setLastExposureDetectionDate(detectionDate)
--- a/src/ExposureContext.tsx
+++ b/src/ExposureContext.tsx
@@ -33,7 +33,7 @@ export interface ExposureState {
   getRevisionToken: () => Promise<string>
   lastExposureDetectionDate: Posix | null
   storeRevisionToken: (revisionToken: string) => Promise<void>
-  refreshExposureInfo: () => Promise<void>
+  refreshExposureInfo: () => void
   checkForNewExposures: () => Promise<OperationResponse>
 }
 
@@ -52,9 +52,7 @@ const initialState = {
   storeRevisionToken: () => {
     return Promise.resolve()
   },
-  refreshExposureInfo: () => {
-    return Promise.resolve()
-  },
+  refreshExposureInfo: () => {},
   checkForNewExposures: () => {
     return Promise.resolve(SUCCESS_RESPONSE)
   },
@@ -87,8 +85,6 @@ const ExposureProvider: FunctionComponent = ({ children }) => {
 
     const detectionDate = await getLastDetectionDate()
     setLastExposureDetectionDate(detectionDate)
-
-    return Promise.resolve()
   }
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


src/ExposureContext.tsx
```diff
@@ -68,20 +71,15 @@ const ExposureProvider: FunctionComponent = ({ children }) => {
...
 
-    handleAppStateChange()
+    const detectionDate = await getLastDetectionDate()
+    setLastExposureDetectionDate(detectionDate)
 
-    return () => {
-      AppState.removeEventListener("change", handleAppStateChange)
-    }
-  }, [])
+    return Promise.resolve()
```
aledustet: I don't think we need to return a response here, doesn't the `async` `await` calms the type signature?

src/ExposureHistory/index.tsx
```diff
@@ -1,16 +1,30 @@
...
+    lastExposureDetectionDate,
+    exposureInfo,
+    refreshExposureInfo,
+  } = useExposureContext()
+
+  const navigation = useNavigation()
+
+  useEffect(() => {
+    // Refresh exposure info each time this screen gets focus
+    navigation.addListener("focus", () => {
```
aledustet: Nice one! I believe that we should return the result of that `addListener` inside a callback to avoid leaving unhandled promises laying around. We could return the same function that is the return value of the `addListener`:

``` typescript
{...}

useEffect(() => {
  const unsubscribe = navigation.addListener("focus", () => {
    refreshExposureInfo()
  })
  return unsubscribe
})

{...}
```

How do you feel about using the hook that `react-navigation` brings with it? That way we don't have to add this wider callback? Something like:

```typescript
import React, { FunctionCOmponent, useCallback } from "react"
import { useNavigation, useFocusEffect } from "@react-navigation/native"
{...}
  useFocusEffect(
    useCallback(() => {
      refreshExposureInfo()
    }, [refreshExposureInfo])
  );
{...}
```

aledustet: Not sure that would do but maybe something in that note? Maybe the docs on that hook help us more https://reactnavigation.org/docs/function-after-focusing-screen/

aledustet (commented):
Looks good, I think it is worth taking a look at either returning the subscription function when we add the listener or using the `useFocusEffect` from `react-navigation` 💯  your call. Great work!

aledustet (approved):
Ooops forgot to move the radio 😅

src/ExposureContext.tsx
juancruzgs: I removed it, thanks!

src/ExposureHistory/index.tsx
juancruzgs: Done!

src/ExposureContext.tsx
```diff
@@ -68,20 +69,13 @@ const ExposureProvider: FunctionComponent = ({ children }) => {
...
 
-  useEffect(() => {
-    const handleAppStateChange = async () => {
-      const exposureInfos = await getCurrentExposures()
-      setExposureInfo(exposureInfos)
-    }
-
-    AppState.addEventListener("change", handleAppStateChange)
+  const refreshExposureInfo = async () => {
+    const exposureInfos = await getCurrentExposures()
```
devinjameson:
```suggestion
    const exposureInfo = await getCurrentExposures()
     setExposureInfo(exposureInfo)
```

src/ExposureHistory/index.tsx
```diff
@@ -1,16 +1,27 @@
-import React, { FunctionComponent } from "react"
+import React, { FunctionComponent, useCallback } from "react"
 
 import { useExposureContext } from "../ExposureContext"
 import History from "./History"
 
 import { ExposureInfo, ExposureDatum } from "../exposure"
+import { useFocusEffect } from "@react-navigation/native"
```
devinjameson: Thoughts on moving this import to line 2 for consistency with the other files?

devinjameson (approved):
🎉

src/ExposureHistory/index.tsx
juancruzgs: Done!

src/ExposureContext.tsx
juancruzgs: Done!

PR #458, opened 2020-09-26
Manually Refresh Exposures

mattThousand:
### Why
We'd like users to be able to manually check for new exposures

### This Commit
This commit adds a "refresh" button to the exposure history screen, which calls through to the native layer and checks the key server for new archives to process.
Sep-26-2020 18-37-30


src/ExposureHistory/History/DateInfoHeader.spec.tsx
```diff
@@ -18,6 +24,36 @@ describe("DateInfoHeader", () => {
     })
   })
 
+  it('displays the "refresh" button', async () => {
```
johnschoeman: What do you think about removing this spec? I think it is fully covered by the spec below, so it a bit redundant.

src/ExposureHistory/History/DateInfoHeader.tsx
```diff
@@ -17,6 +21,8 @@ const DateInfoHeader: FunctionComponent<DateInfoHeaderProps> = ({
   lastDetectionDate,
 }) => {
   const { t } = useTranslation()
+  const { checkForNewExposures } = useExposureContext()
+  const [refreshing, setRefreshing] = useState(false)
```
johnschoeman, line 31: Given that checking exposures is on the order of milliseconds, what do you think about not showing a loading state?  I assume the loading indicator would always just flash, so theres no really need to give visual feedback to the user, and it would be a bit less code and a bit smoother of a user experience to not have the loading state.

src/OperationResponse.ts
```diff
@@ -0,0 +1,9 @@
+import Logger from "./logger"
```
johnschoeman, line 1: This is totally a judgement call, but I am less of a fan of pulling out helpers for situations like this, as every instance of a use case often wants to grow in different directions, so it tends to turn into spaghetti-ish code over time.  What do you think about inlining this and keeping the  typing scoped to the feature?

johnschoeman (approved):
Nice work! LGTM

src/ExposureHistory/History/DateInfoHeader.tsx
mattThousand: hm, I might push to leave this in place; if the user has a poor network connection and there are a handful of key archives to pull down, it could conceivably take a few seconds. You're quite right, more often than not it will be nearly instantaneous. We can add an animation to improver the user experience a tad: 
Sep-27-2020 11-44-03

src/ExposureHistory/History/DateInfoHeader.tsx
```diff
@@ -29,13 +41,75 @@ const DateInfoHeader: FunctionComponent<DateInfoHeaderProps> = ({
...
+          rotate: rotation.interpolate({
+            inputRange: [0, 1],
+            outputRange: ["0deg", "360deg"],
+          }),
+        },
+      ],
+    },
+  }
+
+  Animated.loop(
```
mattThousand, line 65: @johnschoeman let me know what you think of this animation, totally ok removing it if you think adds unnecessary complication 🧐

src/gaen/nativeModule.ts
```diff
@@ -101,6 +101,10 @@ export const fetchLastExposureDetectionDate = async (): Promise<Posix | null> =>
   }
 }
 
+export const checkForNewExposures = async (): Promise<void> => {
+  return await exposureHistoryModule.detectExposures()
```
juancruzgs, line 105: The method `DebugMenuModule#detectExposuresNow` is no longer needed, right?
i can remove it from the UI if so

src/OperationResponse.ts
aledustet: I agree with john here, these tend to drive toward the too generic side when they are used across the board and maybe they do so in detriment of the actual usefulness they can get.

src/ExposureHistory/History/DateInfoHeader.tsx
johnschoeman: Ah, that makes total sense to me. If its possible to take more than an instant we definitely want the loading state.

src/ExposureHistory/History/DateInfoHeader.tsx
johnschoeman: Its a neat feature, I would say its a nice to have, but you've already coded it, so we should ship it!

and while it does add complexity, it is self contained so not the type of complexity I would be worried about leaking around the system.

PR #459, opened 2020-09-28
Overtime "List" vs "Calendar" Header

mattThousand:
### Why 
We'd like the user to be able to toggle back and forth between list and calendar views on the `Overtime` screen

### This Commit
This commit implements toggling back and forth between list and calendar views on the `Overtime` screen

Sep-28-2020 10-36-05

### DIFF
--- a/src/MyHealth/OverTime.tsx
+++ b/src/MyHealth/OverTime.tsx
@@ -151,47 +151,40 @@ const OverTime: FunctionComponent = () => {
     CALENDAR,
   }
 
-  const [viewSelection, setViewSelection] = useState<ViewSelection>(ViewSelection.LIST)
+  const [viewSelection, setViewSelection] = useState<ViewSelection>(
+    ViewSelection.LIST,
+  )
 
   const noSymptomHistory = dailyLogData.length === 0
 
-  const last14DaysStyle =
-    viewSelection === ViewSelection.LIST ? null : style.hidden
+  const headerTitle =
+    viewSelection === ViewSelection.LIST
+      ? t("symptom_checker.last_14_days")
+      : t("symptom_checker.history")
 
-  const historyStyle =
-    viewSelection === ViewSelection.CALENDAR ? {} : style.hidden
+  const headerIcon =
+    viewSelection === ViewSelection.LIST ? Icons.Calendar : Icons.Hamburger
+
+  const handleOnPressToggleView = () => {
+    if (viewSelection === ViewSelection.LIST) {
+      setViewSelection(ViewSelection.CALENDAR)
+    } else {
+      setViewSelection(ViewSelection.LIST)
+    }
+  }
 
   return (
     <>
-      <View style={[style.headerContainer, last14DaysStyle]}>
-        <GlobalText style={style.headerText}>
-          {t("symptom_checker.last_14_days")}
-        </GlobalText>
+      <View style={style.headerContainer}>
+        <GlobalText style={style.headerText}>{headerTitle}</GlobalText>
         <TouchableOpacity>
           <SvgXml
-            xml={Icons.Calendar}
+            xml={headerIcon}
             width={Iconography.xSmall}
             height={Iconography.xSmall}
-            testID="calendar-button"
-            onPress={() => {
-              setViewSelection(ViewSelection.CALENDAR)
-            }}
-          />
-        </TouchableOpacity>
-      </View>
-      <View style={[style.headerContainer, historyStyle]}>
-        <GlobalText style={style.headerText}>
-          {t("symptom_checker.history")}
-        </GlobalText>
-        <TouchableOpacity>
-          <SvgXml
-            xml={Icons.Hamburger}
-            width={Iconography.xSmall}
-            height={Iconography.xSmall}
-            testID="list-view-button"
+            accessibilityLabel="Toggle symptom log view"
             onPress={handleOnPressToggleView}
-              setViewSelection(ViewSelection.LIST)
-            }}
+            fill={Colors.primary100}
           />
         </TouchableOpacity>
       </View>
@@ -213,8 +206,6 @@ const OverTime: FunctionComponent = () => {
     </>
   )
 }
-
-const headerHeight = 50
 
 const style = StyleSheet.create({
   container: {
@@ -292,7 +283,7 @@ const style = StyleSheet.create({
   },
   headerContainer: {
     width: "100%",
-    height: headerHeight,
+    height: Spacing.xxHuge,
     flexDirection: "row",
     alignItems: "flex-end",
     justifyContent: "space-between",
@@ -303,7 +294,7 @@ const style = StyleSheet.create({
   headerText: {
     ...Typography.header3,
     paddingRight: Spacing.xxLarge,
-    paddingTop: 8,
+    paddingTop: Spacing.xxSmall,
   },
   hidden: {
     display: "none",
--- a/src/MyHealth/OverTime.tsx
+++ b/src/MyHealth/OverTime.tsx
@@ -159,7 +159,7 @@ const OverTime: FunctionComponent = () => {
     viewSelection === ViewSelection.LIST ? null : style.hidden
 
   const historyStyle =
-    viewSelection === ViewSelection.CALENDAR ? null : style.hidden
+    viewSelection === ViewSelection.CALENDAR ? {} : style.hidden
 
   return (
     <>
--- a/src/MyHealth/OverTime.tsx
+++ b/src/MyHealth/OverTime.tsx
@@ -146,30 +146,24 @@ const OverTime: FunctionComponent = () => {
   const { t } = useTranslation()
   const { dailyLogData } = useSymptomLogContext()
 
-  enum ViewSelection {
-    LIST,
-    CALENDAR,
-  }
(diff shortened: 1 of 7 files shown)


johnschoeman (commented):
Code looks good to me! But we are waiting to do the calendar feature until after some of the other work is done, so we should probably not merge this in until we are ready to start the calendar in full.

src/MyHealth/OverTime.tsx
```diff
@@ -144,27 +146,76 @@ const OverTime: FunctionComponent = () => {
   const { t } = useTranslation()
   const { dailyLogData } = useSymptomLogContext()
 
+  enum ViewSelection {
```
devinjameson: What are your thoughts on putting this right below the imports?

src/MyHealth/OverTime.tsx
```diff
@@ -144,27 +146,76 @@ const OverTime: FunctionComponent = () => {
...
+    LIST,
+    CALENDAR,
+  }
+
+  const [viewSelection, setViewSelection] = useState(ViewSelection.LIST)
```
devinjameson:
```suggestion
  const [viewSelection, setViewSelection] = useState<ViewSelection>(ViewSelection.LIST)
```

src/MyHealth/OverTime.tsx
devinjameson: We could also do:

```
type ViewSelection = "List" | "Calendar"
```

Totally doesn't matter, but thought I'd share the option!

src/MyHealth/OverTime.tsx
```diff
@@ -144,27 +146,76 @@ const OverTime: FunctionComponent = () => {
...
+
   const noSymptomHistory = dailyLogData.length === 0
 
+  const last14DaysStyle =
+    viewSelection === ViewSelection.LIST ? null : style.hidden
```
devinjameson:
```suggestion
    viewSelection === ViewSelection.LIST ? {} : style.hidden
```

src/MyHealth/OverTime.tsx
```diff
@@ -144,27 +146,76 @@ const OverTime: FunctionComponent = () => {
...
+
+  const historyStyle =
+    viewSelection === ViewSelection.CALENDAR ? null : style.hidden
```
devinjameson:
```suggestion
    viewSelection === ViewSelection.CALENDAR ? {} : style.hidden
```

src/MyHealth/OverTime.tsx
```diff
@@ -144,27 +146,76 @@ const OverTime: FunctionComponent = () => {
...
+        <GlobalText style={style.headerText}>
+          {t("symptom_checker.history")}
+        </GlobalText>
+        <TouchableOpacity>
+          <SvgXml
+            xml={Icons.Hamburger}
+            width={Iconography.xSmall}
+            height={Iconography.xSmall}
+            testID="list-view-button"
+            onPress={() => {
```
devinjameson:
```suggestion
            onPress={handleOnPressToggleView}
```

devinjameson:
```
const handleOnPressToggleView = () => {
  setViewSelection(!viewSelection)
}
```

src/MyHealth/OverTime.tsx
```diff
@@ -144,27 +146,76 @@ const OverTime: FunctionComponent = () => {
...
+      <View style={[style.headerContainer, historyStyle]}>
+        <GlobalText style={style.headerText}>
+          {t("symptom_checker.history")}
+        </GlobalText>
+        <TouchableOpacity>
+          <SvgXml
+            xml={Icons.Hamburger}
+            width={Iconography.xSmall}
+            height={Iconography.xSmall}
+            testID="list-view-button"
```
devinjameson: I would probably remove this testID and instead give the `TouchableOpacity` an `accessibilityLabel`. Then you can do something like `fireEvent.press(getByLabelText("Toggle symptom log view"))` in your spec.

src/MyHealth/OverTime.tsx
```diff
@@ -144,27 +146,76 @@ const OverTime: FunctionComponent = () => {
...
-    <ScrollView
-      style={style.container}
-      contentContainerStyle={style.contentContainer}
-      alwaysBounceVertical={false}
-    >
-      {noSymptomHistory ? (
-        <GlobalText style={style.noSymptomHistoryText}>
-          {t("symptom_checker.no_symptom_history")}
+    <>
+      <View style={[style.headerContainer, last14DaysStyle]}>
```
devinjameson: Do these have to be two separate views here? Or can we conditionally change the content based on the state of `viewSelection`? Let me know if you'd like to pair on this for a few minutes.

src/MyHealth/OverTime.tsx
```diff
@@ -239,6 +290,24 @@ const style = StyleSheet.create({
     alignSelf: "center",
     ...Typography.body1,
   },
+  headerContainer: {
+    width: "100%",
+    height: headerHeight,
```
devinjameson: I think this can just be set to a `Spacing` var instead of using `headerHeight`, unless you"re using `headerHeight` elsewhere.

src/MyHealth/OverTime.tsx
```diff
@@ -239,6 +290,24 @@ const style = StyleSheet.create({
...
+    flexDirection: "row",
+    alignItems: "flex-end",
+    justifyContent: "space-between",
+    backgroundColor: Colors.white,
```
devinjameson:
```suggestion
    backgroundColor: Colors.primaryLightBackground,
```

src/MyHealth/OverTime.tsx
```diff
@@ -239,6 +290,24 @@ const style = StyleSheet.create({
...
+    paddingHorizontal: Spacing.small,
+    paddingBottom: Spacing.small,
+  },
+  headerText: {
+    ...Typography.header3,
+    paddingRight: Spacing.xxLarge,
+    paddingTop: 8,
```
devinjameson: I would try to use a `Spacing` var here if possible!

src/MyHealth/index.tsx
```diff
@@ -22,7 +22,7 @@ const MyHealthScreen: FunctionComponent = () => {
 
   return (
     <>
-      <StatusBar backgroundColor={Colors.secondary10} />
+      <StatusBar backgroundColor={Colors.white} />
```
devinjameson:
```suggestion
      <StatusBar backgroundColor={Colors.primaryLightBackground} />
```

src/MyHealth/index.tsx
```diff
@@ -54,12 +54,12 @@ const MyHealthScreen: FunctionComponent = () => {
 const style = StyleSheet.create({
   container: {
     flexGrow: 1,
-    backgroundColor: Colors.secondary10,
+    backgroundColor: Colors.white,
```
devinjameson:
```suggestion
    backgroundColor: Colors.primaryLightBackground,
```

src/MyHealth/index.tsx
```diff
@@ -54,12 +54,12 @@ const MyHealthScreen: FunctionComponent = () => {
...
     paddingTop: Spacing.large,
   },
   headerContainer: {
     paddingHorizontal: Spacing.large,
-    backgroundColor: Colors.secondary10,
+    backgroundColor: Colors.white,
```
devinjameson:
```suggestion
    backgroundColor: Colors.primaryLightBackground,
```

src/assets/svgs/calendar.ts
```diff
@@ -0,0 +1,14 @@
+export default `
```
devinjameson, line 1: With `svg` that use one fill color, I generally find it best to:
- Flatten the vector in Figma so that there is no stroke information. We want the vector to have a fill only.
- Remove any `fill` declarations from the svg so we can set the fill color in the `SvgXml` props

src/MyHealth/OverTime.tsx
```diff
@@ -144,27 +146,76 @@ const OverTime: FunctionComponent = () => {
...
+          />
+        </TouchableOpacity>
+      </View>
+      <View style={[style.headerContainer, historyStyle]}>
+        <GlobalText style={style.headerText}>
+          {t("symptom_checker.history")}
+        </GlobalText>
+        <TouchableOpacity>
+          <SvgXml
+            xml={Icons.Hamburger}
```
devinjameson: Thoughts on using the `fill` prop here? This way we can change this fill color easily without having to dig around in the svg.

```
<SvgXml
  fill={Colors.primary100}
  ...
/>
```

devinjameson (commented):
Have some thoughts, hope it's helpful! Happy to discuss.

src/MyHealth/OverTime.tsx
mattThousand: Thanks! Is one syntax more "idiomatic" than the other? I can go either way

devinjameson: I prefer `type` because it's simpler. Then we don't need to do `ViewSelection.LIST`, we just do "List", and we still get type safety.

devinjameson (approved):
🔥  This will be great to merge once we're ready, nice work!

PR #463, opened 2020-09-28
Fix issue with type of object passed to linking prop

devinjameson:


devinjameson:
Closing this because we are fixing it in another PR with a bunch of type fixes.

PR #465, opened 2020-09-28
Update tsconfig for ts 4.0

johnschoeman:
Why:
Currently our tsconfig is set to only run the ts compiler on files that
end with `.d.ts` which means that the complier is not running on any of
our source files. This was caused by an update to the behaviour of the
includes tsconfig option from being inclusive to exclusive.

This commit:
updates the tsconfig file to fix the issue as well as removes no no
longer necessary config options. We updated some type errors that made
their way into the codebase during the few days of not type checking the
entire codebase

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,75 +1,33 @@
 {
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "React Native",
   "compilerOptions": {
-    "skipLibCheck": true,
-    /* Basic Options */
-    // "incremental": true,                   /* Enable incremental compilation */
-    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
-    "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
-    "lib": [
-      "es2019"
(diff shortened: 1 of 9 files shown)


devinjameson (approved):
🎉 :shipit:

tsconfig.json
```diff
@@ -1,75 +1,21 @@
...
   },
-  "include": ["*.d.ts"],
-  "exclude": [
-    "node_modules",
-    "babel.config.js",
-    "metro.config.js",
-    "jest/*",
-    "**/*.svg"
-  ]
+  "include": ["src/**/*", "App.tsx", "**/*.d.ts"],
```
aledustet: curious why the root folder is not included?

aledustet (approved):

tsconfig.json
johnschoeman: I suppose because src/ and App.tsx are the only ts files? but no reason to not have it be the root, we will update!

tsconfig.json
```diff
@@ -1,75 +1,33 @@
...
   },
-  "include": ["*.d.ts"],
-  "exclude": [
-    "node_modules",
-    "babel.config.js",
-    "metro.config.js",
-    "jest/*",
-    "**/*.svg"
-  ]
+  "include": ["**/*.ts", "**/*.d.ts"],
```
aledustet: should we have the tsx files here too?

devinjameson: I believe it works for `tsx` if we use `ts` but it might be good to add `tsx` to be exhaustive?

johnschoeman: They are captured!

PR #466, opened 2020-09-28
Allow images on how it works flow to be configured

aledustet:
Why:
----

We need to allow for overrides of the images displayed on our application how it works flow.

Caveat:
----

This approach is an attempt to allow for all image overrides. In this first pass, only the HIW flow will be pulled. There are a number of modifications to our scripts that can be generalized once we agree this is a way we want to proceed. Like with the welcome image that is a separate asset in this moment

This Commit:
----
- Rename assets to be closer to where they are used, the names of the images shown on the HIW flow were a reflection of the actual content but for making it easier to override, the names will no reflect where they are used.
- Add setup script to update image assets. This script will download and copy all images inside the zip file provided and overwrite the ones that are provided, making them not all required.

This depends on a configuration PR. To test the script that commit sha can be used.


devinjameson (approved):
🚢 :shipit:🔥 😷

bin/helpers.rb
```diff
@@ -85,3 +87,7 @@ def mobile_resources_commit
 
   commit
 end
+
+def resources_base_url(access_token)
```
johnschoeman, line 91: 💯 nice extract method

johnschoeman (approved):

PR #467, opened 2020-09-28
Add CTA to Self Screener from Exposure Detail

amandabeiner:
Why:
As an LA user who has received an EN, I want to be able to enter the
self screener flow so that I can get personalized information about what
I should do next.

This commit:
-Adds a SelfScreener stack with placeholder text
- Displays a button to navigate to the SelfScreener if it is enabled for
  the HA
- Updates the styling of the ExposureDetail page to match the updated
  styles for this flow.

self screener

### DIFF
--- a/src/ConfigurationContext.tsx
+++ b/src/ConfigurationContext.tsx
@@ -9,6 +9,7 @@ export interface Configuration {
   displayCallbackForm: boolean
   displayReportAnIssue: boolean
   displayMyHealth: boolean
+  displaySelfScreener: boolean
   findATestCenterUrl: string | null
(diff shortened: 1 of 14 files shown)


src/gaen/nativeModule.ts
```diff
@@ -215,6 +215,7 @@ export const detectExposuresNow = async (): Promise<string> => {
 }
 
 export const simulateExposure = async (): Promise<"success"> => {
+  console.log(toExposureInfo(debugModule.simulateExposure()))
```
devinjameson: 😱

devinjameson (approved):
👏  🚢

PR #468, opened 2020-09-28
Enable configuration of brand colors

devinjameson:
Why:
----
We would like for brand colors to be configurable.

This commit:
----
- Abstract primary and secondary colors to a config file
- Add necessary build scripts to download config file from a remote repo


bin/set_ha.sh
```diff
@@ -35,7 +36,12 @@ else
 end
 
 def download_all_assets
-system("./bin/download_assets.sh #{HA_LABEL} #{ACCESS_TOKEN}") && download_copy_file(HA_LABEL, ACCESS_TOKEN) && download_links_file(HA_LABEL, ACCESS_TOKEN) && download_welcome_image(HA_LABEL, ACCESS_TOKEN) && fetch_configurable_images(HA_LABEL, ACCESS_TOKEN)
+  system("./bin/download_assets.sh #{HA_LABEL} #{ACCESS_TOKEN}")
```
johnschoeman, line 39: @aledustet , is there a reason we need these "&&" s?

johnschoeman (approved):

bin/set_ha.sh
aledustet: Those were mostly since some of the scripts were returning false if the operation was not successful, on the new ones I am mostly exiting with a `1` status code when an error occurs so we should be fine there. I'll take a pass at the error statuses on the other scripts to make sure that they halt when an error happens.

PR #469, opened 2020-09-29
Symptom Log Calendar

mattThousand:
### Why 
We'd like users to be able to view their longer-term (4 week) symptom log/checkin history in the form of a calenday

### This Commit 
This commit implements a calendar view containing a 4-week window of the user's checkins. Because data older than 14 days is deleted from the device, dates prior to 14 days from the current date are disabled in the calendar.

Oct-01-2020 11-50-34

### DIFF
--- a/src/MyHealth/Calendar.spec.tsx
+++ b/src/MyHealth/Calendar.spec.tsx
@@ -6,16 +6,22 @@ import { factories } from "../factories"
 
 import Calendar from "./Calendar"
 import { DayLogData } from "./symptoms"
+import dayjs from "dayjs"
 
 afterEach(cleanup)
+const buildLogDataHistory = () => {
+  const logData = factories.dayLogData.build()
+  const totalDays = 30
+  return toLogDataHistory([logData], totalDays)
+}
 
 describe("Calendar", () => {
-  it("renders", () => {
-    const logDataHistory = buildLogDataHistory()
-    const onSelectDate = (_logData: DayLogData) => {}
-    const selectedDay = logDataHistory[0]
+  const logDataHistory = buildLogDataHistory()
+  const onSelectDate = (_logData: DayLogData) => {}
+  const selectedDay = logDataHistory[0]
 
-    const { getByTestId } = render(
+  it("displays 4 weeks of data", () => {
+    const { getByText } = render(
       <Calendar
         logDataHistory={logDataHistory}
         onSelectDate={onSelectDate}
@@ -23,12 +29,28 @@ describe("Calendar", () => {
       />,
     )
 
-    expect(getByTestId("myhealth-history-calendar")).not.toBeNull()
+    const dayNumber = parseInt(dayjs(logDataHistory[0].date).format("D"))
+
+    Array(30)
+      .fill(0)
+      .forEach((_, i) => {
+        if (Math.abs(dayNumber - i) > 3) {
+          expect(getByText(`${i + 1}`)).toBeDefined()
+        }
+      })
+  })
+
+  it("displays the days of the week", () => {
+    const labels = ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"]
+    const { getByText } = render(
+      <Calendar
+        logDataHistory={logDataHistory}
+        onSelectDate={onSelectDate}
+        selectedDay={selectedDay}
+      />,
+    )
+    labels.forEach((label) => {
+      expect(getByText(label)).toBeDefined()
+    })
   })
 })
-
-const buildLogDataHistory = () => {
-  const logData = factories.dayLogData.build()
-  const totalDays = 30
-  return toLogDataHistory([logData], totalDays)
-}
--- a/src/MyHealth/Calendar.spec.tsx
+++ b/src/MyHealth/Calendar.spec.tsx
@@ -1,0 +1,34 @@
+import React from "react"
+import { cleanup, render } from "@testing-library/react-native"
+
+import { toLogDataHistory } from "./logDataHistory"
+import { factories } from "../factories"
+
+import Calendar from "./Calendar"
+import { DayLogData } from "./symptoms"
+
+afterEach(cleanup)
+
+describe("Calendar", () => {
+  it("renders", () => {
+    const logDataHistory = buildLogDataHistory()
+    const onSelectDate = (_logData: DayLogData) => {}
+    const selectedDay = logDataHistory[0]
+
+    const { getByTestId } = render(
+      <Calendar
+        logDataHistory={logDataHistory}
+        onSelectDate={onSelectDate}
+        selectedDay={selectedDay}
+      />,
+    )
+
+    expect(getByTestId("myhealth-history-calendar")).not.toBeNull()
+  })
+})
+
+const buildLogDataHistory = () => {
+  const logData = factories.dayLogData.build()
+  const totalDays = 30
+  return toLogDataHistory([logData], totalDays)
+}
--- a/src/MyHealth/Calendar.tsx
+++ b/src/MyHealth/Calendar.tsx
@@ -1,0 +1,120 @@
+import React, { FunctionComponent } from "react"
+import { Text, TouchableOpacity, View, StyleSheet } from "react-native"
+import dayjs from "dayjs"
+
+import { Spacing, Typography } from "../styles"
+import { GlobalText } from "../components"
+import { DayLogData } from "./symptoms"
+import DayIndicator from "./DayIndicator."
+
+interface CalendarProps {
+  logDataHistory: DayLogData[]
+  onSelectDate: (logData: DayLogData) => void
+  selectedDay: DayLogData | null
+}
+
+const Calendar: FunctionComponent<CalendarProps> = ({
+  logDataHistory,
+  onSelectDate,
+  selectedDay,
+}: CalendarProps) => {
+  const lastMonth = dayjs().subtract(1, "month")
+  const title = `${lastMonth.format("MMMM")} | ${dayjs().format(
+    "MMMM",
+  )}`.toUpperCase()
+
+  const week1 = logDataHistory.slice(0, 7)
+  const week2 = logDataHistory.slice(7, 14)
+  const week3 = logDataHistory.slice(14, 21)
+  const week4 = logDataHistory.slice(21, 28)
+
+  interface CalendarRowProps {
+    week: DayLogData[]
+  }
+
+  const CalendarRow = ({ week }: CalendarRowProps) => {
+    return (
+      <View style={styles.calendarRow}>
+        {week
+          .filter((logData: DayLogData | null) => logData.checkIn)
+          .map((logData: DayLogData) => {
+            const isSelected =
+              logData.checkIn.date === selectedDay?.checkIn.date
+
+            return (
+              <TouchableOpacity
+                key={`calendar-day-${logData.checkIn.date}`}
+                testID={`calendar-day-${logData.checkIn.date}`}
+                onPress={() => onSelectDate(logData)}
+              >
+                <DayIndicator isSelected={isSelected} logData={logData} />
+              </TouchableOpacity>
+            )
+          })}
+      </View>
+    )
+  }
+
+  const DayLabels = () => {
+    const labels = ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"]
+    return (
+      <View style={styles.calendarRow}>
+        {labels.map((label: string, idx: number) => {
+          return (
+            <View style={styles.labelStyle} key={`calendar-label-${idx}`}>
+              <Text style={styles.labelTextStyle}>{label}</Text>
+            </View>
+          )
+        })}
+      </View>
+    )
+  }
+
+  return (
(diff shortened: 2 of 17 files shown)


src/MyHealth/Calendar.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { cleanup, render } from "@testing-library/react-native"
+
+import { toLogDataHistory } from "./logDataHistory"
+import { factories } from "../factories"
+
+import Calendar from "./Calendar"
+import { DayLogData } from "./symptoms"
+import dayjs from "dayjs"
+
+afterEach(cleanup)
```
devinjameson, line 11: Haven’t seen this pattern before in our tests, is this necessary?

src/MyHealth/Calendar.tsx
```diff
@@ -0,0 +1,132 @@
...
+  onSelectDate,
+  selectedDay,
+}: CalendarProps) => {
+  const lastMonth = dayjs().subtract(1, "month")
+  const title = `${lastMonth.format("MMMM")} | ${dayjs().format(
+    "MMMM",
+  )}`.toUpperCase()
+
+  useEffect(() => {
+    if (!selectedDay) {
```
devinjameson, line 28: What’s happening here is a bit unclear to me. How do you feel about naming a function or two and extracting the logic?

src/MyHealth/Calendar.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  interface CalendarRowProps {
+    week: DayLogData[]
+  }
+
+  const CalendarRow = ({ week }: CalendarRowProps) => {
+    return (
+      <View style={styles.calendarRow}>
+        {week
+          .filter((logData: DayLogData | null) => logData.checkIn)
```
devinjameson, line 51: How do you feel about pulling some of these operations out into functions as opposed to keeping all the logic within the view? It may be just me but I find it a bit hard to grok.

src/MyHealth/Calendar.tsx
```diff
@@ -0,0 +1,132 @@
...
+          })}
+      </View>
+    )
+  }
+
+  const DayLabels = () => {
+    const labels = ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"]
+    return (
+      <View style={styles.calendarRow}>
+        {labels.map((label: string, idx: number) => {
```
devinjameson, line 74: I would expect the type of these arguments to be inferred?

src/MyHealth/Calendar.tsx
```diff
@@ -0,0 +1,132 @@
...
+              <Text style={styles.labelTextStyle}>{label}</Text>
+            </View>
+          )
+        })}
+      </View>
+    )
+  }
+
+  return (
+    <View testID={"myhealth-history-calendar"} style={styles.container}>
```
devinjameson, line 86: Super super doesn’t matter but we don’t need curly braces here if we are passing a string!

src/MyHealth/Calendar.tsx
```diff
@@ -0,0 +1,132 @@
...
+    flexDirection: "row",
+    justifyContent: "space-between",
+    alignItems: "center",
+  },
+  labelStyle: {
+    justifyContent: "center",
+    alignItems: "center",
+    width: Spacing.xHuge,
+  },
+  labelTextStyle: {},
```
devinjameson, line 129: Empty boi?

src/MyHealth/DayIndicator.tsx
```diff
@@ -0,0 +1,145 @@
+import React, { FunctionComponent } from "react"
+import { View, Text, ViewStyle, TextStyle, StyleSheet } from "react-native"
+import dayjs from "dayjs"
+
+import { DateTimeUtils } from "../utils"
+import { Outlines, Colors, Typography, Spacing } from "../styles"
```
devinjameson, line 6: Convention for organizing imports is usually:
- External libraries
- Internal imports
- Style or image/icon imports

src/MyHealth/DayIndicator.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const baseStyles: IndicatorStyle = [styles.circleBase, styles.textBase]
+
+  const [[circleStyle, textStyle]] = [baseStyles]
+    .map(applyIsTodayStyle)
+    .map(applyRiskStyle)
+    .map(applyIsSelectedStyle)
+
+  const dayNumber = dayjs(logData.date).format("D")
+
+  const indicator = <Text style={textStyle}>{dayNumber}</Text>
```
devinjameson, line 122: This seems a bit odd to me, why not inline?

src/MyHealth/Calendar.tsx
johnschoeman, line 28: Do we need to autoselect the day for the user? if so, would it make sense to remove the null option for the selected day on the prop?

src/MyHealth/Calendar.tsx
johnschoeman, line 51:
```suggestion
          .filter((logData: DayLogData | null) => logData.checkIn !== undefined)
```

src/MyHealth/DayIndicator.tsx
```diff
@@ -0,0 +1,145 @@
+import React, { FunctionComponent } from "react"
+import { View, Text, ViewStyle, TextStyle, StyleSheet } from "react-native"
+import dayjs from "dayjs"
+
+import { DateTimeUtils } from "../utils"
+import { Outlines, Colors, Typography, Spacing } from "../styles"
+import { CheckInStatus, DayLogData } from "./symptoms"
+
+interface DayIndicatorProps {
```
johnschoeman, line 9:
```suggestion
interface CalendarDayProps {
```
I think we initially used the `Indicator` terminology because there was a naming conflict, so what do you think about `CalendarDay`?

src/MyHealth/DayIndicator.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+type IndicatorStyle = [ViewStyle, TextStyle]
+
+const DayIndicator: FunctionComponent<DayIndicatorProps> = ({
+  logData,
+  isSelected,
+}: DayIndicatorProps) => {
+  const isToday = DateTimeUtils.isToday(logData.date)
+
+  const applyRiskStyle = ([
```
johnschoeman, line 22:
```suggestion
  const applyStyle = ([
```
Similarly, we no longer have a concept of `Risk` on the calendar view for this.

src/MyHealth/DayIndicator.tsx
```diff
@@ -0,0 +1,145 @@
...
+    circleStyle,
+    textStyle,
+  ]: IndicatorStyle): IndicatorStyle => {
+    switch (logData.checkIn.status) {
+      case CheckInStatus.NotCheckedIn: {
+        return [{ ...circleStyle }, { ...textStyle, color: Colors.primaryText }]
+      }
+      case CheckInStatus.TooOld: {
```
johnschoeman, line 30: This CheckInStatus.TooOld seems to be a bit out of place. Was the reason it was included to be able to have a calendar view display a different color for days older that 14? If thats the case then it feels like we are adding complexity to the data model to accommodate a the view layer. What do you think about keeping the data model agnostic of the UI solution by having a view layer only solution to encode this?  something like
```
type CalendarHistory = CalendarDatum[]
interface CalendarDatum = {
   date: Posix
   logData: LogData | null
}
```
then this applied style would be something like:
```
  if (logData.checkIn) {
    return hasDataStyle
  } else {
     return noDataStyle
  }
```
This way if we ever change up the calendar or the data model, the view logic wont be coupled to the data model.

src/MyHealth/DayIndicator.tsx
```diff
@@ -0,0 +1,145 @@
...
+      return [circleStyle, textStyle]
+    }
+  }
+
+  const applyIsSelectedStyle = ([
+    circleStyle,
+    textStyle,
+  ]: IndicatorStyle): IndicatorStyle => {
+    if (isSelected) {
+      switch (logData.checkIn.status) {
```
johnschoeman, line 75: Why do we need to switch on the checkIn.status twice? can we consolidate the conditional logic here?

src/MyHealth/DaySummary.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const dayJsDate = posixToDayjs(date)
+
+  return (
+    <>
+      {dayJsDate && (
+        <GlobalText style={style.dateText}>
+          {dayJsDate.local().format("MMMM D, YYYY")}
+        </GlobalText>
+      )}
+      {checkIn && <CheckInSummary status={checkIn.status} />}
```
johnschoeman, line 134:
```suggestion
      {Boolean(checkIn) && <CheckInSummary status={checkIn.status} />}
```

src/MyHealth/History.spec.tsx
```diff
@@ -0,0 +1,118 @@
...
+jest.mock("@react-navigation/native")
+;(useNavigation as jest.Mock).mockReturnValue({ navigate: jest.fn() })
+
+const defaultDate = dayjs(DateTimeUtils.daysAgo(2)).startOf("day").valueOf()
+const fifteenDaysAgo = dayjs(DateTimeUtils.daysAgo(15)).startOf("day").valueOf()
+
+describe("History", () => {
+  describe("when a day is selected", () => {
+    describe("when a day > 14 days ago is selected", () => {
+      it('displays the "Datstore data older than 14 days" message', () => {
```
johnschoeman, line 23:
```suggestion
      it('displays the "Datastore data older than 14 days" message', () => {
```

src/MyHealth/History.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { Typography, Colors, Outlines, Spacing } from "../styles"
+import Calendar from "./Calendar"
+import { toLogDataHistory } from "./logDataHistory"
+import DaySummary from "./DaySummary"
+import { DayLogData } from "./symptoms"
+
+const History: FunctionComponent = () => {
+  const {
+    dailyLogData,
+    selectLogEntry,
```
johnschoeman, line 14: what do you think about having the `selectLogEntry` and `selectedLogEntry` live on the History component itself? I dont think any other component will need access to these and we probably want to minimize the surface area of state things.

src/MyHealth/OverTime.tsx
```diff
@@ -178,25 +46,28 @@ const OverTime: FunctionComponent = () => {
...
-          <GlobalText style={style.noSymptomHistoryText}>
-            {t("symptom_checker.no_symptom_history")}
-          </GlobalText>
-        ) : (
-          dailyLogData.map((logData) => {
-            return <DaySummary key={logData.date} dayLogData={logData} />
-          })
-        )}
-      </ScrollView>
+      {viewSelection === "List" ? (
```
johnschoeman, line 52:
```suggestion
      {viewSelection === "List" ? (
          <SymptomList />
          ) : (
           <SymptomCalendar />
          )
```

src/MyHealth/logDataHistory.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const toLogDataHistory = (
+  dayLogData: DayLogData[],
+  dayCount: number,
+  now = Date.now(),
+): DayLogData[] => {
+  const calendar = calendarDays(now, dayCount)
+  const dateLogDataMap: Record<Posix, DayLogData> = dayLogData.reduce(
+    (dict, element) => {
+      const startOfDay = dayjs(element.date).startOf("day").valueOf()
```
johnschoeman, line 16: do we need to localize for the timezone here?

src/utils/dateTime.ts
```diff
@@ -36,6 +36,10 @@ export const isInFuture = (date: Posix): boolean => {
   return date > Date.now()
 }
 
+export const isOlderThan14Days = (date: Posix): boolean => {
```
johnschoeman, line 39: I might keep this one colocated with where it is used, i dont think we know if this will need to support multiple callers at this point.

johnschoeman (approved):
Code is looking good.

I would not add the `tooOld` concept to the CheckIn enum directly, as this isnt actually a valid state for a CheckIn but a bit of state we want to manage just for the calendar, which we can do isolated in the calendar view itself. This will keep the complexity in the rest of the system down.

I would also consider making the feature a bit simpler by not allowing the user to tap days that are older than 14 days at all, if they are grayed out i dont think there would be an expectation to be able to tap that day.

johnschoeman:
Also, lets not merge this this week so that we can cut a release candidate build with just the symptom checker

src/MyHealth/History.tsx
```diff
@@ -0,0 +1,119 @@
...
+  } = useSymptomLogContext()
+
+  const logDataHistory = toLogDataHistory(dailyLogData, 30)
+
+  const handleOnSelectDate = (logData: DayLogData) => {
+    selectLogEntry(logData)
+  }
+
+  return (
+    <>
```
devinjameson, line 25: Fragment necessary?

devinjameson (approved):
🔥 Few thoughts!

PR #470, opened 2020-09-29
Remove welcome image downloading script

aledustet:
Why:
----

We need to consolidate the number of scripts we run prior to building the app. The separation of the welcome image from the rest of in-app image assets is not necessary anymore.

This Commit:
----

- Remove the download_welcome_image script and helpers since the welcome image will from now on be included on the configurable_images zip file
- Add diagnosis to the image download script to detect miss-matches during the build process.

Reviewers:
----

This depends on a PR on the configuration repo, the resource commit is updated accordingly for when it is merged.


johnschoeman (approved):
LGTM,
I wonder if we can get away from the `ProtectPrivacy1, 2, 3, 4, 5` if we name based off the content of the screen rather than the position? e.g.
```
ProtectPrivacyTop
ProtectPrivacyExchangeKeys
ProtectPrivacySubmitKeys
ProtectPrivacyReceiveKeys
ProtectPrivacyReceiveNotification
```

probably not worth doing more work on this one here, but for future consideration maybe.

aledustet:
I think we should not wait for these considerations since it requires reworking all of the asset's zip files and docs. I think the time to rename them is now. Will proceed

PR #471, opened 2020-09-29
Update translations and fix typo on HIW copy

aledustet:
Why:
----

We had a type on one of the HIW screens.

Following the elimination of chunks of the code, there were a number of unused translations on lokalise. Since the push does not force the update we need to trim it once in a while.

This Commit:
----

- Ran i18n extract and cleaned unused keys from this process on lokalise
- Pull latest from lokalise

Reviewers:
----

It is a lot that changed but most of the meat is on the `en.json` file, the rest are mostly following thru


johnschoeman (approved):
LGTM

PR #473, opened 2020-09-29
Standardize naming of Screen types

devinjameson:
Why:
----
Prior to this commit, we had some screen types using the pattern
"XScreen" (e.g. "HomeScreen") and others using the pattern
"XStackScreen" (e.g. "HomeStackScreen"). We would like the pattern we
use to be consistent.

This commit:
----
Updates all Screen types to use the pattern "XStackScreen" as in
"HomeStackScreen".

### DIFF
--- a/src/Activation/AcceptTermsOfService.tsx
+++ b/src/Activation/AcceptTermsOfService.tsx
@@ -13,7 +13,7 @@ import { SvgXml } from "react-native-svg"
 
 import { Icons } from "../assets"
 import { GlobalText, Button, GradientBackground } from "../components"
-import { ActivationScreens, useStatusBarEffect } from "../navigation"
+import { ActivationStackScreens, useStatusBarEffect } from "../navigation"
 import { useConfigurationContext } from "../ConfigurationContext"
 
 import {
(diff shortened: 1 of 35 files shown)


johnschoeman (approved):
LGTM, just curious as to why we chose 'XStackScreen' over 'XScreen'?

devinjameson:
> LGTM, just curious as to why we chose 'XStackScreen' over 'XScreen'?

@johnschoeman Few reasons. It seems more clear because we are typing the screens within the stack and not within a screen itself. We also have screens within stacks that use the same name as the stack e.g. `Home` (the actual home screen component) and `HomeScreen` (which was being used as a type for any screen in the home stack). That's a bit confusing because sometimes we use `XStack` to name a component (other times not).

We also had issues with the `HowItWorksScreen` because it was being used as a component, but also the name of the union type for screens in the stack. So all around it seems to make more sense to use `XStackScreen` for clarity and to ensure no overlap between the names of components and types.

johnschoeman:
> @johnschoeman Few reasons. It seems more clear because we are typing the screens within the stack and not within a screen itself. We also have screens within stacks that use the same name as the stack e.g. `Home` (the actual home screen component) and `HomeScreen` (which was being used as a type for any screen in the home stack). That's a bit confusing because sometimes we use `XStack` to name a component (other times not).
> 
> We also had issues with the `HowItWorksScreen` because it was being used as a component, but also the name of the union type for screens in the stack. So all around it seems to make more sense to use `XStackScreen` for clarity and to ensure no overlap between the names of components and types.

Makes a lot of sense to me, thanks for the explanation!

PR #474, opened 2020-09-29
Allow configuration of emergencies phone number

aledustet:
Why:
----

The emergency contact phone number varies from authority to authority and should be allowed to be configured.

This Commit:
----

- Adds the `emergencyPhoneNumber` argument to the configuration context, this one has a default value since all of the current use cases won't require an update

### DIFF
--- a/src/ConfigurationContext.tsx
+++ b/src/ConfigurationContext.tsx
@@ -2,6 +2,8 @@ import React, { FunctionComponent, createContext, useContext } from "react"
 import { Platform } from "react-native"
 import env from "react-native-config"
 
+const USA_EMERGENCY_NUMBER = "911"
+
 export interface Configuration {
   appDownloadLink: string
   appPackageName: string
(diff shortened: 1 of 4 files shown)


src/ConfigurationContext.tsx
```diff
@@ -2,13 +2,16 @@ import React, { FunctionComponent, createContext, useContext } from "react"
 import { Platform } from "react-native"
 import env from "react-native-config"
 
+const USA_EMERGENCY_NUMBER = "911"
```
johnschoeman: What are your thoughts on not providing a default?

We would need to handle the null cases by not rendering the emergency call buttons, but perhaps that's better behavior than defaulting to 911?

johnschoeman (approved):
:shipit:

src/ConfigurationContext.tsx
aledustet: No thoughts other than editing 33 files to add the value 😬  but I agree it breaks with the pattern we have.

devinjameson (approved):
💯 🔥

src/ConfigurationContext.tsx
johnschoeman: 100%, lets not edit the 33 files right now, but lets cut a card to consider for the a future update.

aledustet: No worries, I updated the PR with that consideration and added the values on the resources 👍

PR #475, opened 2020-09-29
Add MIT license to package.json

devinjameson:
Why:
----
We would like to show a license in our package.json. Otherwise, we see a warning when
running terminal commands like `yarn`.

This commit:
----
- Add MIT license to package.json


johnschoeman (approved):
🥇

PR #477, opened 2020-09-29
Update exposure history icons to allow fill

johnschoeman:
Why:
Currently the icons on the exposure history screen do not allow for
updating the fill color from a component.

This commit:
Replaces the svgs with ones that do not have a fill.


aledustet (approved):
Good one, I was gonna do this when we start moving the graphics to be configurable ... btw should we?

johnschoeman:
> Good one, I was gonna do this when we start moving the graphics to be configurable ... btw should we?

I dont think these icons need to be configurable. at least we havent gotten any asks to make them such

PR #478, opened 2020-09-29
(Android) Add Bluetooth/Location listeners

juancruzgs:
#### Description:

We were only reflecting Bluetooth/Location status changes once the app was resumed, missing some updates that happened after the app was resumed.
This PR adds listeners to reflect the status as soon as it changes.

#### Linked issues:

https://pathcheck.atlassian.net/browse/GAEN-259

### DIFF
--- a/src/SystemServicesContext/useLocationPermissions.ts
+++ b/src/SystemServicesContext/useLocationPermissions.ts
@@ -5,6 +5,7 @@ import useOnAppStateChange from "./useOnAppStateChange"
 import {
   isLocationEnabled,
   doesDeviceSupportLocationlessScanning,
+  subscribeToLocationStatusEvents,
 } from "../gaen/nativeModule"
 
 export type LocationPermissions = "NotRequired" | "RequiredOff" | "RequiredOn"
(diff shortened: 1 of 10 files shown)


src/gaen/nativeModule.ts
```diff
@@ -50,6 +50,34 @@ export const subscribeToEnabledStatusEvents = (
...
 }
 
+export const subscribeToBluetoothStatusEvents = (
+  cb: (enabled: boolean) => void,
+): EventSubscription => {
+  const ExposureEvents = new NativeEventEmitter(
+    NativeModules.ExposureEventEmitter,
+  )
+  return ExposureEvents.addListener(
+    "onBluetoothStatusUpdated",
```
aledustet, line 60: Curious if this event is also registered on the iOS side? If it isn't should we guard against it?

aledustet (approved):
Looks ready to me, just one question about platform availability. :shipit:

src/gaen/nativeModule.ts
juancruzgs: We are listening to that event on both platforms.
iOS is not emitting it, but that shouldn't be an issue

src/SystemServicesContext/useLocationPermissions.ts
```diff
@@ -41,6 +42,20 @@ const useLocationPermissions = (): LocationPermissions => {
...
+    const subscription = subscribeToLocationStatusEvents((enabled: boolean) => {
+      if (enabled) {
+        setLocationEnabledState("On")
+      } else {
+        setLocationEnabledState("Off")
+      }
+    })
+
+    return () => {
+      subscription?.remove()
```
johnschoeman: why do we need the nullish operator here?

johnschoeman (approved):
LGTM

src/SystemServicesContext/useLocationPermissions.ts
juancruzgs: I copied it from another subscription, I'll remove it

PR #481, opened 2020-09-30
Use standard button for checking for exposures

devinjameson:
Why:
----
As a user, I would like an intuitive way to check for new exposures. The previous button to check for new exposures was not entirely clear.

This commit:
----
- Update the check for new exposures to use a standard button
- Update our thinner buttons to use `Buttons.medium` instead of custom padding for each
- Remove the pull to refresh functionality from the exposure history because it is now redundant
- Remove the `lodash.isequal` dependency

gif
gif


src/ExposureHistory/History/index.spec.tsx
```diff
@@ -1,12 +1,13 @@
...
+import { fireEvent, render, waitFor } from "@testing-library/react-native"
 import { useNavigation } from "@react-navigation/native"
 
 import { ExposureDatum } from "../../exposure"
 import { DateTimeUtils } from "../../utils"
 import { factories } from "../../factories"
+import { ExposureContext } from "../../ExposureContext"
 
-import History from "."
+import History from "./index"
```
johnschoeman, line 10: ❤️

mattThousand (approved):

johnschoeman (approved):
Nice work!

aledustet (approved):
🚢  :shipit:

PR #482, opened 2020-09-30
Add SymptomGroup to context

amandabeiner:
Why:
As a user, I want to know how at risk I am so that I can get accurate
recommendations for my demographic information and symptoms.

This commit:
Adds a symptomGroup to the SelfScreenerContext and a list of possible
symptom groups. As a user fills out the self screener, we use the
useSymptomGroup hook to update the context with their risk category so
that we can show them useful information.

This required a significant change to our GeneralSymptoms and how they
are stored in the context. Though GeneralSymptoms are displayed together
in the UI, they are actually divided into PrimarySymtoms,
SecondarySymptoms and OtherSymptoms, which represent how relevant and
severe they are to determining whether the user at an elevated risk to
COVID-19.

Since the distinction between primary, secondary, and other symptoms is
core to the purpose of the SelfScreener (giving the user personalized
information), and displaying them as one screen is an implementation
detail for the view layer, I opted to change the data structure of the
SelfScreenerContext to separate symptoms by category instead of showing
them all together.

Though this means that each category of symptom now has its own hook and
setter function, it clarifies the complex conditionals for assigning
symptom groups in the core logic a lot.

### DIFF
--- a/src/SelfScreener/GeneralSymptoms.tsx
+++ b/src/SelfScreener/GeneralSymptoms.tsx
@@ -7,35 +7,34 @@ import { SelfScreenerStackScreens } from "../navigation"
 import { useSelfScreenerContext } from "../SelfScreenerContext"
 
 import { Button, GlobalText } from "../components"
-import {
-  GeneralSymptom,
-  OtherSymptom,
-  PrimarySymptom,
-  SecondarySymptom,
-} from "./selfScreener"
+import { OtherSymptom, PrimarySymptom, SecondarySymptom } from "./selfScreener"
 
 import SymptomCheckbox from "./SymptomCheckbox"
 
 const GeneralSymptoms: FunctionComponent = () => {
   const { t } = useTranslation()
   const navigation = useNavigation()
-  const { FEVER_OR_CHILLS, COUGH, DIFFICULTY_BREATHING } = PrimarySymptom
+  const {
+    FEVER_OR_CHILLS,
+    COUGH,
+    MODERATE_DIFFICULTY_BREATHING,
+  } = PrimarySymptom
   const { ACHING, LOSS_OF_SMELL_TASTE_APPETITE } = SecondarySymptom
   const { VOMITING_OR_DIARRHEA, OTHER } = OtherSymptom
   const {
     primarySymptoms,
     secondarySymptoms,
     otherSymptoms,
-    updatePrimarySymptoms,
-    updateSecondarySymptoms,
-    updateOtherSymptoms,
+    updateSymptoms,
   } = useSelfScreenerContext()
 
-  const symptomToString = (symptom: GeneralSymptom) => {
+  const symptomToString = (
+    symptom: PrimarySymptom | SecondarySymptom | OtherSymptom,
+  ) => {
     switch (symptom) {
       case FEVER_OR_CHILLS:
         return t("self_screener.general_symptoms.fever_or_chills")
-      case DIFFICULTY_BREATHING:
+      case MODERATE_DIFFICULTY_BREATHING:
         return t("self_screener.general_symptoms.difficulty_breathing")
       case COUGH:
         return t("self_screener.general_symptoms.cough")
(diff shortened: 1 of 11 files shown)


src/SelfScreener/selfScreener.ts
```diff
@@ -34,3 +41,149 @@ export enum AgeRange {
   EIGHTEEN_TO_SIXTY_FOUR,
   SIXTY_FIVE_AND_OVER,
 }
+
+export enum SymptomGroup {
+  EMERGENCY_SYMPTOM_GROUP,
```
devinjameson: It may be nice to shorten these a bit? 🤔

```
export enum SymptomGroup {
  EMEREGENCY,
  PRIMARY_1,
  PRIMARY_2,
  ...
}
```

devinjameson (approved):
Going out with a bang I see! Hot dang, this is 🔥 .

src/SelfScreener/selfScreener.ts
```diff
@@ -34,3 +41,149 @@ export enum AgeRange {
...
+}
+
+const hasUnderlyingCondition = (
+  underlyingConditions: UnderlyingCondition[],
+): boolean => {
+  return underlyingConditions.length > 0
+}
+
+const isOver65 = (ageRange: AgeRange | null): boolean => {
+  return ageRange !== null && ageRange === AgeRange.SIXTY_FIVE_AND_OVER
```
aledustet: Curious here doesn't the `===` for the value rules out a `null`

src/SelfScreener/selfScreener.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+  determineSymptomGroup,
+  SecondarySymptom,
+  OtherSymptom,
+  PrimarySymptom,
+  SymptomGroup,
+  UnderlyingCondition,
+  AgeRange,
+} from "./selfScreener"
+
+describe("determineSymptomGroup", () => {
```
aledustet, line 13: 😅  this is some test suite 👏👏 👏

src/SelfScreener/GeneralSymptoms.tsx
```diff
@@ -48,51 +54,53 @@ const GeneralSymptoms: FunctionComponent = () => {
...
-        onPress={() => updateGeneralSymptoms(COUGH)}
-        checked={generalSymptoms.includes(COUGH)}
+        onPress={() => updatePrimarySymptoms(COUGH)}
+        checked={primarySymptoms.includes(COUGH)}
       />
       <SymptomCheckbox
         label={symptomToString(LOSS_OF_SMELL_TASTE_APPETITE)}
-        onPress={() => updateGeneralSymptoms(LOSS_OF_SMELL_TASTE_APPETITE)}
-        checked={generalSymptoms.includes(LOSS_OF_SMELL_TASTE_APPETITE)}
+        onPress={() => updateSecondarySymptoms(LOSS_OF_SMELL_TASTE_APPETITE)}
```
aledustet: Wondering if it would be too overkill to expose `one` method from the context to update the symptoms and then make the determination there on where to store it. And the same to determine if a symptom has been selected. I feel like the separation is priceless for the decision in what the symptoms mean but we could make the view layer simpler by hiding that abstraction. It would probably make a pretty gnarly switch but the types are 💯  soooo it might not be that bad.

src/SelfScreener/selfScreener.spec.ts
johnschoeman, line 13: No comments other than this is a very nice set of specs for a function. 🥇

src/SelfScreener/selfScreener.ts
```diff
@@ -34,3 +41,149 @@ export enum AgeRange {
   EIGHTEEN_TO_SIXTY_FOUR,
   SIXTY_FIVE_AND_OVER,
 }
+
+export enum SymptomGroup {
```
aledustet, line 49: ❤️  when types are what takes it to that next level

src/SelfScreener/selfScreener.ts
```diff
@@ -34,3 +41,149 @@ export enum AgeRange {
...
+export type SelfScreenerAnswers = {
+  emergencySymptoms: EmergencySymptom[]
+  primarySymptoms: PrimarySymptom[]
+  secondarySymptoms: SecondarySymptom[]
+  otherSymptoms: OtherSymptom[]
+  underlyingConditions: UnderlyingCondition[]
+  ageRange: AgeRange | null
+}
+
+export const determineSymptomGroup = (
```
aledustet, line 69: decision tree by types 💪

aledustet (approved):
Really thorny logic made like a story with the names and types. I have but one comment that could have some substance but I think this has the explicitness it needs. 👏 👏  🚢 :shipit:

src/SelfScreener/selfScreener.ts
```diff
@@ -34,3 +41,149 @@ export enum AgeRange {
...
+  PRIMARY_SYMPTOM_GROUP_1,
+  PRIMARY_SYMPTOM_GROUP_2,
+  PRIMARY_SYMPTOM_GROUP_3,
+  SECONDARY_SYMPTOM_GROUP_1,
+  SECONDARY_SYMPTOM_GROUP_2,
+  NON_COVID_SYMPTOM_GROUP,
+  ASYMPTOMATIC_GROUP,
+}
+
+export type SelfScreenerAnswers = {
```
johnschoeman, line 60: I dig how this data structure came together!

src/SelfScreenerContext.spec.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/exhaustive-deps*/
```
johnschoeman, line 1: why was this es lint disable required?

src/SelfScreener/selfScreener.ts
amandabeiner: Got a little carried away 😅

src/SelfScreener/GeneralSymptoms.tsx
amandabeiner: This was actually a fun refactor. I love it!

src/SelfScreenerContext.tsx
```diff
@@ -71,17 +66,32 @@ export const SelfScreenerProvider: FunctionComponent = ({ children }) => {
     ageRange,
   })
 
+  const updateSymptoms = (symptom: GeneralSymptom) => {
```
aledustet, line 69: pure 🔥

aledustet (approved):
:shipit: 💪 🔥

johnschoeman (commented):
A bit late to the party, but LGTM

src/SelfScreenerContext.spec.tsx
amandabeiner: @johnschoeman This was because this rule was directing me to put dependencies in the array of the `useEffect` hook, but when I did that in the test env, it entered an infinite loop and crashed 😅. 

Since this is a test, and I wanted the hook to run only only once period, I thought it was okay to keep the dependency array empty, so I eslint-disabled so that it wouldn't give warning when we run yarn validate.

Could definitely be worth taking a second look, though! 🧐

PR #483, opened 2020-09-30
(Android) Do not translate app name

juancruzgs:
#### Description:

Translations for the `app_name` were added in this [PR](https://github.com/Path-Check/gaen-mobile/pull/297). That makes the app show different app names based on the OS language.
This PR removes those translations since we only have one app name per HA.

#### Linked issues:

https://pathcheck.atlassian.net/browse/GAEN-347


johnschoeman (approved):
NIce catch! Ill look into lokalize and remove anything there that might cause this to happen again.

PR #484, opened 2020-09-30
Cleanup stale my health data when view is loaded

aledustet:
Why:
----

The data we are storing on the device to populate the symptom log and the daily check-ins, should not be kept longer than the required period, currently 14 days.

This Commit:
----

- Add methods to delete the stale data on the iOS native side
- Add methods to delete stale data on the Android side
- Move native module for the my health into the same folder
- Cleanup old stale data before pulling info on the symptom log

Co-authored-by: Juan Cruz Soler <juancruz.gsoler@gmail.com>

### DIFF
--- a/src/MyHealth/nativeModule.ts
+++ b/src/MyHealth/nativeModule.ts
@@ -1,0 +1,55 @@
+import { NativeModules } from "react-native"
+
+import {
+  CheckIn,
+  SymptomLogEntry,
+  SymptomLogEntryAttributes,
+} from "../MyHealth/symptoms"
+
+// Check In Module
+const checkInModule = NativeModules.CheckInModule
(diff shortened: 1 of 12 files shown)


src/MyHealth/nativeModule.ts
```diff
@@ -0,0 +1,55 @@
+import { NativeModules } from "react-native"
```
johnschoeman, line 1: Nice, thanks for pulling this module out.

johnschoeman (approved):
💯

src/MyHealth/nativeModule.ts
aledustet: 💯  credit to you

PR #486, opened 2020-09-30
Add Guidance screen to Self Screener

amandabeiner:
Why:
As a user, I want to get personalized guidance when I say that I am
experiencing symptoms and give my demographic information.

This commit:
Adds a `Guidance` screen, which reads the calculated `symptomGroup`
value from the SelfScreener context. Based on the user's symptom group,
it displays the appropriate CDC-recommended guidance and tells the user
what to do.

This commit also updates the EmergencySymptomsQuestions component to use
the `symptomGroup` from context instead of re-calculating the value
based on the length of the emergencySymptoms array.

guidance

### DIFF
--- a/src/SelfScreener/Guidance.spec.tsx
+++ b/src/SelfScreener/Guidance.spec.tsx
@@ -1,0 +1,126 @@
+import React from "react"
+import { render } from "@testing-library/react-native"
+import { SelfScreenerContext } from "../SelfScreenerContext"
+import Guidance from "./Guidance"
+import { factories } from "../factories"
+import { SymptomGroup } from "./selfScreener"
+
+describe("Guidance", () => {
+  it("displays the appropriate heading for primary 1 symptoms", () => {
+    const context = factories.selfScreenerContext.build({
+      symptomGroup: SymptomGroup.PRIMARY_1,
+    })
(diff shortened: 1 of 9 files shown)


src/SelfScreener/Guidance.spec.tsx
```diff
@@ -0,0 +1,125 @@
+import React from "react"
+import { render } from "@testing-library/react-native"
+import { SelfScreenerContext } from "../SelfScreenerContext"
+import Guidance from "./Guidance"
+import { factories } from "../factories"
+import { SymptomGroup } from "./selfScreener"
+describe("Guidance", () => {
```
johnschoeman, line 8: space above here?

src/SelfScreener/Guidance.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+  it("displays the appropriate heading for primary 1 symptoms", () => {
```
johnschoeman, line 9:
```suggestion
  describe("when the symptom answers are primary 1 symptoms", () => {
    it("displays the no medical risk header", () => {
```

johnschoeman (approved):
💯

PR #489, opened 2020-10-01
Remove local eula and unused mocks

aledustet:
Why:
----

The EULA storing passed to be online and maintaining those files in our project is confusing since they are not in use. We also have accumulated test setup logic that is no longer relevant since those dependencies were removed or replaced.

This Commit:
----

- Delete the `locales/eula` folder with the locally stored eulas
- Remove react-native-local-resource since it was only used for the html files
- Remove mocks for unused dependencies

How to test:
----

- Tests still pass and app continues to build properly


johnschoeman (approved):
So nice. 💯

PR #490, opened 2020-10-01
Add missing translations and fix typo

aledustet:
Why:
----

There were some empty keys that resulted in failures on the CI process. Also, a type on one of the symptoms was reported from lokalise. There were available translations for the most recent content added.

This Commit:
----

- Pushed and pulled new changes from lokalise
- Updated translations and typo


johnschoeman (approved):
💯

PR #493, opened 2020-10-01
Navigate to self screener from my health

johnschoeman:
Why:
When a user taps 'Unwell' on the my health screen, we would like for
them to start the self screener flow.

This commit:
Adds the logic to navigate to the self screener from the my health tab.
Note that given the way that the modal navigation is factored, we need
to mount two instances of the component with a prop to determine where
to go back to after closing the modal. A future commit might want to
refactor this such that we dont need to have many mounted instances of
the same component.


aledustet (approved):
:shipit:

devinjameson (approved):
👋

PR #495, opened 2020-10-02
Update documentation

johnschoeman:
Why:
Some of our documentation is out of date and we would like for new users
to the project be able to onboard without the friction of stale info.

This commit:
- Updates the example.env.bt
- Updates the readme
- Removes the unused changelog
- Updates the pull request template to follow the team's conventions


aledustet (approved):
Could the example file change be a reason why it is not building right now?

johnschoeman:
> Could the example file change be a reason why it is not building right now?

Potentially, though that would be very odd. Ill try separating them out.

PR #496, opened 2020-10-02
Support metric system units

johnschoeman:
Why:
Currently we only have support for units using the imperial system. We
would like for the app to be able to support metric units.

This commit:
Adds a configuration based of an environment variable for setting the
measurement system to either metric or imperial. If metric is not
provided, we default to imperial.


aledustet (approved):
🚢  :shipit:

PR #497, opened 2020-10-02
Add copy for self screener start

johnschoeman:
Why:
Currently we arent providing copy for the self screener intro screen,
which is causing CI to fail for any commit that updates the translations
file.

This commit:
Adds the copy to the self screener intro screen.

Screen Shot 2020-10-02 at 9 16 34 AM

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -272,6 +272,7 @@
       "urgent_medical_attention_needed": "Urgent medical attention may be needed. Please call 911 or go to the Emergency Department"
     },
     "emergency_symptoms": {
+      "are_you_experiencing": "Are you experiencing any of these emergency symptoms?",
       "chest_pain": "Severe, constant chest pain or pressure",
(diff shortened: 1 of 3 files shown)


src/locales/en.json
```diff
@@ -257,6 +257,11 @@
     "how_the_app_works": "How the app works",
     "legal": "Legal"
   },
+  "self_screen": {
```
aledustet: this was a typo on the `CallEmergencyServices.tsx` if we update the start of the key we can delete this one

aledustet (approved):
Beat me to it 😄  other than the type on the `self_screen` key this is a go for me. I'll delete it on `lokalise` so if you delete it from here we are a go.

aledustet:
@johnschoeman I checked, no need to update lokalise, as long as we merge this one first and the other `911` updates after

johnschoeman:
@aledustet , so i cant pass ci without including `self_screen`, is lokalize updated yet?

PR #502, opened 2020-10-03
Remove "CheckIn" Model, Update "My Health" Screen

mattThousand:
### Why
We'd like to remove the concept of a "Check In" from the app because it added complexity in its implementation and did not provide significant value to the user. 

### This Commit
This commit consolidates the "today" and "overtime" screens into a single screen that 
- allows the user to log symptoms
- allows the user to navigate to the symptom checker if they are not feeling well

This commit also removes the native iOS `CheckInModule` and the related code for storing and fetching `CheckIns` from the database

Simulator Screen Shot - iPhone 11 Pro - 2020-10-03 at 16 58 56


aledustet (commented):
Great idea! I feel like we could use this edit to also clean up the UI a bit and re-arrange the CTA properly. Should we remove also the remaining navigation constants? And perhaps the android side resolvers?

mattThousand:
> Great idea! I feel like we could use this edit to also clean up the UI a bit and re-arrange the CTA properly. Should we remove also the remaining navigation constants? And perhaps the android side resolvers?

Sure! Do we know how we'd like the UI to look? For the android side, I was imagining that @juancruzgs would make a separate PR when he has availability

src/locales/en.json
```diff
@@ -171,15 +171,6 @@
     "question_icon": "Question mark icon",
     "unchecked_checkbox": "Unchecked checkbox"
   },
-  "my_health": {
```
aledustet, line 174: There is also a portion in there for the health assessments and probably the section on the results of the daily check-in. Should we remove them on this PR? Perhaps create a couple of cleanup tickets to go with some follow-up work?

src/MyHealth/MyHealthContent.tsx
```diff
@@ -132,43 +80,71 @@ const DaySummary: FunctionComponent<DaySummaryProps> = ({
...
-      {logEntries.map((logEntry) => {
+      {symptomLogEntries.map((logEntry) => {
         return <LogEntry key={logEntry.id} logEntry={logEntry} />
       })}
     </>
   )
 }
 
-const OverTime: FunctionComponent = () => {
+const MyHealthContent: FunctionComponent = () => {
```
aledustet: Isn't this a little too generic? Now that we have the one screen with multiple elements and the data safety check for an empty state I think it would make sense to break this one up in a couple of small components and leave the data guard aspect on a `MyHealth.tsx` or even an `index` file here in this folder. Thoughts?

src/MyHealth/SymptomLogContext.spec.tsx
```diff
@@ -78,10 +26,9 @@ describe("SymptomLogProvider", () => {
         date: mockedDate,
         symptoms,
       }
-      ;(getCheckIns as jest.Mock).mockResolvedValue([])
       ;(getLogEntries as jest.Mock).mockResolvedValue([])
 
-      const AddCheckInStatus = () => {
+      const AddCheckSymptomLogEntry = () => {
```
aledustet: Apologies but I'm a little confused about this naming, is this inner component to test that a new symptom log is created? Should it be `AddSymptomLog` or `LogSymptoms`?

src/MyHealth/symptoms.ts
```diff
@@ -83,54 +67,29 @@ const groupLogEntriesByDay = (allEntries: SymptomLogEntry[]): LogDataByDay => {
...
 
     if (groupedEntries[entryDateBeginningOfDay]) {
-      const newLogEntries = groupedEntries[entryDateBeginningOfDay].logEntries
+      const newLogEntries = groupedEntries[entryDateBeginningOfDay]
       newLogEntries.push(entry)
       newLogEntries.sort(compareDatesAscending)
-      groupedEntries[entryDateBeginningOfDay].logEntries = newLogEntries
+      groupedEntries[entryDateBeginningOfDay] = newLogEntries
     } else {
-      groupedEntries[entryDateBeginningOfDay] = {
```
aledustet, line 91: Part of the reason this got so bloated and separated was because of the data massaging that we needed to do. Could a simpler grouping improve this complicated logic now that we don't need a data structure that invested? Could a `[date, [log]]` mapping be enough, right before we display the collection?

aledustet (approved):
This looks good, I'm a little concerned about loose ends remaining after we move on from the cleanup, that's the nature of some of my comments but I feel like this could definitely be an iterative process, no need to tackle them all right now unless we feel we need to do them now.

devinjameson:
LGTM, UI looks good enough and can be improved in a future PR.

devinjameson:
Also gotta love that "just removed 1000 lines of code" feeling. 😍

src/MyHealth/MyHealthContent.spec.tsx
```diff
@@ -1,61 +1,73 @@
...
-    it("displays a 'no logs' message", () => {
-      const { getByText } = render(
+describe("MyHealthContent", () => {
+  it("allows the user to add a symptom log entry", async () => {
+    const navigateSpy = jest.fn()
+    ;(useNavigation as jest.Mock).mockReturnValue({
+      navigate: navigateSpy,
+    })
+    const { getByLabelText } = render(
+      <SymptomLogContext.Provider value={factories.symptomLogContext.build({})}>
```
johnschoeman:
```suggestion
      const initialContext = factories.symptomLogContext.build()
      <SymptomLogContext.Provider value={initialContext}>
```

I try to avoid inlining things like this as they tend to grow unruly over time.

src/MyHealth/Today.spec.tsx
```diff
@@ -1,192 +0,0 @@
-import React from "react"
```
johnschoeman, line 1: This seems like a pretty big change in its own right. What do you think about separating the prs of 'Removing the save CheckIn feature' And 'Remove the Today Tab' ? its not clear to me that they need to go together? and its preferable to have PRs be as small as possible.

Additionally, Why are we removing the Today vs Over time tabs? I think we still want to have those separate tabs? and we just wanted to remove the 'Feeling Good' button for the time being until we can figure out what its meant to do?

johnschoeman (commented):
Before merging this, can we clarify why we are removing the OverTime and Today tabs?

PR #508, opened 2020-10-05
CI Fix for PC Track

mattThousand:
### Why
We'd like to be able to generate development archives of the iOS app for testers to use

### This Commit
This commit updates the `BT_Development` scheme to archive using the `Debug` build configuration


ios/COVIDSafePaths.xcodeproj/xcshareddata/xcschemes/BT_Development.xcscheme
```diff
@@ -111,7 +111,7 @@
       buildConfiguration = "Debug-BT">
    </AnalyzeAction>
    <ArchiveAction
-      buildConfiguration = "Release-BT"
+      buildConfiguration = "Debug-BT"
```
aledustet, line 114: This is probably me talking from lack of knowledge but won't this build the schema that needs the metro server?

aledustet (approved):
:shipit:  I have one question that would most definitely educate me 😄

johnschoeman (approved):

ios/COVIDSafePaths.xcodeproj/xcshareddata/xcschemes/BT_Development.xcscheme
mattThousand: good question - since it's archiving it will not be an issue 👍

PR #510, opened 2020-10-05
Add bluetooth and location events to ios

johnschoeman:
Why:
A previous commit added the `onBluetoothStatusUpdated` and
`onLocationStatusUpdated` events to for listening to required status
changes on the android side. This causes errors on the ios side when
initializing the app as these events are not registered in the native
module.

This commit:
Adds the events to the ios native layer. Note that these events are not
needed to be used on ios builds, but we still subscribe to them on the
js layer for sake of simplicity.


aledustet (approved):
🚢 :shipit:

PR #521, opened 2020-10-05
Style symptom log screen

johnschoeman:
Why:
We would like to update the styles and layout of the symptom log screen.

This commit:
Moves the create symptom log button to the bottom of the screen. We also
refactored the list to be flatter, previously we were rendering lists
of lists to group the symptom logs by date. By rendering the full date on each symptom log entry, we were able
to greatly reduce the complexity in the codebase.

Before       |  After
:-------------------------:|:-------------------------:
Simulator Screen Shot - iPhone 11 - 2020-10-05 at 18 28 19 | Simulator Screen Shot - iPhone 11 - 2020-10-05 at 17 21 19


src/MyHealth/symptoms.ts
```diff
@@ -44,53 +42,3 @@ export const determineHealthAssessment = (
   }
   return HealthAssessment.NotAtRisk
 }
-
-type WithADate = { date: Posix }
-const compareDatesAscending = (
```
aledustet, line 49: Did we end up sorting the logs in any form? Should we have them sorted? It seems on the screenshot that they are not.

src/MyHealth/index.tsx
```diff
@@ -1,64 +1,106 @@
...
+
+  const NoSymptomHistory = () => {
+    return (
+      <Text style={style.noSymptomHistoryText}>
+        {t("symptom_checker.no_symptom_history")}
+      </Text>
+    )
+  }
+
+  const SymptomHistory = () => {
```
aledustet, line 34: I see we are not grouping entries now, but shouldn't we sort them to stay consistent in displaying them?

src/MyHealth/SymptomLogContext.spec.tsx
```diff
@@ -102,7 +103,6 @@ describe("SymptomLogProvider", () => {
     it("fetch all daily check-in and log entries to serialize", async () => {
```
aledustet: This description seems outdated with what the body of the test is doing, should it be `get the log entries on load`

src/MyHealth/SymptomLogContext.spec.tsx
```diff
@@ -132,7 +132,7 @@ describe("SymptomLogProvider", () => {
 })
 
 const CurrentCheckInStatus = () => {
```
aledustet: Maybe not the best name for this dummy component since no check-in logic is present anymore.

src/locales/en.json
```diff
@@ -365,7 +365,7 @@
     "get_tested": "We recommend you get tested for COVID-19.",
     "guidance": "Guidance",
     "log_symptoms": "Log symptoms",
-    "no_symptom_history": "No symptom history",
+    "no_symptom_history": "No symptom history…",
```
aledustet, line 368: Did we run a 'yarn i18n:check' just in case since we removed a file?

aledustet (approved):
Good decision removing the grouping, it definitely makes the data easier to consume.

src/MyHealth/index.tsx
johnschoeman: Yes! great catch, we are not sorting anymore. We were thinking about grabbing that in a follow up pr, but it makes sense in this one i think.

PR #522, opened 2020-10-06
Remove at risk recommendation screen

johnschoeman:
Why:
Currently if a user creates a symptom log they are taken to a
recommendation screen. We would like to fully separate the feature of
making a health recommendation and tracking symptoms as this allows the
features to be more clear and more easily maintained.

This commit:
Removes the AtRiskRecommendation Screen from the app.


aledustet (approved):
🚢  :shipit:

PR #523, opened 2020-10-06
Update styling to make screens use similar patterns

devinjameson:
Why:
----
We were using different design patterns on different screens, which made the app
feel disjointed and created a suboptimal user experience.

This commit:
----
Updates various screens to use the same design patterns. For example, showing
the primary action button in a similarly styled container that is stuck to the bottom
of each screen, using the same floating container styles throughout, and using
similar background colors.

gif

### DIFF
--- a/src/Connect/index.tsx
+++ b/src/Connect/index.tsx
@@ -145,7 +145,8 @@ const style = StyleSheet.create({
     marginBottom: Spacing.large,
   },
   headerText: {
-    ...Typography.header2,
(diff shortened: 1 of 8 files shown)


devinjameson:
@johnschoeman Thoughts on the new design for the symptom log time header? It feels cleaner to me.

johnschoeman (approved):
LGTM

aledustet (approved):
🚢  :shipit:

PR #529, opened 2020-10-07
(Android) Remove CheckIn model from Realm

juancruzgs:
#### Why:

The concept of "Check In" has been removed from the UI

#### This commit:

Removes CheckIn model from Realm and runs the data migration. 
iOS has already removed it here https://github.com/Path-Check/gaen-mobile/pull/502


aledustet (approved):
🚢  :shipit:

johnschoeman (approved):

PR #531, opened 2020-10-08
Use full week of covid data and show cases trend

aledustet:
Why:
----

The info shown to users should be of a more actionable nature. This could be achieved by showing a trend of the cases instead of just the cold numbers.

This Commit:
----
- Switch endpoint to pull full week's worth of data. Now the endpoint uses the same data set but it is aggregated for a week worth of data.
- Calculate trend from last week compared to today cases
- Switch the component on the home screen to a trend data clip. Instead of showing all 4 values, we derive the trend from today's cases count and last week average.

Reviewers:
----

Currently, the `StateData` component is not being used, the plan is to include it on the following commit as a details screen for the data.

Co-authored-by: Devin Jameson <devin@toughtbot.com>
Co-authored-by: Samuel Zimmerman <zimmerman.samuel@gmail.com>

Screen Shot 2020-10-08 at 9 29 02 AM

### DIFF
--- a/src/COVIDDataDashboard/covidDataAPI.ts
+++ b/src/COVIDDataDashboard/covidDataAPI.ts
@@ -1,9 +1,9 @@
-import { Posix, posixToDayjs } from "../utils/dateTime"
 import Logger from "../logger"
 
-const COVID_DATA_ENDPOINT = "https://api.covidtracking.com/v1/states/"
-const DATE_ARGUMENT_FORMAT = "YYYYDDMM"
-const ENDPOINT_FORMAT = "json"
+const COVID_DATA_ENDPOINT =
+  "https://localcoviddata.com/covid19/v1/cases/covidTracking?state="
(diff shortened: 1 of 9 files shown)


src/COVIDDataDashboard/covidDataAPI.ts
```diff
@@ -38,51 +34,44 @@ export interface RequestFailure {
...
+  const endpointURL = `${COVID_DATA_ENDPOINT}${state}${DAYS_IN_PAST_PARAM}`
   try {
     const response = await fetch(endpointURL, {
       method: "GET",
       headers: requestHeaders,
     })
 
     const jsonResponse = await response.json()
+    const historicData = jsonResponse[DATA_KEY]
+    console.log(historicData)
```
johnschoeman, line 57: remove this console.log?

johnschoeman (approved):

PR #532, opened 2020-10-08
(Android) Simulate exposures

juancruzgs:
#### Why:

We need a way to test the exposure history flow without having to wait 24 hours until a real exposure is detected.

#### This commit:

Reads exposures from Realm instead of calling the API.
In that way we can created mocked exposures from the "Simulate Exposure" button.


johnschoeman (approved):
💯

PR #533, opened 2020-10-08
Implement new home screen design

devinjameson:
Why:
----

The home screen does not currently display the most important and useful content to users. We would like for the home screen to be more helpful.

This commit:
----

- Implements a new design for the home screen
- Renames `ProxmityTracing` to `ExposureNotifications` because Exposure Notifications is the term used at the OS level to describe the system service.
- Uses the term `Exposure Detection` to describe the activation status of the app. If `Exposure Detection` is on, then the app is working properly and exposures are being detected.
- Moves the `Exposure Detection` status information to a new screen that is accessible from the Home screen
- Adds the "Talk to a contact tracer" button to the home screen
- Adds a `useExposureDetectionStatus` hook that provides an `exposureDetectionStatus` boolean. We added this hook to extract the logic of determining the status of `Exposure Detection` so it can be reused. We did not add this logic to the `SystemServicesContext` because the hook uses information from the `PermissionsContext` and `SystemServicesContext`, so we thought it would be simpler to put it in a hook, at least for now.

Notes:
----

- In a future PR, we will solve some navigation-related issues on the Home screen. Some of the screens open up by pushing the screen to the left, but we would like for these to come up from the bottom like standard modals.
- The Home screen is currently very stark, so we would like to add some helpful illustrations to the floating containers in a future PR.

GIF:
----

gif


src/Activation/ActivateExposureNotifications.tsx
```diff
@@ -17,7 +17,7 @@ import { isPlatformiOS } from "../utils"
 
 import { Spacing, Typography, Buttons, Colors } from "../styles"
 
-const ActivateProximityTracing: FunctionComponent = () => {
+const ActivateExposureNotifications: FunctionComponent = () => {
```
johnschoeman, line 20: IRC, we landing on `Proximity Tracing` as not all users know or care about the term `exposure notification` which is a google apple specific concept, would it make sense to keep it as `Proximity Tracing`?

johnschoeman: Chatted, irl: The reasoning is that while `Proximity Tracing` may be better english, the whole point of this screen is to help users understand why the exposure notifications feature is not working. This is what it is called in the os, and it would be confusing to change the name in the app.

src/Home/ExposureDetectionStatus.tsx
```diff
@@ -0,0 +1,60 @@
...
+      style={style.container}
+      contentContainerStyle={style.contentContainer}
+    >
+      <Text style={style.headerText}>
+        {t("screen_titles.exposure_scanning")}
+      </Text>
+      <Text style={style.subheaderText}>{subheaderText}</Text>
+      <BluetoothActivationStatus />
+      <ExposureNotificationsActivationStatus />
+      <LocationActivationStatus />
```
johnschoeman: What do you think about moving the logic that determines whether or not to show this component into its parent, i.e. `ExposureDetectionStatus` ?

src/Home/index.tsx
```diff
@@ -82,198 +86,263 @@ const Home: FunctionComponent = () => {
...
         style={style.container}
         contentContainerStyle={style.contentContainer}
       >
-        <View style={style.topScrollViewBackground} />
-        <GradientBackground gradient={Colors.gradient100} angleCenterY={1}>
-          <View style={style.topContainer}>
-            <SettingsButton />
-            <View style={style.topIconContainer}>
+        <Text style={style.headerText}>{t("screen_titles.home")}</Text>
+        <TouchableOpacity
```
johnschoeman: what do you think about `<SettingsButton />`

src/Home/index.tsx
```diff
@@ -82,198 +86,263 @@ const Home: FunctionComponent = () => {
...
+          onPress={handleOnPressSettings}
+        >
+          <SvgXml
+            xml={Icons.Gear}
+            width={Iconography.small}
+            height={Iconography.small}
+            fill={Colors.neutral100}
+          />
+        </TouchableOpacity>
+        <TouchableOpacity
```
johnschoeman: What do you think about `<ExposureStatusIndicator />` ?

src/Home/index.tsx
```diff
@@ -82,198 +86,263 @@ const Home: FunctionComponent = () => {
...
+              width={Iconography.tiny}
+              height={Iconography.tiny}
+            />
           </View>
-          <View style={style.ctaContainer}>
+        </TouchableOpacity>
+        <ShareLink />
+        {displayCovidData && <COVIDDataDashboard />}
+        {displayCallbackForm && (
+          <View style={style.floatingContainer}>
```
johnschoeman: What do you think about `<CallbackFormClip />`?

src/Home/index.tsx
```diff
@@ -82,198 +86,263 @@ const Home: FunctionComponent = () => {
...
-                  ...style.button,
-                  ...style.selfAssessmentButton,
-                }}
-                customButtonInnerStyle={style.buttonInner}
-                outlined
-              />
-            )}
           </View>
+        )}
+        <View style={style.floatingContainer}>
```
johnschoeman: What do you think about `<ReportAPositiveTestClip />`?

src/Home/index.tsx
```diff
@@ -82,198 +86,263 @@ const Home: FunctionComponent = () => {
...
+          <Button
+            onPress={handleOnPressReportTestResult}
+            label={t("home.submit_test_result_code")}
+            customButtonStyle={style.button}
+            customButtonInnerStyle={style.buttonInner}
+            hasRightArrow
+          />
         </View>
+        {displaySelfAssessment && (
+          <View style={style.floatingContainer}>
```
johnschoeman: what do you think about `<SelfAssessmentClip /> ?

src/Home/index.tsx
```diff
@@ -82,198 +86,263 @@ const Home: FunctionComponent = () => {
...
+            </Text>
+            <Button
+              onPress={handleOnPressTakeSelfAssessment}
+              label={t("home.bluetooth.take_self_assessment")}
+              customButtonStyle={style.button}
+              customButtonInnerStyle={style.buttonInner}
+            />
+          </View>
+        )}
+        <TouchableOpacity
```
johnschoeman: What do you think about `<CallEmergencyNumber />`

johnschoeman (approved):
💯 LGTM 💯

src/Home/index.tsx
devinjameson: 👍

src/Home/index.tsx
devinjameson: 👍

src/Home/index.tsx
devinjameson: 👍

src/Home/index.tsx
devinjameson: 👍

src/Home/index.tsx
devinjameson: 👍

src/Home/ExposureDetectionStatus.tsx
devinjameson: 👍

PR #534, opened 2020-10-08
Fix Display Bug in Debug Menu TEK List

mattThousand:
### Why
We'd like the `TEK` list (inside the debug menu) to display all keys fetched from the native layer

### This commit
This commit ads a unique id to the `ExposureKey` model in the iOS codebase such that it the collection can be displayed in a list
IMG_EEAB9602BA40-1

### DIFF
--- a/src/Settings/ENLocalDiagnosisKeyScreen.tsx
+++ b/src/Settings/ENLocalDiagnosisKeyScreen.tsx
@@ -64,10 +64,12 @@ const ENLocalDiagnosisKeyScreen: FunctionComponent<ENLocalDiagnosisKeyScreenProp
 
   return (
     <ScrollView>
-      {diagnosisKeys.map((_key: ENDiagnosisKey, index: number) => {
+      {diagnosisKeys.map((key: ENDiagnosisKey, index: number) => {
         return (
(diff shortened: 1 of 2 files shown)


johnschoeman (approved):

src/Settings/ENLocalDiagnosisKeyScreen.tsx
```diff
@@ -63,17 +63,15 @@ const ENLocalDiagnosisKeyScreen: FunctionComponent<ENLocalDiagnosisKeyScreenProp
...
-            Rolling start number: {item.item.rollingStartNumber}
-          </Text>
-        </View>
-      )}
-    />
+    <ScrollView>
+      {diagnosisKeys.map((_key: ENDiagnosisKey, index: number) => {
+        return (
+          <View key={index.toString()} style={style.flatlistRowView}>
+            <Text style={style.itemText}>{index}</Text>
```
johnschoeman:
```suggestion
            <Text style={style.itemText}>{`${index}:${key.rollingStartNumber}`}</Text>
```

johnschoeman (approved):

PR #535, opened 2020-10-08
Self screener -> self assessment

johnschoeman:
Why:
We were using the terms self screener and self assessment
interchangeably. We would like to consistently use self assessment

This commit:
- changes instances of screener to assessment.
- note that the image selfScreenerIntro.png is set in the build scripts
and would require updating a lot of the remote repo. We will do this in
a following commit at the same time as updating the remote repo.


devinjameson (approved):
🔍 
🔁 
👍 
🔥

PR #536, opened 2020-10-08
Use context for covid data and show data details

aledustet:
Why:
----

The data clip from the home screen should be the entry point to a more detailed set of data. This will allow us to increase the amount and the way we display the data we retrieve.

This Commit:
----
- Switch data consumption for a data COVID context, in order to make one request from the home screen and then feed the dashboard with it.
- Switch COVID for Covid on named components and files
- Add COVID data context factory
- Add tests for the data clip
- Navigate to the data dashboard from the data clip
- Add styles to the data clip
- Move COVID data context to the main component since the data dashboard will be part of the modal stack

Caveat:
----

Some of the data safety is still missing, for example, there is a repetition of the defensive code of checking for the state abbreviation and the trend is being calculated with incomplete data. More of that is coming.

Screen Shot 2020-10-08 at 5 18 04 PM Screen Shot 2020-10-08 at 5 18 08 PM


src/CovidDataDashboard/CovidDataClip.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+const CovidDataClip: FunctionComponent = () => {
+  const { stateAbbreviation } = useConfigurationContext()
+  const { t } = useTranslation()
+  const {
+    covidDataRequest: { data, status },
+  } = useCovidDataContext()
+
+  if (stateAbbreviation === null) {
```
johnschoeman, line 41: I wouldnt change anything in this pr, but this stateAbbreviation null checking jumps out as a bit of a smell. I wonder if there is a way to pull this up the chain and have some sort of `isAbleToFetchCovidData` validation and only pass data to the presentational components?  Id be afraid of making things too complicated, but seems like one of the harder parts of this feature in general, so might make sense to think through it on a follow up.

johnschoeman (approved):
Nice work this looks great!

App.tsx
```diff
@@ -53,8 +54,10 @@ const App: FunctionComponent = () => {
                   <ExposureProvider>
```
devinjameson, line 54: For context sharing, noting that we talked about cleaning up this Context Mayhem because it's getting out of hand.

src/CovidDataDashboard/CovidDataClip.tsx
```diff
@@ -0,0 +1,95 @@
...
+      case CovidDataRequestStatus.SUCCESS:
+        return (
+          <CovidDataStateTrend
+            lastWeekData={data}
+            stateAbbreviation={stateAbbreviation}
+          />
+        )
+      case CovidDataRequestStatus.NOT_STARTED:
+        return (
+          <Text style={style.errorMessageText}>
```
devinjameson, line 66: Why do we return the error message in the `NOT_STARTED` state?

src/CovidDataDashboard/CovidDataStateTrend.tsx
```diff
@@ -0,0 +1,141 @@
...
+  headerText: {
+    ...Typography.header4,
+    ...Typography.base,
+  },
+  subHeaderText: {
+    ...Typography.body3,
+  },
+  percentageContainer: {
+    borderRadius: Outlines.baseBorderRadius,
+    borderWidth: 0,
```
devinjameson, line 108: Do we need this `borderWidth`?

src/CovidDataDashboard/CovidDataStateTrend.tsx
```diff
@@ -0,0 +1,141 @@
...
+  },
+  percentageText: {
+    ...Typography.body1,
+    ...Typography.bold,
+    color: Colors.black,
+  },
+  infoContainer: {
+    marginLeft: Spacing.small,
+  },
+  trendArrowContainer: {
```
devinjameson, line 124: If you want you could probably remove this style and use `paddingLeft: Spacing.xxSmall` on the `percentageText` instead.

src/Home/index.tsx
```diff
@@ -264,7 +264,7 @@ const Home: FunctionComponent = () => {
         <SettingsButton />
         <ExposureDetectionStatus />
         <ShareLink />
-        {displayCovidData && <COVIDDataDashboard />}
+        {displayCovidData && <CovidDataClip />}
```
devinjameson, line 292:
```suggestion
        {displayCovidData && <CovidData />}
```

What are your thoughts on just using `CovidData` vs. `CovidDataClip`?

src/CovidDataDashboard/CovidDataStateTrend.tsx
```diff
@@ -0,0 +1,141 @@
...
+          text: t("covid_data.up_from_last_week"),
+          indicator: <SvgXml xml={Icons.ChevronUp} fill={Colors.black} />,
+        }
+      : {
+          text: t("covid_data.down_from_last_week"),
+          indicator: <SvgXml xml={Icons.ChevronDown} fill={Colors.black} />,
+        }
+
+  return (
+    <TouchableOpacity
```
devinjameson, line 59: Thoughts on pulling the `TouchableOpacity` and text with the `style.sectionHeaderText` out into the `Home/index`? It may be more organized there because that's how all the other "modules" on that page are arranged.

devinjameson (approved):
Few comments! 🔥

src/CovidDataDashboard/CovidDataClip.tsx
aledustet: Right there with you, I'm moving most of it to the data context and adding a few tests there now will come in the next PR 😄

src/CovidDataDashboard/CovidDataClip.tsx
aledustet: I was not sure about it, we probably should never reach this point, since this component shouldn't be rendered with this state but typescript is complaining that it is not complete and I liked not having a default case and have the type system keeping me honest. Maybe the name of the status is confusing?

src/Home/index.tsx
aledustet: `CovidData` is the name of the type we picked. This component is not displaying that per-se is just showing a 'clip' of that data. Does that make sense?

src/CovidDataDashboard/CovidDataStateTrend.tsx
aledustet: There is a component that is surrounding this, the data clip one that is making a number of checks on the data, to make sure that this component only gets displayed if the data is present, making it a bit simpler by splitting that complexity. I feel like moving it up because of the semantics and the positioning would move the complexity back up there. Thoughts?

src/CovidDataDashboard/CovidDataClip.tsx
devinjameson: I think it's fine for now!

src/Home/index.tsx
devinjameson: Ah I see, that works!

src/CovidDataDashboard/CovidDataStateTrend.tsx
devinjameson: Ah I see. Yeah probably simpler this way, at least for now.

PR #537, opened 2020-10-09
Improve Home screen UI

devinjameson:
Why:
----
The Home screen was a wall of text, which made the UX suboptimal. We would like for this screen to feel welcoming. We also had some styling and navigation issues we opted not to fix in the previous home screen redesign PR.

This commit:
----
- Renames the `Home` screen to `Dashboard` for users. This name more clearly communicates the purpose of the screen.
- Updates the design of the modules to remove the button, add an image, and make the entire container the button
- Updates the navigational elements for the screens that are navigated to directly from the home screen so they make sense

gif

### DIFF
--- a/src/AffectedUserFlow/Start.tsx
+++ b/src/AffectedUserFlow/Start.tsx
@@ -1,24 +1,13 @@
 import React, { FunctionComponent } from "react"
-import {
-  StyleSheet,
-  TouchableOpacity,
-  View,
-  Image,
-  ScrollView,
-} from "react-native"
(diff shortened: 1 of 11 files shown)


johnschoeman (approved):
Code looks great! and updates to the design are lovely! Did we check that the images we chose for the home screen work okay for builds that change the image set? (e.g. gr)

devinjameson:
@johnschoeman Yep! It's all good however GR needs to supply an image for the self screener if they want that customized, too. We use the same image from the self screener on the home screen.

PR #538, opened 2020-10-09
Fix language select header display issue

devinjameson:
Why: on large screen sizes, the select language header text was being cut off.

gif


johnschoeman (approved):
💯

PR #541, opened 2020-10-09
Refactor some navigation to improve reusability and accessibility

devinjameson:
Why: we were using suboptimal patterns to create some header navigation components. We would like to use better patterns to improve accessibility, UX, and reusability in the codebase.

This commit:
----
- Updates some navigational elements to use React Navigation to render them instead of rendering them within the component itself
- Removes the `ModalStack` and moves all the screens to the `MainNavigator`

gif


johnschoeman (approved):
💯

PR #546, opened 2020-10-11
Bugfix for COVID Data JSON Response Parsing

mattThousand:
### Why
We'd like the app to parse COVID stats data such that it displays as expected on the home screen

### This Commit
This commit addresses a bug whereby the client side code was expecting numerical data from the API to be returned with the `number` type, however it was coming back as the `string` type

### Before
Simulator Screen Shot - iPhone 11 Pro - 2020-10-10 at 21 09 27

### After
Simulator Screen Shot - iPhone 11 Pro - 2020-10-10 at 21 28 52

### DIFF
--- a/src/CovidDataDashboard/covidDataAPI.ts
+++ b/src/CovidDataDashboard/covidDataAPI.ts
@@ -17,13 +17,13 @@ export type CovidData = {
   peopleDeathNewCt: number
 }
 
-const toCovidData = ( jsonResponse: Record<string, string>[] ): CovidData[] => { 
-  return jsonResponse.map( ( json ) => {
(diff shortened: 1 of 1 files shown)


src/CovidDataDashboard/covidDataAPI.ts
```diff
@@ -17,6 +17,17 @@ export type CovidData = {
   peopleDeathNewCt: number
 }
 
+const toCovidData = (jsonResponse: Record<string, string>[]): CovidData[] => {
```
johnschoeman, line 20: nice!

johnschoeman (approved):
❤️

PR #548, opened 2020-10-11
Navbar Updates

mattThousand:
#### Why:
Several screens which are "pushed" to the nav stack rather than "presented" modally needed to be updated to display the navbar with a `back` button rather than hide the navbar and display an `X` icon

#### This commit:
This commit updates the relevant screens, and also displays the default `Back` label on iOS (to increase the tappable area of this button, which can be difficult to reach on larger iOS devices)

Oct-11-2020 11-29-48


devinjameson:
This is awesome well done!

devinjameson (approved):

PR #549, opened 2020-10-12
Update Modal Header Styles

mattThousand:
#### Why:
We'd like the app's UI to be as stylistically consistent as possible.

#### This commit:
This commit swaps out the "close" icon used across modal headers and changes the header background color to more closely match the UI used in the `Affected User` flow

Simulator Screen Shot - iPhone 11 Pro - 2020-10-11 at 21 06 04
Simulator Screen Shot - iPhone 11 Pro - 2020-10-11 at 21 01 54
Simulator Screen Shot - iPhone 11 Pro - 2020-10-11 at 21 00 41


johnschoeman (approved):

devinjameson:
@mattThousand I actually prefer the style that we are currently using on the Bluetooth, Exposure Notifications, Select Symptoms, etc. screens because it gracefully handles the "fixed header with scrolling content" situation. In particular the `secondary10` background color helps separate the content from the header. How do you feel about using that style throughout instead of the white background with the black x?

On screens where we don't have header text, we can just use the x in the circle instead of the black x with a white background.

PR #550, opened 2020-10-13
Add Settings to tab bar and remove Connect

devinjameson:
Why: we would like for Settings to live in the tab navigator bar of Connect.

This commit:
- Moves `Settings` to the tab bar
- Removes the `Settings` button from the top right of the home screen
- Moves some content from `Connect` to `Settings`
- Adds a spec for Settings with the specs that used to be on `Connect`

gif2
gif2


johnschoeman (approved):
LGTM!

PR #551, opened 2020-10-13
iOS Unit Test Cleanup

mattThousand:
#### Why:
We'd like the iOS unit tests to be as readable and maintainable as possible

#### This commit:
This commit removes duplicate setup code where possible from existing unit tests, and consolidates what is reusable in an extension on `ExposureManagerUnitTests`


johnschoeman (approved):
🧪

PR #552, opened 2020-10-13
(iOS) Open Settings App Url from Main Thread

mattThousand:
#### Why:
Per Apple's documentation, `open(...` should be called from the main thread (https://developer.apple.com/documentation/uikit/uiapplication/1648685-open)

#### This commit:
This commit adjusts the invocation to run on the main thread


johnschoeman (approved):
🔢

PR #554, opened 2020-10-14
Affected User Flow Navigation Bar Updates

mattThousand:
#### Why:
We'd like our navigation bar conventions to be consistent across the app

#### This commit:
This commit swaps out the preexisting navbar (custom back button, `x` icon), with the navbar used elsewhere in the app (system back button), and adds a `Done` button to the "Guidance" screen so that the user can return to the exposure history screen once they have completed the affected user flow.

Oct-13-2020 21-51-21

### DIFF
--- a/src/SelfAssessment/Guidance.tsx
+++ b/src/SelfAssessment/Guidance.tsx
@@ -183,6 +183,8 @@ const Guidance: FunctionComponent<GuidanceProps> = ({
     }
   }
 
+  const displayFindATestCenter = Boolean(findATestCenterUrl)
+
   const handleOnPressFindTestCenter = () => {
     if (findATestCenterUrl) {
       Linking.openURL(findATestCenterUrl)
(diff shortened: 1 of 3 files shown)


src/SelfAssessment/Guidance.tsx
```diff
@@ -183,8 +183,6 @@ const Guidance: FunctionComponent<GuidanceProps> = ({
     }
   }
 
-  const displayFindATestCenter = Boolean(findATestCenterUrl)
```
johnschoeman: Why was this removed? what happens if the jurisdiction does not have a test center url

johnschoeman (commented):
Code looks good, but looks like we are no longer respecting the safe area view?

src/SelfAssessment/Guidance.tsx
mattThousand: good catch 👍

johnschoeman (approved):
💯 

The gif still shows content going behind the notch, did we confirm if this is still an issue here?

PR #561, opened 2020-10-15
Exposure Detection Success/Failure Message

mattThousand:
#### Why:
Currently, when a user taps the `Check for exposures` button on the exposure history screen it updates the `Last updated at` info only.  This is quite subtle and its not always clear to users that the check for exposures happened successfully.  We would like to find a way to indicate to the user whether the check was successful or not, even if there where no new exposures found.

#### This commit:
This commit implements a toast/flash message indicating to the user whether the most recent exposure check was successful or not

Oct-15-2020 10-11-37

### DIFF
--- a/src/ExposureHistory/History/index.tsx
+++ b/src/ExposureHistory/History/index.tsx
@@ -55,12 +55,12 @@ const History: FunctionComponent<HistoryProps> = ({
     const checkResult = await checkForNewExposures()
     if (checkResult.kind === "success") {
       showMessage({
-        message: t("exposure_history.check_successful"),
+        message: t("common.success"),
         ...Affordances.successFlashMessageOptions,
       })
     } else {
       showMessage({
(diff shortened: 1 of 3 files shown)


src/ExposureHistory/History/index.tsx
```diff
@@ -44,7 +52,18 @@ const History: FunctionComponent<HistoryProps> = ({
 
   const handleOnPressCheckForExposures = async () => {
     setCheckingForExposures(true)
-    await checkForNewExposures()
+    const checkResult = await checkForNewExposures()
+    if (checkResult.kind === "success") {
+      showMessage({
+        message: t("exposure_history.check_successful"),
```
johnschoeman: what do you think about using the common translations here?  i.e. `common.success`

This is less to expect users to read and fewer translations for us to need to maintain

johnschoeman (approved):
💯

PR #565, opened 2020-10-16
Add status bar to Guidance screen.

mattThousand:
#### Why:
We'd like the status bar on the `Guidance` screen to have an opaque background so that content does not scroll directly underneath

#### This commit:
This commit adds a status bar background to the `Guidance` screen
Simulator Screen Shot - iPhone 11 Pro - 2020-10-16 at 09 11 48

### DIFF
--- a/src/SelfAssessment/Guidance.tsx
+++ b/src/SelfAssessment/Guidance.tsx
@@ -19,7 +19,7 @@ interface GuidanceProps {
 const Guidance: FunctionComponent<GuidanceProps> = ({
   destinationOnCancel = Stacks.ExposureHistoryFlow,
 }) => {
-  useStatusBarEffect("dark-content", Colors.primaryLightBackground)
+  useStatusBarEffect("dark-content", Colors.secondary10)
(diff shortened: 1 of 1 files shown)


johnschoeman (approved):
💯

src/SelfAssessment/Guidance.tsx
```diff
@@ -192,42 +193,45 @@ const Guidance: FunctionComponent<GuidanceProps> = ({
...
-        <Text style={style.subheaderText}>
-          {introForSymptomGroup(symptomGroup)}
-        </Text>
-      </View>
-      <View style={style.bulletListContainer}>
-        {instructionsForSymptomGroup(symptomGroup)}
-      </View>
-      {displayFindATestCenter && (
+    <>
+      <StatusBar backgroundColor={Colors.primaryLightBackground} />
```
devinjameson: Thoughts on using `secondary10` for both of these background colors?

devinjameson (approved):
Nice!

PR #568, opened 2020-10-16
Update check for exposures button style

devinjameson:
Why: we would like to have a consistent button style for buttons that are fixed to the bottom of the screen. Currently the check for exposures button on the exposure history screen uses a different style than the save button on the the select symptoms screen.

This commit:
- Updates the exposure history screen to use the same button style as the select symptoms screen
- Abstracts the button styling to `Buttons` and `Typography` style modules
- Creates a `LoadingIndicator` component that can easily be applied to any screen
- Uses the new `LoadingIndicator` component to show the loading state when checking for new exposures

gif2


mattThousand (approved):
💯

johnschoeman (approved):
Noice

johnschoeman (approved):
💯

PR #570, opened 2020-10-16
Style symptom history

devinjameson:
Why: we would like the symptom history screen to be styled in a way that optimizes the UX.

This commit:
- Styles the symptom history screen

gif


mattThousand (approved):
👏

PR #573, opened 2020-10-19
Stop sending date argument on the callback form

aledustet:
Why:
----

The Callback api is receiving an optional date argument that it is being sent as null on the request and it is not being consumed on the API side. We need to stop sending it while we get the correct value for us to send.

This Commit:
----

- Remove the `exposureDate` argument from the request that goes out to the callback form api endpoint.


devinjameson (approved):
🚢

PR #574, opened 2020-10-19
Tell user to call emergency services if they have emergency symptoms

devinjameson:
## Why:
We would like to inform users that they should call emergency services when they have symptoms that are considered "emergency" symptoms.

## This commit:
- Adds the logic to the data models for symptoms and symptom history to determine whether the symptoms the user has selected are emergency symptoms
- Updates the flow for logging symptoms to navigate the user to the call emergency symptoms screen if they have emergency symptoms
- Updates the `ModalHeader` component to receive an optional `onDismiss` function. We added this because we won't always want to navigate a user "back" to the previous screen when dismissing a modal.
- Changes the UX for displaying the call emergency services screen on the self assessment flow to use a modal
- Fixes a spelling issue with the word `symptom` (AKA `symptotopom`) because that word is actually impossible to spell

gif


aledustet (approved):
🚢  :shipit:

PR #575, opened 2020-10-19
Allow users to share the symptom history as text

aledustet:
Why:
----
When users add their symptoms to the log, there might be a case where that log would be valuable to send to someone of the users choosing. We need to allow users to pick that data from their log history and take it with them in a text format that can be shared.

This Commit:
----
- Create a format for the symptom history to be shared, the format is of the shape:

> Symptom history from Oct 20, '20 to Oct 7, '20
> 
> Oct 20, '20
> No symptoms were logged
> 
> Oct 19, '20
> You felt well
> 
> Oct 18, '20
> You did not feel well, symptoms included: Cough, Fever
> 
> (for the next 14 entries)
> ...

- Use formatter class to export a symptom history into this format
- Add a share button to the right of the header on the symptom history screen that will open up the share menu with the symptom history in text format to be shared

Large GIF (418x896)

Co-authored-by: Matt Buckley <matt@nicethings.io>

### DIFF
--- a/src/SymptomHistory/index.tsx
+++ b/src/SymptomHistory/index.tsx
@@ -1,6 +1,13 @@
 import React, { FunctionComponent } from "react"
-import { ScrollView, StyleSheet, View } from "react-native"
+import {
+  ScrollView,
+  Share,
+  StyleSheet,
+  TouchableOpacity,
+  View,
+} from "react-native"
 import { useTranslation } from "react-i18next"
+import { SvgXml } from "react-native-svg"
 
 import {
   DAYS_AFTER_LOG_IS_CONSIDERED_STALE,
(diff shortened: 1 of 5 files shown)


src/SymptomHistory/index.tsx
```diff
@@ -10,14 +17,30 @@ import { SymptomEntry } from "./symptomHistory"
 import { Text, StatusBar } from "../components"
 import { useStatusBarEffect } from "../navigation"
 import SymptomEntryListItem from "./SymptomEntryListItem"
+import Formatter from "./Formatter"
```
devinjameson: Is there a better name for this? Maybe something like `FormatSymptomHistory` to be more specific?

src/SymptomHistory/index.tsx
```diff
@@ -10,14 +17,30 @@ import { SymptomEntry } from "./symptomHistory"
...
 const SymptomHistory: FunctionComponent = () => {
   useStatusBarEffect("dark-content", Colors.primaryLightBackground)
   const { t } = useTranslation()
   const { symptomHistory } = useSymptomHistoryContext()
 
+  const handleOnPressShareHistory = async () => {
+    const message = Formatter.forSharing(t, symptomHistory)
+
+    if (message) {
+      try {
```
devinjameson: Could the `const message` go in the `try` block? Not sure if that works or is good practice, but maybe we could get rid of the `if (message)`.

src/SymptomHistory/index.tsx
```diff
@@ -26,9 +49,24 @@ const SymptomHistory: FunctionComponent = () => {
...
         alwaysBounceVertical={false}
       >
-        <Text style={style.headerText}>
-          {t("symptom_history.symptom_history")}
-        </Text>
+        <View style={style.headerContainer}>
+          <Text style={style.headerText}>
+            {t("symptom_history.symptom_history")}
+          </Text>
+
```
devinjameson, line 51: I think we want this to be at the bottom and have the same style as the `ExposureHistory` screen and button.

devinjameson (approved):

PR #576, opened 2020-10-20
Handle empty exposure keys error publishing keys

aledustet:
Why:
----

In the event a user tries to publish their keys on a fresh device, the server will return an error since there can not be an hmac calculated on an empty set of exposure keys.

This Commit:
----

- Add a no-op error handling for the error message of an empty set of exposure keys


devinjameson (approved):
I'm no testing expert but LGTM!

PR #577, opened 2020-10-20
Allow verification codes of 6 to 16 alphanumeric chars

devinjameson:
Why:
----
As a user, I would like to be able to input a verification code on the report positive test result flow that is between 6 and 16 alphanumeric characters. Right now, the input only accepts exactly 8 digits.

This commit:
----
- Updates the code input to take a minimum of 6 and a maximum of 16 characters.
- Updates the code input to only accept alphanumeric characters and show an error message if non-alphanumeric characters are used
- Changes the keyboard type for the code input to be the default keyboard
- Fixes some accessibility issues with the code input screen so that the keyboard does not cover the input

gif


aledustet (approved):
🚢  :shipit:

PR #580, opened 2020-10-21
Activate Bluetooth screen on the activation stack

aledustet:
Why:
----

When the device Bluetooth is turned off, we should let users know so they can turn it on before start using the app.

This Commit:
----
- Create controller to determine the next activation screen, only needed from the bluetooth activation screen or the exposure notifications activation screen
- Remove the steps count from the navigation header on the activation stack

Large GIF (418x896)


src/navigation/ActivationStack.tsx
```diff
@@ -80,6 +88,13 @@ const ActivationStack: FunctionComponent = () => {
     default: activationStepsIOS,
   })
 
+  const activationSummary: ActivationStep = {
+    screenName: ActivationStackScreens.ActivationSummary,
+    component: ActivationSummary,
+  }
+
```
devinjameson:
```suggestion
```

devinjameson (approved):
🚢 :shipit:

PR #582, opened 2020-10-21
Remove loading indicator from button component

devinjameson:
Why:
----
We would like for our `Button` component to not handle any loading logic.

This commit:
----
- Removes the loading prop from the `Button` component
- Uses the `LoadingIndicator` component in place of displaying a loading state within the `Button` itself where applicable


aledustet (approved):
🚢  :shipit:

PR #583, opened 2020-10-21
Remove navigation header within PublishConsentForm

devinjameson:
Why:
----
We were rendering two headers on the `PublishConsentForm` screen. One was coming from the navigation stack and the other was being rendered within the screen component itself. We would like to only have the header coming from the navigation stack.

This commit:
----
- Removes the navigation header within the `PublishConsentForm` screen


aledustet (approved):
🚢  :shipit:

PR #585, opened 2020-10-21
Remove Button component

aledustet:
Why:
----

The complexity of maintaining a button component is out-weighting the benefits of the shared logic, so we would like to remove it an use styles and maintain the logic of the states on the buttons closer to it's usage.

This Commit:
----
- [x] Replace button with styles on error screen
- [x] Switch button for styles on most activation screens
- [x] Replace button with styles on welcome screen
- [x] Replace button with styles on callback form
- [x] Replace button for styles on self assessment
- [x] Replace button for styles on the affected user flow
- [x] Replace button for styles on the how it works screen
- [x] Replace button for styles on the consent for analytics screen
- [x] Replace button for styles on the Exposure details screen
- Remove Button component


src/modals/AnonymizedDataConsentScreen.spec.tsx
```diff
@@ -5,125 +5,104 @@ import { useNavigation } from "@react-navigation/native"
...
 
     describe("and they press to revoke consent", () => {
-      it("updates their consent status and navigates to settings", async () => {
+      it("updates their consent status and navigates back", async () => {
         expect.assertions(2)
         const navigationSpy = jest.fn()
         ;(useNavigation as jest.Mock).mockReturnValue({
-          navigate: navigationSpy,
+          goBack: navigationSpy,
+          setOptions: jest.fn(),
```
devinjameson, line 37: @aledustet Is there a better way to mock `setOptions`? I couldn't figure out how to do it any other way than just mocking it in each describe block.

mattThousand (approved):
🚢

PR #586, opened 2020-10-22
Delete Data Confirmation + Settings Stack Navigation Bar Consistency

mattThousand:
#### Why:
We'd like the "Delete Data" confirmation screen to have appropriate copy, and also we'd like to use consistent navigation bar styling for screens in the settings stack

#### This commit:
This commit updates copy for the "Delete Data" confirmation screen, and also ensures that screens in the settings stack use the standard white nav bar and back button

** Note: at some point the "How the app works" screen will need a different nav bar to reflect the fact that it is presented modally rather than pushed onto the stack

Co-authored by Sam Zimmerman @Ferrumofomega sam.zimmerman@pathcheck.org

Simulator Screen Shot - iPhone 11 Pro - 2020-10-21 at 21 47 36

### DIFF
--- a/src/navigation/SettingsStack.tsx
+++ b/src/navigation/SettingsStack.tsx
@@ -2,10 +2,7 @@ import React, { FunctionComponent } from "react"
 import {
   createStackNavigator,
   StackNavigationOptions,
-  HeaderBackButton,
 } from "@react-navigation/stack"
-import { useTranslation } from "react-i18next"
-import { useNavigation } from "@react-navigation/native"
 
 import Settings from "../Settings/"
 import Legal from "../Settings/Legal"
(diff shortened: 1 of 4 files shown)


devinjameson:
@mattThousand The "How It Works" stack from Settings could also just not use `TransitionsPreset.ModalTransition`, which would make it display from the right instead of from the bottom. I think I prefer that anyways.

src/navigation/SettingsStack.tsx
```diff
@@ -14,49 +11,17 @@ import ENLocalDiagnosisKey from "../Settings/ENLocalDiagnosisKeyScreen"
 import ExposureListDebug from "../Settings/ExposureListDebugScreen"
 import { SettingsStackScreens } from "./index"
 
-import { Colors, Headers } from "../styles"
+import { Headers } from "../styles"
 import DeleteConfirmation from "../Settings/DeleteConfirmation"
+import { applyHeaderLeftBackButton } from "./HeaderLeftBackButton"
```
devinjameson: Thoughts on moving this into the middle group of imports?

devinjameson (approved):
Nice! 🚢 :shipit:

PR #587, opened 2020-10-22
update ios env to EN_API_VERSION=2

artgibson:
#### Why:

The latest iOS EN API Version is 2 and we should stay up to date: https://developer.apple.com/documentation/bundleresources/information_property_list/enapiversion

#### This commit:

Updates env to use EN_API_VERSION=2


mattThousand (approved):
🚢 💯

PR #589, opened 2020-10-22
Small Spec Updates

mattThousand:
#### Why:
We'd like to have tests that confirm that the correct header text is being set on the anonymized data consent screen.

#### This commit:
This commit adds tests that confirm that the correct header text is being set on the anonymized data consent screen.

Co-Authored by Devin Jameson devin@thoughtbot.com


src/modals/AnonymizedDataConsentScreen.spec.tsx
```diff
@@ -75,7 +79,10 @@ describe("AnonymizedDataConsentScreen", () => {
 
       const buttonText = getByText("I Understand and Consent")
```
devinjameson:
```suggestion
```

devinjameson (approved):
LGTM!

PR #593, opened 2020-10-24
Handle for scenario in which app is restored from backup on iOS

mattThousand:
#### Why:
If the app is restored from a backup on a new device, the encryption key used to decrypt the realm database instance will fail to decrypt, and cause a crash. We want to avoid this, and instead create a new empty instance in this scenario.

#### This commit:
This commit handles a decryption failure by creating a new empty realm instance.


Ferrumofomega (approved):
LGTM. Thanks Matt.

PR #594, opened 2020-10-26
Improve home screen card UX

devinjameson:
Why: it was not clear to some users that the home screen cards were
tappable.

This commit:
- Updates the home screen cards to have clear button-like CTAs, thereby
  making it clear that the cards are tappable
- Adds a border above the call emergency services button to separate it
  from the primary actions on the home screen

gif


johnschoeman (approved):
💯 LGTM

PR #596, opened 2020-10-26
Handle network timeout for code submission

johnschoeman:
Why:
We are currently not timing out long network requests for the post
request to submit a verification code. We would like to apply a 5 second
timeout to this requests, like we are doing with the key submission
request.

This commit:
Switches the fetch call with a fetchWithTimeout call and adds and error
message for requests that timeout on the verification code form.


devinjameson (approved):
👍

PR #599, opened 2020-10-26
Fix various styling issues and inconsistencies

devinjameson:
Why: there are styling issues and inconsistencies across the app that should be fixed to improve UX.

This commit:
- Updates most of the buttons in the app to be full width, instead of having some full width buttons and some buttons that are only as big as their children
- Updates the nav headers on the `MoreInfo` and `ExposureDetails` screens to use the updated nav header style
- Updates the `ActivationStatusView` to use the same red and green colors as the `Home` screen
- Updates the self assessment bottom buttons to use the same "fixed bottom" style as several of the other screens
- Adds the `SelfAssessment` image to the self assessment intro screen and updates the styles slightly
- Updates the `DeleteConfirmation` screen styles to use the same "fixed bottom" button style as several of the other screens and improve the UX
- Updates some translations to improve copy clarity

### DIFF
--- a/src/styles/buttons.ts
+++ b/src/styles/buttons.ts
@@ -59,7 +59,7 @@ export const primary: ViewStyle = {
   ...Outlines.lightShadow,
   borderRadius: Outlines.borderRadiusMax,
   paddingHorizontal: Spacing.xHuge,
-  minWidth: 180,
+  width: "100%",
   maxWidth: Layout.screenWidth * 0.95,
   backgroundColor: Colors.primary.shade100,
 }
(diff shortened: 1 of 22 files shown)


src/styles/buttons.ts
```diff
@@ -121,8 +122,22 @@ export const card: ViewStyle = {
...
 export const fixedBottom: ViewStyle = {
   ...base,
-  paddingTop: Spacing.medium,
-  paddingBottom: Spacing.medium,
+  paddingVertical: Spacing.medium,
   backgroundColor: Colors.primary.shade100,
   width: "100%",
 }
+
+export const fixedBottomDisabled: ViewStyle = {
```
johnschoeman, line 130: :100:
i wonder if you could do the nested structure here? something like:
```typescript
Buttons.fixedBottom.disabled
Buttons.fixedBottom.thin
etc...
```

src/styles/typography.ts
```diff
@@ -203,6 +203,10 @@ export const buttonFixedBottom: TextStyle = {
   ...buttonPrimary,
 }
 
+export const buttonFixedBottomDisabled: TextStyle = {
```
johnschoeman, line 206: same here
```typescript
Typography.button.fixedBottomDisabled
Typography.button.secondary
...etc
```

johnschoeman (approved):
Looks great, no comments. I might be nice to show a before and after for other reviewers less familiar with the code base though.

src/styles/buttons.ts
devinjameson: That's the plan! 😄

PR #600, opened 2020-10-26
Verify users are over 18

devinjameson:
Why: we would like to verify that users are over the age of 18 before
they use the app.

This commit:
- Navigates the user to an age verification screen after the welcome screen
- If the user confirms they are over 18, they are navigated to the how it works stack
- If they user says they are not over 18, they are navigated back
- Adds `displayAgeVerification` and `minimumAge` to the `ConfigurationContext`
- Removes the spec for the `Welcome` screen because all it was did was test that the navigation was called with the `HowItWorks` stack, which doesn't seem very useful

gif


devinjameson:
@johnschoeman How do you feel about removing the `Welcome` spec? It doesn't seem very useful but perhaps I could be convinced otherwise.

johnschoeman:
> @johnschoeman How do you feel about removing the `Welcome` spec? It doesn't seem very useful but perhaps I could be convinced otherwise.

I agree. There wants to be one e2e spec to cover the navigation through the onboarding to truly get coverage of this. the mock on the unit spec is a thin approximation at best, and if this ever doesnt work properly it will be caught quickly by the team, so happy to see non-useful specs be pruned out.

johnschoeman (commented):
I feel like this needs to be a configurable feature before merging in. We probably need:
1. the ability to turn this on and off as a whole per build.
2. the ability to configure the age.
3. the ability to configure the copy shown.

Where 1 is a must have, and 2, and 3, might be follow ups based off user asks.

I would defer to you on this, but I would also say that this doesnt meet the design standard for the app. We probably want this to be a fully styled interstitial screen. Any thoughts on this?

devinjameson:
> I would defer to you on this, but I would also say that this doesnt meet the design standard for the app. We probably want this to be a fully styled interstitial screen. Any thoughts on this?

Totally depends what your design goals are. I don’t think using a native alert is intrinsically bad design.

The alert is probably faster and less intrusive for the user than bringing them to a new screen. The full screen modal would look more “on brand” and allow for more text and perhaps an image. Given one of our goals is to make the onboarding shorter overall, I opted for the alert here.

Thoughts? @johnschoeman

johnschoeman:
> Totally depends what your design goals are. I don’t think using a native alert is intrinsically bad design.
> 
> The alert is probably faster and less intrusive for the user than bringing them to a new screen. The full screen modal would look more “on brand” and allow for more text and perhaps an image.

I agree with this in general, but i feel that theres something to the fact that a full screen solution would have the same number of buttons a user needs to click.

> Given one of our goals is to make the onboarding shorter overall, I opted for the alert here.

I would assume that the more onbrand design goal would be put at a higher value here by jurisdictions, but im not sure. I think it makes sense to go with this and see if theres any user feedback generated one way or the other.

johnschoeman (approved):
Code looks great! continued the conversation in the main thread.

devinjameson:
@johnschoeman Updated to use a modal, no real reason not to.

PR #604, opened 2020-10-27
Refactor button styles

devinjameson:
Why: we would like to update the factoring of our button styles to make them more clear and easy to use in the view layer.

This commit:
- Refactors `buttons.ts`
- Updates all the StyleSheets that use the `Buttons` module to use the new style names


johnschoeman (approved):
:100:

PR #605, opened 2020-10-28
Refactor covid data

johnschoeman:
Why:
We would like to update data widget being shown on the covid data card
on the home screen. Currently a lot of the logic in the CovidData
context and feature is implicitly dependent on the format of the data
being pre calculated to a set of trends. In order to better support
changes in this feature, We would like to:

- More fully separate the view layer calculation of aggregates and the
raw data itself.

- More fully separate the parsing of network data and the applications
internal concept of a set of covid data.

This commit:
Is a refactor of the covid data analytics widget feature. No new
functionality has been added.

Screen Shot 2020-10-28 at 10 55 34 AM


src/CovidData/Card/index.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { FunctionComponent } from "react"
+
+import { useCovidDataContext } from "../Context"
+import CovidDataCard from "./CovidDataCard"
+
+const CovidDataCardConnector: FunctionComponent = () => {
+  const { request: dataRequest, locationName } = useCovidDataContext()
```
devinjameson:
```suggestion
  const { request, locationName } = useCovidDataContext()
```

src/CovidData/Card/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+  return <CovidDataCard dataRequest={dataRequest} locationName={locationName} />
```
devinjameson:
```suggestion
  return <CovidDataCard dataRequest={request} locationName={locationName} />
```

src/CovidData/PercentageChart.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { ReactNode, FunctionComponent } from "react"
+import { StyleSheet, View } from "react-native"
+import { SvgXml } from "react-native-svg"
```
devinjameson, line 3: Maybe we can organize these imports like the rest?

src/CovidData/Context.tsx
```diff
@@ -0,0 +1,94 @@
...
+  request: initialRequest,
+}
+
+export const CovidDataContext = createContext<CovidDataContextState>(
+  initialState,
+)
+
+export const CovidDataContextProvider: FunctionComponent = ({ children }) => {
+  const { stateAbbreviation, displayCovidData } = useConfigurationContext()
+  const [request, setCovidDataRequest] = useState(initialRequest)
```
devinjameson: Is there a reason this doesn't use the conventional naming scheme?

src/CovidData/covidDataAPI.ts
```diff
@@ -0,0 +1,184 @@
...
+      date,
+      peoplePositiveNewCasesCt,
+      peoplePositiveCasesCt,
+      peopleDeathCt,
+      peopleDeathNewCt,
+    } = networkDatum
+
+    return {
+      date,
+      positiveCasesNew: peoplePositiveNewCasesCt,
```
devinjameson: Thoughts on swapping this so the assignment happens in the destructuring from `networkDatum`?

devinjameson (approved):

src/CovidData/covidData.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export type CovidDatum = {
+  date: Date
+  positiveCasesTotal: number
+  positiveCasesNew: number
+  deathTotal: number
+  deathNew: number
+}
+
+export const valid = (_data: CovidData): boolean => true
```
johnschoeman: This should probably be removed

src/Home/SectionButton.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { FunctionComponent } from "react"
+import { View, StyleSheet } from "react-native"
+import { SvgXml } from "react-native-svg"
+
+import { Text } from "../components"
```
devinjameson, line 5:
```suggestion
import { Text } from "../components"

import { Icons } from "../assets"
import { Typography, Buttons, Iconography, Colors } from "../styles"
```

src/CovidData/Context.tsx
johnschoeman: nope, can fix!

PR #606, opened 2020-10-28
Flipper Setup

mattThousand:
#### Why:
We'd like to use `Flipper` (https://fbflipper.com/) to debug during development

#### This commit:
This commit sets up flipper for both native platforms

Screen Shot 2020-10-28 at 4 11 22 PM
Screen Shot 2020-10-28 at 3 54 49 PM


johnschoeman (approved):
Nice! thanks for taking this on.

PR #607, opened 2020-10-28
Introduce line charts

johnschoeman:
Why:
We would like to show a line chart in the covid data card to indicate to
users the trends of the virus.

This commit:
- Introduces a CovidDataInfo component to render a the set of covid data
analytic info as a set of labeled visual elements.

- Introduces a LineChart component which can render a scaled svg of a
provided data set as a curve.

- Introduced a SvgPath module to handle the building of an svg path
command from a provided set of XYCoordinates.

To-do:
- [ ] Render correct increasing/decreasing text
- [ ] Render correct down/up X% in past Y days text
- [ ] parameterize the data label
- [ ] show the data source
- [ ] Use a different API
    
Screen Shot 2020-10-28 at 10 28 06 PM

### DIFF
--- a/src/CovidData/Card/CovidDataInfo.tsx
+++ b/src/CovidData/Card/CovidDataInfo.tsx
@@ -1,0 +1,74 @@
+import React, { FunctionComponent } from "react"
+import { useTranslation } from "react-i18next"
+
+import { StyleSheet, View } from "react-native"
+
+import * as CovidData from "../covidData"
+import { Text } from "../../components"
+import LineChart from "../LineChart"
+
+import { Layout, Typography, Spacing, Colors } from "../../styles"
+
+interface CovidDataInfoProps {
+  data: CovidData.CovidData
+  locationName: string
+}
+
+const CovidDataInfo: FunctionComponent<CovidDataInfoProps> = ({
+  data,
+  locationName,
+}) => {
+  const { t } = useTranslation()
+
+  const lineData = CovidData.toLineChartCasesNew(data)
+
+  const lineChartWidth =
+    0.5 * (Layout.screenWidth - 2 * Spacing.medium - 2 * Spacing.medium)
+  const lineChartHeight = 80
+
+  const labelText = t("covid_data.new_cases")
+
+  return (
+    <View testID={"covid-data"}>
+      <Text style={style.headerText}>
+        {t("covid_data.spread_of_the_virus_in", { locationName })}
+      </Text>
+      <View style={style.dataContainer}>
+        <View style={style.trendContainer}>
+          <Text style={style.legendText}>{labelText}</Text>
+        </View>
+        <View style={style.chartContainer}>
+          <LineChart
+            lineData={lineData}
+            width={lineChartWidth}
+            height={lineChartHeight}
+            color={Colors.accent.success100}
+          />
+        </View>
+      </View>
+    </View>
+  )
+}
+
+const style = StyleSheet.create({
+  headerText: {
+    ...Typography.body2,
+    marginBottom: Spacing.xxSmall,
+  },
+  dataContainer: {
+    flexDirection: "row",
+    marginTop: Spacing.xSmall,
+    marginBottom: Spacing.medium,
+  },
+  trendContainer: {
+    flex: 4,
+  },
+  chartContainer: {
+    flex: 3,
+  },
+  legendText: {
+    ...Typography.body2,
+  },
+})
+
+export default CovidDataInfo
--- a/src/CovidData/LineChart.tsx
+++ b/src/CovidData/LineChart.tsx
@@ -1,0 +1,143 @@
+import React, { FunctionComponent } from "react"
+import { Svg, Path } from "react-native-svg"
+
+import * as SvgPath from "./SvgPath"
+import { Colors } from "../styles"
+
+interface LineChartProps {
+  lineData: number[]
+  width: number
+  height: number
+  color: string
+}
+
+const NUMBER_OF_HORIZONTAL_LINES = 6
+
+const LineChart: FunctionComponent<LineChartProps> = ({
+  lineData,
+  width,
+  height,
+  color,
+}) => {
+  if (lineData.length < 2) {
+    return null
+  }
+
+  // Scale Data
+  const max = Math.max(...lineData)
+  const min = Math.min(...lineData)
+  const shrinkPathBy = 2
+  const scaleFactor = height / (max - min) / shrinkPathBy
+  const toScale = (datum: number) => {
+    return (datum - min) * scaleFactor
+  }
+
+  // Fit Path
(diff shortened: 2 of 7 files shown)


src/CovidData/covidData.ts
```diff
@@ -39,3 +39,13 @@ const percentDifference = (a: number, b: number) => {
...
 }
+
+type TrendData = number[]
+
+export const toLineChartCasesPer100k = (data: CovidData): TrendData => {
+  return data.map(toCasesNewPer100k).slice(0, 7)
+}
+
+export const toCasesNewPer100k = (datum: CovidDatum): number => {
+  return datum.positiveCasesNew / 100000
```
johnschoeman: I believe this is incorrect, it should be 100k residents of the population.

src/CovidData/Card/index.tsx
```diff
@@ -1,12 +1,132 @@
...
+  Spacing,
+} from "../../styles"
 
 const CovidDataCardConnector: FunctionComponent = () => {
   const { request, locationName } = useCovidDataContext()
 
   return <CovidDataCard dataRequest={request} locationName={locationName} />
 }
 
+type CardContainerProps = {
```
johnschoeman: ended up inlining this component as the connector component is largely a wrapper around this main content card anyways.

src/CovidData/Card/CovidDataInfo.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <View testID={"covid-data"} style={style.container}>
+      <View style={style.header}>
+        <Text style={style.headerText}>
+          {t("covid_data.spread_of_the_virus_in", { locationName })}
+        </Text>
+      </View>
+      <View style={style.dataContainer}>
+        <View style={style.legendContainer}>
+          <Text style={style.legendText}>{"Cases Per 100k"}</Text>
```
johnschoeman: still need to localize and like pass these in as parameters.

src/CovidData/Card/CovidDataInfo.tsx
```diff
@@ -0,0 +1,79 @@
...
+          <Text style={style.sourceText}>{"source: localcoviddata.org"}</Text>
```
johnschoeman: this would likely come in a follow up pr, but we should be providing the api source.

src/CovidData/LineChart.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    type Segment = [number, number][]
+    const toSvgString = (el: Segment): string => {
+      return `S ${el[0][0]} ${el[0][1]} ${el[1][0]} ${el[1][1]}`
+    }
+
+    const scaledData = lineData.map(toScale)
+    const offsetData = scaledData.map(toOffset)
+    const coordinates = offsetData.map(toCoordinate)
+    const partitions = partition(coordinates, 2)
```
devinjameson: @johnschoeman Pretty sure this partition function is not doing what it should be. Let's check it out together tomorrow.

src/CovidData/LineChart.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const svgStrings = partitions.map(toSvgString)
+
+    const firstSvgPointString = `M ${coordinates[0][0]} ${coordinates[0][1]} `
+    const path = firstSvgPointString + svgStrings.join(" ")
+    return path
+  }
+
+  const buildHorizontalLineYPositions = (
```
devinjameson, line 79: @johnschoeman LOOK LOOK I DID A RECURSION

src/CovidData/LineChart.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    const toCoordinate = (el: number, idx: number): [number, number] => {
+      const xPadding = 10
+      const firstXPosition = xPadding / 2
+      const pathWidth = containerWidth - xPadding
+
+      const xStepWidth = pathWidth / (lineData.length - 1)
+      const xCoordinate = idx === 0 ? firstXPosition : idx * xStepWidth
+
+      return [xCoordinate, containerHeight - el]
```
devinjameson: @johnschoeman Pretty sure we were rendering the line chart upside down before, because the 0 point for SVG viewbox is top left.

src/CovidData/LineChart.tsx
devinjameson: This works, but there may be a cooler way to do it with recursion.

```
const array = [1, 2, 3, 4, 5, 6, 7]
const newArray = []

for(i = 0; i < array.length - 1; i++) {
  newArray.push([array[i], array[i+1]])
}
```

devinjameson: This should also work!

```
const buildPairs = (arr) => {
  return arr.length > 1
    ? [arr[0], arr[1]].concat(buildPairs(arr.slice(1)))
    : []
}
```

src/CovidData/LineChart.tsx
johnschoeman: Hey, good catch!

ios/Podfile.lock
```diff
@@ -272,11 +272,11 @@ PODS:
     - React-Core (= 0.63.2)
     - React-cxxreact (= 0.63.2)
     - React-jsi (= 0.63.2)
-  - Realm (5.4.7):
```
mattThousand: could we keep these dependencies as they were before?

src/CovidData/covidData.ts
mattThousand: good catch, I suppose we'll need to pass the HA population in through the `env`, and the calculation would be `datum.positiveCasesNew` / (`totalPopulation`/ `100000`)? 🤔

johnschoeman: Yup, i think this makes sense to be coming from the api, we dont want to be keeping track of the populations.  https://covidactnow.org/ looks like it has this available directly.

src/CovidData/LineChart.tsx
johnschoeman: nice work! though i dont think we need partitions at all, looking the svg docs, it looks like we were approaching building the path incorrectly. The S command only one point and one controlPoint, not a start and endpoint. it infers the start point and start controlpoint from the previous command.

ref: https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Paths

ios/Podfile.lock
johnschoeman: yup! for context dev's and my podfiles.lock are off, i think theres been a recent update that didnt have the dependencies resolved consistently.

src/CovidData/LineChart.tsx
devinjameson: Yeah I saw that as well, but I think if the control point is the next point, it actually works out fairly well — then we don’t have to calculate a control point.

devinjameson (approved):

PR #608, opened 2020-10-29
Handle Rate Limiting Gracefully

mattThousand:
#### Why:
We'd like to handle the case where the daily exposure detection threshold has been exceeded by circumventing the `Failure` UI. Users should be able to press the manual exposure detection button as often as they'd like without being exposed to messaging for this specific error. 

#### This commit:
This commit adds a check for the specific error code representing the case in which a user has exceeded the maximum number of daily calls to the exposure notifications framework, and presents the `Success` UI in that case

co-authored by: John Schoeman <john.schoeman@thoughtbot.com>
co-authored by: Devin Jameson <devin.jameson@thoughtbot.com>


johnschoeman (approved):
LGTM!

PR #609, opened 2020-10-29
Add trend info to covid data widget

johnschoeman:
Why:
We would like to show the trend info for the new cases covid data line
chart.

This commit:
Introduces the regression library to calculate the trend in the data
array. We also added better styling of the text and showed the source. A
future commit would likely remove the regression library if we can
change the api we are consuming to one that has this data already
calculated.

----

Simulator Screen Shot - iPhone 11 Pro Max - 2020-10-29 at 15 09 57

### DIFF
--- a/src/CovidData/Card/CovidDataInfo.tsx
+++ b/src/CovidData/Card/CovidDataInfo.tsx
@@ -1,11 +1,11 @@
 import React, { FunctionComponent } from "react"
-import { useTranslation } from "react-i18next"
-
 import { StyleSheet, View } from "react-native"
+import { useTranslation } from "react-i18next"
+import regression from "regression"
 
 import * as CovidData from "../covidData"
-import { Text } from "../../components"
 import LineChart from "../LineChart"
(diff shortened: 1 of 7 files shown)


src/CovidData/Card/CovidDataInfo.tsx
```diff
@@ -20,12 +20,25 @@ const CovidDataInfo: FunctionComponent<CovidDataInfoProps> = ({
...
+  }
+  const newCasesPoints = newCasesData.map(toPoint)
+  const result = regression.linear(newCasesPoints)
+  const trend = result.equation[0]
 
   const lineChartWidth =
     0.5 * (Layout.screenWidth - 2 * Spacing.medium - 2 * Spacing.medium)
   const lineChartHeight = 80
 
+  const trendText = trend > 0 ? "Trending up" : "Trending down"
```
devinjameson: Translation?

devinjameson (approved):
LGTM!

PR #610, opened 2020-10-29
Refactor typography styles

devinjameson:
Why: we would like for our typography styles to be more organized and easier to use.

This commit:
- Refactors `typography.ts` and updates all the places where the `Typography` module is used

Note:
- Calling font sizes, line heights, etc. "small", "medium", "large", etc. becomes hard to maintain. For this reason, this commit shifts to using "x10", "x20", "x30", etc., with "x10" being the smallest size.
- This also enables us to use the same patterns for sizing across different font properties, e.g. `size.x10` and `lineHeight.x10`. This ends up being easier to use and more readable than `smallFontSize` and `mediumLineHeight`.
- In the future we can pull this pattern into other styling modules, e.g. `Sizing.layout.x10`.

### DIFF
--- a/src/Activation/AcceptTermsOfService.tsx
+++ b/src/Activation/AcceptTermsOfService.tsx
@@ -142,7 +142,7 @@ const style = StyleSheet.create({
     paddingBottom: Spacing.xxxLarge,
   },
   headerText: {
-    ...Typography.header1,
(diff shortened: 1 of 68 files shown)


johnschoeman (approved):
LGTM!

mattThousand (approved):

PR #621, opened 2020-11-03
posixToDayjs test not handling UTC properly

stefanoschrs:
#### Why:
Tests fail because posixToDayjs test is not handling the UTC date properly

#### This commit:
Changed the test to take into consideration the UTC date.

#### Screenshots:
image

#### How to test:
```yarn test```

### DIFF
--- a/src/utils/dateTime.spec.ts
+++ b/src/utils/dateTime.spec.ts
@@ -35,9 +35,10 @@ describe("isToday", () => {
 
 describe("posixToDayjs", () => {
   it("converts a valid posix timestamp into a Dayjs instance", () => {
-    const posixTimeStamp = new Date("2020/07/01").getTime()
+    const date = new Date("2020/07/01")
+    const posixTimeStamp = date.getTime()
     expect(posixToDayjs(posixTimeStamp)?.format("YYYY-MM-DD")).toEqual(
-      "2020-07-01",
+      date.toISOString().split("T")[0],
     )
   })
 

### CODE REVIEW

**Review by johnschoeman — APPROVED** [review]
LGTM!
Though it strikes me as a bit odd that a posix time is affected by time zones at all.

It seems that "2020-07-01" and "2020/07/01" are parsed to slightly different epoch times, `1593561600000` and `1593576000000` respectively on my machine. What do you think about using the same date format in the setup and expectation?
so:
```typescript
const posix = Date.parse("2020-07-01")

const result = posixToDayjs(posix)

expect(result?.format("YYYY-MM-DD")).toEqual("2020-07-01")
```

I could also be convinced that this isnt all that useful of a test altogether and could be removed. thoughts?

stefanoschrs:
@johnschoeman using `Date.parse` is the same thing

```
> new Date(Date.parse("2020-07-01"))
2020-07-01T00:00:00.000Z
> new Date("2020/07/01")
2020-06-30T21:00:00.000Z
```

johnschoeman:
> @johnschoeman using `Date.parse` is the same thing

yup, I misspoke prior to my comments edit, i think the issue is related to the Date parsing "2020-07-01" and "2020/07/01" to slightly different times?

stefanoschrs:
Yeah it seems to be an inconsistency due to formating.

```
> new Date("2020/07/01 21:00")
2020-07-01T18:00:00.000Z
> new Date("2020-07-01 21:00")
2020-07-01T18:00:00.000Z
> 
> new Date("2020/07/01")
2020-06-30T21:00:00.000Z
> new Date("2020-07-01")
2020-07-01T00:00:00.000Z
```

PR #622, opened 2020-11-03
Provide recommendation to users for covid related symptoms

johnschoeman:
Why:
When a user enters symptoms to the symptom log that indicate they may
have Covid-19, we would like to show recommended guidance. We would also
like to prompt users to enter their symptom log for the current day on
the app's Dashboard.

This commit:
- Introduces a recommendation screen that will be navigated to if a user
has logged covid related symptoms.
- Introduces a module on the dashboard that will show a similar form to
the add symptoms form in the symptom history flow. Note that new screens
were added to the HomeStack so that the form can be open in a modal
directly from the home screen.
- Refactored the exposure detection component from the home screen to a
separate folder.

| Dashboard CTA | SymptomHistory |
|----|----|
|covid-recommendation-cta|covid-recommendation|

### DIFF
--- a/src/CovidRecommendation.tsx
+++ b/src/CovidRecommendation.tsx
@@ -1,0 +1,198 @@
+import React, { FunctionComponent } from "react"
+import {
+  Linking,
+  Image,
+  ScrollView,
+  StyleSheet,
+  TouchableOpacity,
+  View,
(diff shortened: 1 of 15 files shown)


mattThousand (approved):
👍 👏

PR #623, opened 2020-11-04
Improvement: Android get applicationId from .env.bt

stefanoschrs:
#### Why:
Since it's already filled in the .env files there is no reason to hardcode it in the build.gradle. This way allowing easier fork maintenance for the jurisdictions.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -148,7 +148,7 @@ android {
     }
 
     defaultConfig {
-        applicationId "org.pathcheck.covidsafepaths.bt"
+        applicationId project.env.get("ANDROID_APPLICATION_ID")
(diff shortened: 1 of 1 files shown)


johnschoeman (commented):
Looks like a good idea to me.  @juancruzgs if you have a moment, could we get a gut check from you on this one as well? would this conflict with any of the bitrise workflows?

juancruzgs:
It should work on Bitrise as well 💯 
This [function](https://github.com/Path-Check/gaen-mobile/blob/cc13d93f23bb80044becf51b53536a218cfbb153/bin/configure_builds.sh#L156) won't be needed anymore, it can be removed in a future PR

johnschoeman (approved):

PR #625, opened 2020-11-04
Matomo Analytics Events (optional)

mattThousand:
#### Why:
We'd like to implement a default set of analytics events to be tracked for users who have consented to providing anonymized data. 

#### This commit:
This commit introduces a small number of events that are sent to the Matomo analytics server for users who have opted in. In addition, this commit adjusts the preexisting event payload structure to send a category, action, and name.


App.tsx
```diff
@@ -49,16 +49,16 @@ const App: FunctionComponent = () => {
               userHasCompletedOnboarding={isOnboardingComplete}
             >
               <PermissionsProvider>
-                <ExposureProvider>
-                  <AnalyticsProvider>
+                <AnalyticsProvider>
```
johnschoeman:
```suggestion
                <ProductAnalyticsProvider>
```

src/ProductAnalytics/Context.tsx
```diff
@@ -6,22 +6,27 @@ import React, {
   useContext,
 } from "react"
 import Matomo from "react-native-matomo-sdk"
+
 import { StorageUtils } from "../utils"
-import { actions } from "./index"
 import { useConfigurationContext } from "../ConfigurationContext"
+import * as analyticsClient from "./analyticsClient"
```
johnschoeman:
```suggestion
import * as AnalyticsClient from "./analyticsClient"
```

johnschoeman: I like the idea of the AnalyticsClient, what do you think about passing it in as a prop to the provider?

so something like:
```typescript
<ProductAnalyticsProvider analyticsClient={Matomo}>
   ...consumers
</ProductAnalyticsProvider>
```

src/ProductAnalytics/analyticsClient.ts
```diff
@@ -0,0 +1,15 @@
+import Matomo from "react-native-matomo-sdk"
+
+import { EventCategory, EventAction } from "./Context"
+
+export const trackEvent = async (
```
johnschoeman: Not 100% sure i understand the value of these wrapper functions. Is there a plan to do something extra here? or are you wanting to add some separation from our application code and the Matomo sdk?

johnschoeman (approved):
Looks good! a few minor comments.

src/ProductAnalytics/analyticsClient.ts
mattThousand: they're intended to facilitate testability - i.e. the context's `trackEvent` function will always be called, but this allows us to test that the context only calls through to the `analyticsClient`'s client's function if the user has consented. Open to thoughts if you have a different approach in mind :)

PR #628, opened 2020-11-04
Track number of exposure notifications preceding positive diagnosis

mattThousand:
#### Why:
For users who have consented to share anonymized data, we'd like to track the number of exposure notifications received prior to submission of a positive diagnosis. 

#### This commit:
This commit sends an event to the analytics server at the time of key submission that includes the number of exposure notifications received up until that time. This commit also adjusts the naming schema for events, eliminating the `EventAction` type


johnschoeman (approved):
Looks good to me!

PR #630, opened 2020-11-05
isolate -> quarantine, correct medical term

artgibson:
#### Why:

Noted by Epidemiologist that Isolate is not the correct term, it should be quarantine

#### This commit:

replaces isolate with quarantine. and adds new quarantine term in EN language file.

#### Notes
- I updated the term to quarantine for the isolate key
- added a new quarantine key for future use
- I did not update code to minimize language translation needs. (Ideally switch to new term when we get translations)

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -117,7 +117,8 @@
       "schedule_callback": "Schedule a call to get support from a contact tracer from the {{healthAuthorityName}}.",
       "speak_with_contact_tracer": "Speak with a contact tracer",
       "wash_your_hands": "Wash your hands",
-      "wear_a_mask": "Wear a mask"
+      "wear_a_mask": "Wear a mask",
+      "quarantine": "Quarantine"
     },
     "exposure_window": {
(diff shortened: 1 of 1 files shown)


mattThousand (approved):

src/locales/en.json
```diff
@@ -111,13 +111,14 @@
...
-      "isolate": "Isolate",
+      "isolate": "Quarantine",
       "next_steps": "Next steps",
       "personalize_my_guidance": "Personalize my guidance",
       "schedule_callback": "Schedule a call to get support from a contact tracer from the {{healthAuthorityName}}.",
       "speak_with_contact_tracer": "Speak with a contact tracer",
       "wash_your_hands": "Wash your hands",
-      "wear_a_mask": "Wear a mask"
+      "wear_a_mask": "Wear a mask",
+      "quarantine": "Quarantine"
```
johnschoeman, line 121: If this isnt being used in the codebase, the key will be stripped during the ci process, so we will have to update the call sites as well to have this.

johnschoeman (approved):

PR #631, opened 2020-11-05
Use covidactnow api

johnschoeman:
Why:
We would like to use the covidactnow.org api as it has a data set that
will allow us to show covid statistic and metrics for both states and
counties.

This commit:
Introduces an api fetch that consumes the covidactnow.org data. We also
introduced the `ts.data.json` package to use json decoders for decoding
the response json.

Note that the package name includes several `.` characters with metro
bundler does not handle correctly when calculating the sha1 hash for the
cache key in the bundler haste map. We resolved this issue by
introducing a postinstall script for yarn that will rename the package
within nodemodules.

Co-Authored-By: matt buckley <matt@nicethings.io>

Simulator Screen Shot - iPhone 11 - 2020-11-05 at 10 43 43

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "clean": "rm -rf node_modules && yarn",
     "start": "yarn && react-native start",
     "tsc": "tsc --noEmit",
     "prettier": "prettier \"src/**/*.+(js|jsx|ts|tsx|json)\"",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -40,6 +40,7 @@
   "dependencies": {
     "@bugsnag/react-native": "^7.3.4",
     "@react-native-community/async-storage": "^1.8.1",
+    "@react-native-community/cli": "^4.13.0",
```
mattThousand: was this meant to be included?

johnschoeman: nope, can remove!

src/CovidData/API/covidActNow.ts
```diff
@@ -0,0 +1,153 @@
...
+      "ActualsTimeseries",
+    ),
+  },
+  "StateTimeseriesDecoder",
+)
+
+const toCovidData = (stateData: StateCovidData): CovidData => {
+  const timeseries = stateData.actualsTimeseries
+
+  const compact = <T>(arr: Array<T | null>): T[] => {
```
mattThousand, line 101: nice 👍

mattThousand (approved):

src/CovidData/API/covidActNow.ts
```diff
@@ -0,0 +1,153 @@
...
+type StateCovidData = {
+  fips: string
+  country: string
+  actuals: Actuals
+  actualsTimeseries: ActualsTimeseries
+}
+
+const ActualsDecoder = JsonDecoder.object<Actuals>(
+  {
+    date: JsonDecoder.isUndefined("2020-10-10"),
```
johnschoeman: we should use todays date here or handle or something here.

PR #636, opened 2020-11-06
fix About text sentence structure

artgibson:
#### Why:

Updated default About Text. The grammar was incorrect.

#### This commit:

- Adds correct grammar for about text in english text
- Updates test case

#### Notes:

If my suggestion is not valid, please do correct.


johnschoeman (approved):
LGTM, though looks like you'll need to run `yarn format` to get the validation script passing.

PR #638, opened 2020-11-06
Account for archive url 404 (iOS)

mattThousand:
### Bug
The `index` file on the key server may specify urls of key archives that have been deleted from the server. These urls will return `404`s.

### Fix
Instead of aborting the operation in the case of a `404` for an individual archive file, we proceed and download the files that _are_ present on the server


ios/BT/ExposureManager.swift
```diff
@@ -601,7 +601,13 @@ private extension ExposureManager {
           case .success (let package):
             downloadedPackages.append(package)
           case .failure(let error):
-            reject(error)
+            // The index file may list key archive URLs that have been deleted
+            // from the key server. These will return 404's. Instead of aborting
+            // the entire operation, we continue and download the key archives that
+            // are present on the server
+            if (error as? GenericError != GenericError.notFound) {
```
johnschoeman, line 608: What do you think about a switch statement here?

johnschoeman (approved):
Lgtm

PR #642, opened 2020-11-09
Remove file processing limit for iOS 13.7+

mattThousand:
### Why
Devices running `iOS 13.7+` are not limited to processing a specific number of key archives in a given day. For this reason, we want to allow the app to process as many files as are available on a given day. 
Screen Shot 2020-11-09 at 12 19 16 PM

### This Commit
This commit removes the upper limit on the number of files the iOS app processes per day for users running `iOS 13.7+`

### DIFF
--- a/ios/BT/ExposureManager.swift
+++ b/ios/BT/ExposureManager.swift
@@ -32,6 +32,8 @@ final class ExposureManagerError: NSObject, LocalizedError {
   }
 }
 
+enum ENAPIVersion { case V1, V2 }
+
 @objc(ExposureManager)
 /**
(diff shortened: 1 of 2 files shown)


ios/BT/ExposureManager.swift
```diff
@@ -32,6 +32,8 @@ final class ExposureManagerError: NSObject, LocalizedError {
   }
 }
 
+enum ENAPIVersion { case V1, V2 }
```
johnschoeman, line 35: 💯

johnschoeman (approved):
LGTM

PR #645, opened 2020-11-09
(Android) Upgrade exposure notifications SDK to 1.7.2

juancruzgs:
#### This commit:

- Upgrades exposure notifications SDK to **1.7.2**
- Upgrades WorkManager to **2.5.0-beta01**

Following the [best practices for background work](https://developers.google.com/android/exposure-notifications/implementation-guide#create-background-work) we are now marking the background worker as important (`setForegroundAsync`).
While the worker is running it will display a notification.

#### Screenshots:

device-2020-11-09-154631


johnschoeman (approved):

android/app/src/main/res/values/strings.xml
```diff
@@ -6,4 +6,5 @@
   <string name="exposure_notification_channel_name">Exposure Notification</string>
   <string name="exposure_notification_message">Someone you were near has tested positive for COVID-19. Tap for more info.</string>
   <string name="exposure_notification_title">Possible COVID-19 exposure</string>
+  <string name="background_worker_notification_title">Checking exposures…</string>
```
juancruzgs, line 9: What should we show while we are downloading diagnosis keys?
It also needs to be translated

johnschoeman: I think this copy is fine.  We should be able to add these to the android translations.  Happy to do it if you wanna cut card for it.

juancruzgs: Sure!
https://pathcheck.atlassian.net/browse/GAEN-463

PR #647, opened 2020-11-09
Fix various accessibility issues across app

devinjameson:
Why: there are accessibility issues in the app that result in a poor experience for screen readers.

This commit:
- Updates the error message text on the `CodeInputForm` to only be accessible when there is an error message.
- Removes the screen number dots from the `HowItWorksScreen` because they suggest to users that the screens should be navigable with swipe gestures (they are not)
- Adds an `accessibilityRole` of `button` to the "How we protect your privacy" button
- Adds an `accessibilityLabel` of "Select language" to the language button on the `WelcomeScreen` and also adds an `accessibilityRole` of `button`
- Adds an `accessibilityLabel` of "Select language" to the language button on the `Settings` index
- For screen readers, focuses the first language option on the `LanguageSelection` screen when the screen mounts
- Removes duplicate accessibility labels on the bottom tab bar (we currently have accessibility labels on the tab bar icons and the svgs within the tab bar icons)
- Updates the settings tab bar icon to use the correct name and label (it is currently using the old name and label for the `Connect` screen)

### DIFF
--- a/src/components/ListItem.tsx
+++ b/src/components/ListItem.tsx
@@ -18,7 +18,7 @@ const ListItem: FunctionComponent<ListItemProps> = ({
   icon,
 }) => {
   return (
-    <TouchableOpacity onPress={onPress} accessible accessibilityLabel={label}>
+    <TouchableOpacity onPress={onPress} accessibilityLabel={label}>
       <View style={style.listItem}>
         <SvgXml
           fill={Colors.primary.shade100}
@@ -26,8 +26,6 @@ const ListItem: FunctionComponent<ListItemProps> = ({
           width={Iconography.small}
           height={Iconography.small}
           style={style.icon}
-          accessible
-          accessibilityLabel={label}
         />
         <Text style={style.listItemText}>{label}</Text>
       </View>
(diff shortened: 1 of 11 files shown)


src/components/ListItem.tsx
```diff
@@ -8,26 +8,27 @@ import { Iconography, Spacing, Typography, Colors } from "../styles"
...
   icon: string
 }
 
 const ListItem: FunctionComponent<ListItemProps> = ({
   label,
+  accessibilityLabel,
   onPress,
   icon,
 }) => {
+  console.log(accessibilityLabel)
```
johnschoeman: extra conosle?

src/locales/en.json
```diff
@@ -217,7 +218,7 @@
     "home": "Dashboard",
     "more_info": "More Info",
     "settings": "Settings",
-    "symptom_history": "Symptom Checker"
+    "symptom_log": "Symptom Log"
```
johnschoeman:
```suggestion
    "symptom_history": "Symptom History"
```

johnschoeman (approved):
LGTM! Awesome work

PR #651, opened 2020-11-10
Fix various issues with product analytics

johnschoeman:
Why:
Currently we have a handful of visual and logic bugs in the product
analytics feature.

This commit:
- Verifies if the correct Matomo variables have been provided before
initializing the Matomo SDK
- moves the logic of determining if the build should enable Product
analytics out of the Product analytics context to the app initialization
to reduce complexity in the context.
- Separates the analytics consent screen into two components, one for
the onboarding activation flow and one for the settings screen to reduce
complexity in the navigation logic and reduce complexity in the forms
themselves.
- explicitly states if product analytics should be enabled rather than
inferring from the presence of the Matomo url
- removes the attempted tracking of a user opting into product analytics
as its current behavior only tracks when a user opts out of sharing
data, but still labels it opting in. We will add this back in a future
commit.
- removes the tracking call of if a user received an exposure
notification as this caused crashes in some cases. We will add this back
in a future commit.


App.tsx
```diff
@@ -45,21 +45,39 @@ const App: FunctionComponent = () => {
   }, [])
 
   const initializeMatomo = () => {
-    const displayAnalytics = env.DISPLAY_ANALYTICS
     const matomoUrl = env.MATOMO_URL
     const matomoSiteId = parseInt(env.MATOMO_SITE_ID)
-    if (displayAnalytics) {
+
+    if (matomoUrl && !isNaN(matomoSiteId)) {
```
devinjameson: Is it worth giving `!isNaN(matomoSiteId)` a name?

App.tsx
```diff
@@ -45,21 +45,39 @@ const App: FunctionComponent = () => {
...
+    const nullClient = {
+      trackEvent: async (_category: string, _action: string) => {},
+      trackView: async (_routes: string[]) => {},
+    }
 
-  const productAnalyticsClient: ProductAnalyticsClient = {
-    trackEvent: Matomo.trackEvent,
-    trackView: Matomo.trackView,
+    if (displayAnalytics) {
+      const client = initializeMatomo()
```
devinjameson, line 73: 💯

src/Activation/ProductAnalyticsConsentForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+  TouchableOpacity,
+} from "react-native"
+import { useNavigation } from "@react-navigation/native"
+import { useTranslation } from "react-i18next"
+
+import { useConfigurationContext } from "../ConfigurationContext"
+import { useProductAnalyticsContext } from "../ProductAnalytics/Context"
+import { ActivationStackScreens, useStatusBarEffect } from "../navigation"
+
+import { Text } from "../components"
```
devinjameson, line 15: I think stuff from `components` may want to live in the second group of imports.

src/Activation/ProductAnalyticsConsentForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const handleOnPressMaybeLater = () => {
+    navigation.navigate(ActivationStackScreens.ActivationSummary)
+  }
+
+  const handleOnPressPrivacyPolicy = () => {
+    Linking.openURL(healthAuthorityPrivacyPolicyUrl)
+  }
+  const privacyPolicyLinkText = t("product_analytics.privacy_policy")
+
+  const buttonText = t("product_analytics.i_understand_and_consent")
```
devinjameson, line 43: Is there a reason these bits of text are pulled out?

devinjameson (approved):
Great work!

src/Activation/ProductAnalyticsConsentForm.tsx
johnschoeman: artifact of the refactor, can inline.

App.tsx
johnschoeman: i think so!

mattThousand (approved):

PR #653, opened 2020-11-10
Disable "Check for Exposure" button while check is in progress

mattThousand:
#### Why:
The native layer will infrequently return an error code if the "Check for Exposures" button is tapped in _very_ rapid succession. This results in an error toast being presented to the user, which we'd like to avoid/
Nov-10-2020 14-18-17

#### This commit:
This commit disables the "Check for Exposures" button while an exposure check is in progress.


johnschoeman (approved):
Noice

PR #654, opened 2020-11-10
Update location permissions related copy and design

devinjameson:
Why: we would like to make it more clear that the app never accesses or tracks location data.

This commit:
- Updates the copy and design of the Activate Location screen in the onboarding flow
- Updates the copy and design of the Location card on the Exposure Detection Status screen
- Updates the copy and design of the Location Info screen that is accesses via the Location card on the Exposure Detection Status screen

Screen Shot 2020-11-10 at 2 30 27 PM
Screen Shot 2020-11-10 at 2 44 41 PM
Screen Shot 2020-11-10 at 2 56 01 PM


johnschoeman (approved):
LGTM, though we should double check the API number being 10 and below.

PR #655, opened 2020-11-10
Fix flash message display issue

devinjameson:
Why: the styling of the flash message does not take into account the
screen insets, so the flash message is showing up on with the correct
amount of top padding only on iOS devices with notches.

This commit:
- Updates the top padding for the flash message to use the screen insets

gif


johnschoeman (approved):
LGTM

PR #656, opened 2020-11-10
Add header to Settings screen

devinjameson:
Why: we would like the Settings screen to have a header to match the rest of the tab screens.

Screen Shot 2020-11-10 at 5 14 54 PM

### DIFF
--- a/src/Home/index.tsx
+++ b/src/Home/index.tsx
@@ -151,6 +151,7 @@ const SymptomHistory: FunctionComponent = () => {
   const { t } = useTranslation()
 
   const handleOnPressSymptomHistory = () => {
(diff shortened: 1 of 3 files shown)


src/Home/index.tsx
```diff
@@ -151,6 +151,7 @@ const SymptomHistory: FunctionComponent = () => {
   const { t } = useTranslation()
 
   const handleOnPressSymptomHistory = () => {
+    ;<Text style={style.headerText}>{t("screen_titles.home")}</Text>
```
johnschoeman: this looks incorrect?

johnschoeman (approved):

src/Home/index.tsx
devinjameson: Yep it is!

PR #659, opened 2020-11-10
Make publish consent screen shorter

devinjameson:
Why: there is currently too much content on the screen which results in a poor UX, especially on smaller devices.

Screen Shot 2020-11-10 at 6 13 13 PM


johnschoeman (approved):
should we run `yarn i18n:extract` on this?

devinjameson:
> should we run `yarn i18n:extract` on this?

Since this will remove other translations that don't pertain to this PR, I would opt for putting that in a future PR.

PR #664, opened 2020-11-12
Update activation flow ordering and styling

devinjameson:
Why: we would like for the Share Analytics Consent screen to be at the start of the Activation flow. This will enable us to gather analytics on the rest of the Activation flow if the user opts in to sharing analytics.

This commit:
- Moves the Share Analytics Consent screen to the start of the Activation flow
- Updates the header for the Activation screens to only have a "Back" button instead of the "App Setup" title and a close button

gif

### DIFF
--- a/src/navigation/ActivationStack.tsx
+++ b/src/navigation/ActivationStack.tsx
@@ -67,7 +67,7 @@ const ActivationStack: FunctionComponent = () => {
     }
     baseActivationSteps.unshift(acceptTermsOfService)
   }
-
+  ActivationSummary
   if (!isBluetoothOn) {
     const activateBluetooth: ActivationStep = {
       screenName: ActivationStackScreens.ActivateBluetooth,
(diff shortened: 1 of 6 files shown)


src/navigation/ActivationStack.tsx
```diff
@@ -67,7 +58,7 @@ const ActivationStack: FunctionComponent = () => {
     }
     baseActivationSteps.unshift(acceptTermsOfService)
   }
-
+  ActivationSummary
```
johnschoeman: not sure this is needed?

johnschoeman (approved):
LGTM, it would be nice to get the navigation to be less imperative and have the 'next screen' be decoupled from the screens themselves here in a future commit.

src/navigation/ActivationStack.tsx
devinjameson: Oopsie!

PR #665, opened 2020-11-12
Reset product analytics on delete all data

johnschoeman:
Why:
When a user elects to 'Delete all data' we would like to be deleting the
data related to remembering that the user has opted into sharing product
analytics.

This commit:
Adds the logic the to product analytics context to reset and invokes
this logic in the delete all data feature.

### DIFF
--- a/src/ProductAnalytics/Context.tsx
+++ b/src/ProductAnalytics/Context.tsx
@@ -18,6 +18,7 @@ export type ProductAnalyticsContextState = {
     value?: number,
   ) => Promise<void>
   trackScreenView: (screen: string) => Promise<void>
+  resetUserConsent: () => Promise<void>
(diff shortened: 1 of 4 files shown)


src/ProductAnalytics/Context.tsx
```diff
@@ -18,13 +18,15 @@ export type ProductAnalyticsContextState = {
     value?: number,
   ) => Promise<void>
   trackScreenView: (screen: string) => Promise<void>
+  resetProductAnaltyics: () => Promise<void>
```
mattThousand: current naming certainly works, but what do you think about calling this something like `resetUserConsent` instead?

mattThousand (approved):
LGTM 👍

PR #666, opened 2020-11-12
Emit analytics event when en is received

johnschoeman:
Why:
We would like to log events of when Exposure Notifications have been
received.

This commit:
Adds a new event subscriber to the Exposure Notifications event emitter
that will post an analytics event to capture the product usage for this
feature.


devinjameson (approved):

mattThousand (approved):
👍

PR #668, opened 2020-11-12
Update UX for publish consent form

devinjameson:
Why: currently, if the user submits a positive test result code, arrives
at the PublishConsentForm, and hits the "Back" button, they will not be
able to resubmit the same positive test result code. This is confusing
behavior from the perspective of the user. We would like clarify that
exiting the flow will void the positive test result code.

This commit:
- Removes the header back button from the PublishConsentForm screen
- Adds a "Never mind" button to the bottom of the screen
- The "Never mind" button triggers an alert dialog that notifies the
  user that continuing will void their positive test result code and
  gives them the option to cancel
- Updates the global modal header component so that the size of the
  header text does not scale base on the system accessibility text size
  setting. We think this is the best decision because otherwise the
  modal header becomes so large that the screen content is no longer
  visible
- Updates the global primary button style to have a smaller border
  radius due to the button text not being fully visible at very large
  accessibility text sizes with a max border radius

Screen Shot 2020-11-12 at 5 14 36 PM


johnschoeman (approved):
LGTM, but we want to not commit the different assets.

PR #673, opened 2020-11-16
Ios test cleanup

mattThousand:
### This Commit:
Splits iOS test code into logical groups and adds comments to provide context for developers


johnschoeman (approved):
Didnt look too closely at every bit, but looks pretty good to me!

PR #674, opened 2020-11-16
Update localizations

johnschoeman:
Why:
Currently we are not handling localizations for 'Back' navigations or
datetimes from dajs

We would like to have dayjs to be localized correctly when the user
updates the local and for the Back button to be translated.

This commit:
- Updates the navigation header left button to use a translated label.
- Updates the language.ts file to correctly determine the correct IETF
language tag for the current locale and sets the dayjs locale to that.
Note: that some languages the app supports, are not supported by dayjs,
e.g. hmn and so


src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -259,7 +259,7 @@ const style = StyleSheet.create({
     fontSize: Typography.size.x60,
     textAlignVertical: "center",
     textAlign: "center",
-    letterSpacing: 4,
+    letterSpacing: 3,
```
devinjameson, line 262: Legend.

src/locales/languages.ts
```diff
@@ -45,40 +46,15 @@ import zh_Hant from "./zh_Hant.json"
...
-
-export async function setUserLocaleOverride(locale: Locale): Promise<void> {
-  await setLocale(locale)
-  await StorageUtils.setUserLocaleOverride(locale)
-}
+type Resource = Record<Locale.Locale, ResourceLanguage>
 
 /* eslint-disable no-underscore-dangle */
-const AVAILABLE_TRANSLATIONS: Resource = {
+const ALL_LANGUAGE_RESOURCES: Record<Locale.Locale, ResourceLanguage> = {
```
devinjameson: Based on the `type` above it seems like you may have wanted this to be `Resource`.

devinjameson (approved):
LGTM!

PR #679, opened 2020-11-17
Update delete my data screen

devinjameson:
Why: currently the delete my data screen does not accurately describe which data are deleted when the user presses the delete my data button.

This commit: 
- Adds bullet points to describe which data are deleted
- Updates some of the copy that explains what data is deleted
- Conditionally tells the user that the symptom logs and user consent data are deleted only if those features are active in their build
- Shows a confirmation alert if the user presses delete my data to confirm they want to delete their data
- Updates the spec for the delete my data screen to only test if the confirmation alert appears when the user presses the delete my data button

gifdelete


mattThousand (approved):

johnschoeman (approved):
LGTM

PR #681, opened 2020-11-17
Localise authority name, refactor custom copy

johnschoeman:
Why:
We would like for the name of the heath authority to be localised in the
app.

This commit:
Refactors the authorityCopy.ts file into a simpler api in a
useCustomCopy hook such that callers dont need to understand the locale
or 'load' any copy, they can simply use the correct custom copy. We also
added healthAuthorityName to the custom copy.

Note that we structured the logic for custom copy to follow this order
of priority:

1. If the locale and the key for the custom copy exists use that value.

2. If the locale exists but the key for that locale does not default to
the en custom copy key value.

3. If the en custom copy exits, but the key does not exist, return
undefined. (callers are expected to provide the either a default value
or null handling in this case)

4. If the en custom copy does not exist at all return undefined.
(callers are expected to provide the either a default value or null
handling in this case)


mattThousand (approved):

devinjameson (approved):
LGTM!

PR #686, opened 2020-11-18
Update docs to include data dictionary

johnschoeman:
Why:
We would like to transparently document all data that leaves the device.
While this project is open source and all of if this is determinable by
reviewing the source code, we would like to make this inspection process
accessible to those that do not have programming experience.

This commit:
Introduces a data dictionary and product analytics events document. We
also moved the developer setup guide to a stand alone file for better
organization and clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,34 +22,6 @@
 </table>
 <hr />
 
-# Table of contents
-
-- [Project Overview](#project-overview)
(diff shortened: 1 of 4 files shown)


artgibson (approved):
data dictionary lgtm.

PR #687, opened 2020-11-18
Fully remove detox

johnschoeman:
Why:
A previous commit removed a majority of the detox code, but left the
detox npm dependency. While not used in the ultimate binary, currently
this dependency has a registered security vulnerability.

This commit:
Removes the detox dependency and some residual related setup
boilerplate.


devinjameson (approved):
LGTM!

PR #689, opened 2020-11-19
Require minimum phone digits on callback form

devinjameson:
Why: we would like for a minimum number of digits to be required on the phone number input to submit the callback form. Given that the app is used in various countries with different phone number formats, we need to make this number configurable by jurisdiction.


mobile_resources_commit
```diff
@@ -1 +1 @@
-6476f5d54c5397507237160a9b911c2f30136fa6
+9aac1a2d0bed58c6b243bec7c160200cf4b46d4a
```
johnschoeman, line 1: just to double check was this intentional for this pr?

johnschoeman (approved):
LGTM!

mobile_resources_commit
devinjameson: Yes because it relates to the env update.

PR #690, opened 2020-11-19
Fix styling issue with tab bar

devinjameson:
Why: the tab bar icons were appearing too close to one another, which made the labels beneath them difficult to read.

This commit:
-Adds more spacing between the tab bar icons.

Screen Shot 2020-11-19 at 1 29 16 PM


johnschoeman (approved):
LGTM!

PR #700, opened 2020-11-23
Refactor and update behavior of activation flow

devinjameson:
Why: we would like to update the factoring of the activation flow to be simpler and more maintainable. We would also like solve some edge cases in the activation flow that have to do with the interaction between ENs and Bluetooth.

This commit:
- Refactors the activation flow by creating a useActivationNavigation hook that exposes a goToNextScreenFrom function as well as the set of activation steps that should be included. This allows us to colocate the logic for determine the correct activation sequence and introduces no changes in functionality.
- Updates exposure notification status value returned from the native layer to be a single string rather than a tuple with an authorized and enabled state
- Removes the separated concept of "Bluetooth" activation from the app, as the "Bluetooth" status the app cares about is encompassed by the Exposure Notifications Status. If Bluetooth is not on when the user is trying to activate Exposure Notifications, they will see an alert telling them to turn on Bluetooth.


src/Activation/AcceptTermsOfService.tsx
```diff
@@ -26,13 +26,14 @@ import {
...
   const configuration = useConfigurationContext()
+  const { goToNextScreen } = useActivationNavigation()
+
   const [boxChecked, toggleCheckbox] = useState(false)
-  const { t } = useTranslation()
-  const navigation = useNavigation()
 
   const handleOnPressNext = () => {
-    navigation.navigate(ActivationStackScreens.ActivateExposureNotifications)
+    goToNextScreen("AcceptTermsOfService")
```
johnschoeman: what do you think about:
```suggestion
    goToNextScreenFrom("AcceptTermsOfService")
```
?

src/navigation/ActivationStack.tsx
```diff
@@ -1,105 +1,80 @@
...
-    }
-    activationSteps.unshift(anonymizedDataConsent)
-  }
-
-  const activationSummary: ActivationStep = {
-    screenName: ActivationStackScreens.ActivationSummary,
-    component: ActivationSummary,
-  }
-  activationSteps.push(activationSummary)
+  const { activationSteps: inUseActivationSteps } = useActivationNavigation()
```
johnschoeman, line 65: nice

src/navigation/ActivationStack.tsx
```diff
@@ -119,4 +94,23 @@ const ActivationStack: FunctionComponent = () => {
   )
 }
 
+export const applyHeaderLeftBackButton = () => {
```
johnschoeman: would we be able to use the common component for this one? not sure that we should, but iirc we added this to updated the back behavior, which we no longer need to do.

src/Activation/AcceptTermsOfService.tsx
devinjameson: Good call!

src/navigation/ActivationStack.tsx
devinjameson: Ah good catch!

src/Activation/useActivationNavigation.ts
```diff
@@ -0,0 +1,104 @@
...
+  | "ActivateExposureNotifications"
+  | "NotificationPermissions"
+  | "ActivationSummary"
+
+export const toScreen = (step: ActivationStep): ActivationStackScreen => {
+  switch (step) {
+    case "AcceptTermsOfService":
+      return ActivationStackScreens.AcceptTermsOfService
+    case "AnonymizedDataConsent":
+      return ActivationStackScreens.AnonymizedDataConsent
```
johnschoeman: would it make sense to update these names in this pr?

src/Activation/useActivationNavigation.spec.ts
```diff
@@ -0,0 +1,51 @@
+import {
+  determineActivationSteps,
+  Environment,
+} from "./useActivationNavigation"
+
+describe("useActivationNavigation", () => {
+  describe("when location is required and off, bluetooth is off, display accept terms of service is true, product analytics are turned on", () => {
```
devinjameson: missing `and`

src/Activation/useActivationNavigation.spec.ts
```diff
@@ -0,0 +1,51 @@
+import {
+  determineActivationSteps,
+  Environment,
+} from "./useActivationNavigation"
+
+describe("useActivationNavigation", () => {
```
johnschoeman: It might be helpful for future readers to have these specs be more granular.

```suggestion
describe("determineActivationSteps", () => {
  describe("when the locationPermissions are required and off", () => {
     it("includes the activate location step", () => {
     
   describe("when bluetooth is not enabled", () => {
     it("includes the enable bluetooth step", () => {
     
   etc...
```

src/Activation/useActivationNavigation.ts
```diff
@@ -0,0 +1,104 @@
...
+  const isLocationRequired = locationPermissions !== "NotRequired"
+
+  const activationSteps: ActivationStep[] = []
+
+  displayAcceptTermsOfService && activationSteps.push("AcceptTermsOfService")
+  enableProductAnalytics && activationSteps.push("AnonymizedDataConsent")
+  isLocationRequired && activationSteps.push("ActivateLocation")
+  !isBluetoothOn && activationSteps.push("ActivateBluetooth")
+  activationSteps.push("ActivateExposureNotifications")
+  Platform.OS === "ios" && activationSteps.push("NotificationPermissions")
```
johnschoeman, line 96: to keep this from needed to be mocked in a spec it might be nice to pas the os in as part of the env object.

johnschoeman (approved):
Looking pretty good!

src/Activation/useActivationNavigation.spec.ts
devinjameson: Sounds good!

src/Activation/useActivationNavigation.ts
devinjameson: It doesn't appear to need to be mocked. Am I missing something?

src/Activation/useActivationNavigation.ts
devinjameson: Yeah I think so

android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/reactmodules/ExposureNotificationsModule.java
```diff
@@ -51,7 +51,7 @@ public void requestExposureNotificationAuthorization(final Promise promise) {
     FutureCallback<Void> callback = new FutureCallback<Void>() {
       @Override
       public void onSuccess(Void result) {
-        promise.resolve((Util.getEnStatusWritableArray(true)));
+        promise.resolve("Active");
```
johnschoeman: @juancruzgs , we ended up being able to change the api for the NativeModule function to be a just a string, rather than the tuple.  With `Active` being the value that indicates everything is on and `Disabled` would be the other string that could be return.

Is what we have here the correct way to implement this in the android side? If so how would be pass `Disabled` for the failed state, and would it make sense to remove the WritableArray stuff?

ref for the ios enum: https://github.com/Path-Check/gaen-mobile/pull/700/files#diff-4d5a3efb2c0c0f0dcda1b313644b7073911db8069a3578aa98e42733063822fdR109

ios/BT/ExposureManager.swift
```diff
@@ -104,24 +104,35 @@ final class ExposureManager: NSObject {
 
   // MARK: == State ==
 
-  enum EnabledState: String {
-    case enabled = "ENABLED"
+  enum ExposureNoticationStatus: String {
+    case unknown = "UNKNOWN"
```
johnschoeman: What do we think about downcasing all of these? so `Unknown`, to follow the convention we have in the JS layer.

johnschoeman (approved):
Continues to look good to me.

android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/reactmodules/ExposureNotificationsModule.java
```diff
@@ -70,7 +70,7 @@ public void onFailure(@NotNull Throwable exception) {
   public void getCurrentENPermissionsStatus(final Promise promise) {
     ExposureNotificationClientWrapper.get(getReactApplicationContext())
         .isEnabled()
-        .addOnSuccessListener(enabled -> promise.resolve(Util.getEnStatusWritableArray(enabled)))
+        .addOnSuccessListener(enabled -> promise.resolve("Active"))
```
juancruzgs: This should be changed to `promise.resolve(enabled ? "Active" : "Disabled")`

android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/reactmodules/ExposureNotificationsModule.java
juancruzgs: Yes, this is correct.
To improve it you could add constants for "Active" and "Disabled" in `CallbackMessages.java`.

You also have to change `EventSender.sendExposureNotificationStatusChanged` to use those constants and then `getEnStatusWritableArray` can be removed

android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/reactmodules/ExposureNotificationsModule.java
devinjameson: Thanks!

android/app/src/main/java/org/pathcheck/covidsafepaths/bridge/EventSender.kt
```diff
@@ -12,7 +12,7 @@ object EventSender {
     fun sendExposureNotificationStatusChanged(reactContext: ReactContext?, enabled: Boolean) {
         reactContext
             ?.getJSModule(RCTDeviceEventEmitter::class.java)
-            ?.emit(EN_STATUS_CHANGED_EVENT, Util.getEnStatusWritableArray(enabled))
+            ?.emit(EN_STATUS_CHANGED_EVENT, enabled ? CallbackMessages.EN_STATUS_ACTIVE : CallbackMessages.EN_STATUS_DISABLED)
```
devinjameson: @juancruzgs: @johnschoeman and I updated based on your comments. Can you please do one more quick review and confirm this is all set? Thank you!

android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/reactmodules/ExposureNotificationsModule.java
```diff
@@ -51,7 +51,7 @@ public void requestExposureNotificationAuthorization(final Promise promise) {
     FutureCallback<Void> callback = new FutureCallback<Void>() {
       @Override
       public void onSuccess(Void result) {
-        promise.resolve((Util.getEnStatusWritableArray(true)));
+        promise.resolve(enabled ? CallbackMessages.EN_STATUS_ACTIVE : CallbackMessages.EN_STATUS_DISABLED);
```
juancruzgs: This method only calls `onSuccess` if EN was enabled, this line should be:
`promise.resolve(CallbackMessages.EN_STATUS_ACTIVE)`

devinjameson: Thanks @juancruzgs updated. Aside from that is it all good?

juancruzgs: Yes! 💯

juancruzgs (approved):
Android code looks good to me 💯

PR #703, opened 2020-11-24
Remove unused import

mattThousand:


johnschoeman:
@mattThousand , I think we nabbed this fix on #700 already. Can this one be closed?

PR #704, opened 2020-11-24
Improve Onboarding Copy

mattThousand:
#### Why:
Based on user feedback, it has come to our attention that the language "No personally identifiable information ever leaves your phone" is somewhat unclear

#### This commit:
This commit replaces the above copy with "Your personally identifiable information never leaves your phone"


devinjameson (approved):
Love it!

PR #705, opened 2020-11-24
Update navigation on Active Exposure Notifications and Activate Location

devinjameson:
Why: currently, when a user reaches either the Activate Exposure Notifications or Activate Location screen and has already activated the respective service, they will be automatically navigated to the next screen in the activation flow. We would like to instead inform the user that the respective service is turned on and present a button they can use to continue to the next activation step.

This commit:
- Removes the code that navigates the user to the next activation screen automatically
- Conditionally renders either the button to enable the service or a notice that the service is on and a continue button

Screen Shot 2020-11-24 at 3 10 30 PM


johnschoeman (approved):
LGTM! no comments

PR #710, opened 2020-11-26
Push Notification/Linking Update

mattThousand:
#### Why:
We'd like exposure notifications to navigate users to the exposure history screen when tapped.

#### This commit:
This commit specifies the requisite screen for `Linking`
Nov-26-2020 09-59-04


johnschoeman (approved):
LGTM, I wonder if theres a way we could some typesafety around the screen configs? maybe something like an extension to the react-navigation.

Or consider reintroducing detox if we are feeling like we have the resources for it.

PR #717, opened 2020-11-30
Do not subscribe to location events on iOS

devinjameson:
Why: we are currently attempting to subscribe to location status events on iOS, but this concept only exists on Android. Therefore, we have to only subscribe to location status events if the platform in Android.

This commit:
- Updates the `useLocationPermissions` hook to only run a side effect that subscribes to location status events if the platform is Android
- Slightly refactors the setting of the location enabled state to use a ternary to make the code easier to read


mattThousand (approved):

johnschoeman (approved):
LGTM

PR #718, opened 2020-11-30
Fix layout for wide buttons

johnschoeman:
Why:
Currently, wide buttons will allow text that is too long to overflow to
the edge of the button. We would like for buttons to have a consistent
horizontal padding.

|Before|After|
|---|---|
|Simulator Screen Shot - iPhone 11 Pro Max - 2020-11-30 at 09 48 51|Simulator Screen Shot - iPhone 11 Pro Max - 2020-11-30 at 09 48 37|


devinjameson (approved):
Nice! The fix makes sense, I might just double check that `heightBase` and `heightThin` still make sense as names for the view styles.

PR #719, opened 2020-11-30
Remove staging env var

johnschoeman:
Why:
Currently we have a staging environment variable in .env.bt which is
used to determine if error reporting and the debug menu should be shown.
We would like to explicitly enable error reporting and the debug menu
with descriptive configuration variable names, rather than inferring the
behavior based off a notion of 'staging' as we do not have single
staging environment for this project.

This commit:
Removes env.STAGING and adds env.ENABLE_ERROR_REPORTING and
env.DISPLAY_DEBUG_MENU


devinjameson (approved):

mattThousand (approved):
LGTM

PR #720, opened 2020-12-01
Provide better error messages on request gaen api

johnschoeman:
Why:
There are many ways that the request to enable gean permissions for the
app could fail. Most of these response failures mean that the user needs to
manually enable the permissions for the app. We would like to be able to
provide specific instructions to the user on how to enable the
permissions.

this commit:
Refactors the api for the requestAuthorization function in the gaen
native module such that it is more clear what the native layer's
response actually was and to make it easier to respond to the specific
error. We provide a mapping on the native layer for a complete list of
error responses such that we can respond to each case on an individual basis.

Co-Authored-By; Matt Buckley <matt@nicethings.io>

### DIFF
--- a/src/gaen/nativeModule.ts
+++ b/src/gaen/nativeModule.ts
@@ -67,8 +67,6 @@ const toStatus = (data: string): ENPermissionStatus => {
 
 // Permissions Module
 
-type GAENAPIError = "ExceededCheckRateLimit" | "Unknown" | "AppRestricted"
-
 const permissionsModule = NativeModules.ENPermissionsModule
 
 export type RequestAuthorizationResponse =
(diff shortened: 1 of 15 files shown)


src/gaen/nativeModule.ts
```diff
@@ -80,35 +78,86 @@ export type RequestAuthorizationSuccess = {
...
 }
 
 export const requestAuthorization = async (): Promise<
   RequestAuthorizationResponse
 > => {
   try {
-    const status = await permissionsModule.requestExposureNotificationAuthorization()
-    const enStatus = toStatus(status)
+    const enStatus = await permissionsModule.requestExposureNotificationAuthorization()
+    console.log({ enStatus })
```
devinjameson:
```suggestion
```

devinjameson (approved):
LGTM!

PR #721, opened 2020-12-01
Handle rate limit error on check for exposures

johnschoeman:
Why:
Currently we are handling the case where the gaen api returns a rate
limit error when a user is checking for exposures on the exposure
history screen by rendering a 'Something Went Wrong' flash message. This
is not a great user experience. We would like to improve it. Given that
the rate limit is an implementation detail from the perspective of the
user, we would like to show a 'Success' response here as the user did
successfully check for exposures even despite the rate limit.

Furthermore, we would like to better factor this flow to handle error
messages better in general.

This commit:
- Refactors the error handling of the checkForExposures function.
- Renames the checkForExposures function to `detectExposures`.
- Shows a success flash message when the user hits the rate limit on
detectExposures.

Co-Authored-By: Matt Buckely <matt@nicethings.io>

### DIFF
--- a/ios/BT/ExposureManager.swift
+++ b/ios/BT/ExposureManager.swift
@@ -441,9 +441,8 @@ final class ExposureManager: NSObject {
         btSecureStorage.storeExposures(newExposures)
         completionHandler(.success(processedFileCount))
       case let .failure(error):
-        let exposureError = ExposureError.default(error.localizedDescription)
         btSecureStorage.exposureDetectionErrorLocalizedDescription = error.localizedDescription
-        completionHandler(.failure(exposureError))
+        completionHandler(.failure(error))
       }
     }
   }
--- a/ios/BT/ExposureManager.swift
+++ b/ios/BT/ExposureManager.swift
@@ -261,8 +261,9 @@ final class ExposureManager: NSObject {
       switch result {
       case .success:
         resolve(String.genericSuccess)
-      case .failure(let exposureError):
-        reject(exposureError.localizedDescription, exposureError.errorDescription, exposureError)
+      case .failure(let error):
+        let errorString = error._code.enErrorString
+        reject(errorString, error.localizedDescription, error)
       }
     }
   }
--- a/src/ExposureHistory/History/index.spec.tsx
+++ b/src/ExposureHistory/History/index.spec.tsx
@@ -1,4 +1,5 @@
 import React from "react"
+import { Alert } from "react-native"
 import {
   cleanup,
   fireEvent,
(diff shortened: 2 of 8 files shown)


src/ExposureHistory/History/index.spec.tsx
```diff
@@ -134,10 +152,10 @@ describe("History", () => {
...
 
         await waitFor(() => {
-          expect(showMessageSpy).toHaveBeenCalledWith(
-            expect.objectContaining({
-              message: "Something went wrong",
-            }),
+          expect(alertSpy).toHaveBeenCalledWith(
+            "Something Went Wrong",
+            "Something unexpected happened. Please close and reopen the app and try again.",
+            [{ text: "Ok" }],
```
devinjameson:
```suggestion
            [{ text: "Okay" }],
```
Thoughts on using `Okay` vs. `Ok`?

src/locales/en.json
```diff
@@ -32,6 +32,7 @@
     "no_thanks": "No thanks",
     "off": "OFF",
     "on": "ON",
+    "ok": "Ok",
```
devinjameson: Maybe we can use `Okay` here?

devinjameson (approved):
🚢

ios/BT/ExposureManager.swift
```diff
@@ -440,9 +441,8 @@ final class ExposureManager: NSObject {
         btSecureStorage.storeExposures(newExposures)
         completionHandler(.success(processedFileCount))
       case let .failure(error):
-        let exposureError = ExposureError.default(error.localizedDescription)
         btSecureStorage.exposureDetectionErrorLocalizedDescription = error.localizedDescription
-        completionHandler(.failure(exposureError))
+        completionHandler(.failure(error))
```
johnschoeman, line 445: @mattThousand  we noticed a bug where the API ENError was being converted to our code's 'ExposureError' and losing all of the enum info from ENError.  This was our fix, could you give it a double check?

src/ExposureHistory/History/index.tsx
```diff
@@ -55,21 +58,38 @@ const History: FunctionComponent<HistoryProps> = ({
...
-        message: t("common.something_went_wrong"),
-        ...errorFlashMessageOptions,
-      })
+      switch (result.error) {
+        case "RateLimited":
+          showMessage({
+            message: t("common.success"),
+            ...successFlashMessageOptions,
+          })
+          break
```
johnschoeman, line 74:
```suggestion
  case "NotAuthorized":
    showAlert(t("exposure_notification_alerts.not_authored_error_title"), t("exposure_notification_alerts.not_authored_error_body"))
```

src/ExposureHistory/History/index.tsx
```diff
@@ -55,21 +58,38 @@ const History: FunctionComponent<HistoryProps> = ({
...
+        default:
+          showAlert(
```
johnschoeman, line 86:
```suggestion
          ENErrorAlert.showAlert("Unknown")
```

ios/BT/ExposureManager.swift
mattThousand: 👍

PR #723, opened 2020-12-01
Show exposure notification error alerts before exposure check

devinjameson:
Why: if the user does *not* have exposure notifications enabled, the check for exposures will fail. So, before checking for exposures, we would like to ensure the user has exposure notifications enabled.

This commit:
- Checks whether the user has exposure notifications enabled before checking for exposures
- If the user does not have exposure notifications enabled, we tell them that we can't check for exposures unless that have exposure notifications enabled and provide a "Tell Me How" button
- If the user presses the "Tell Me How" button, we give them guidance on how to enable exposure notifications

### DIFF
--- a/src/ExposureHistory/History/index.tsx
+++ b/src/ExposureHistory/History/index.tsx
@@ -29,6 +29,8 @@ import {
   Affordances,
   Iconography,
 } from "../../styles"
+import { usePermissionsContext } from "../../Device/PermissionsContext"
+import { useRequestExposureNotifications } from "../../useRequestExposureNotifications"
 
 type Posix = number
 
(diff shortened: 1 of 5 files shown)


johnschoeman (approved):
Nice work!

src/ExposureHistory/History/index.tsx
```diff
@@ -56,6 +62,25 @@ const History: FunctionComponent<HistoryProps> = ({
...
+      t("exposure_notification_alerts.cant_check_for_exposures_title"),
+      t("exposure_notification_alerts.cant_check_for_exposures_body"),
+      [
+        {
+          text: t("common.back"),
+          style: "cancel",
+        },
+        {
+          text: t(
+            "exposure_notification_alerts.cant_check_for_exposures_button",
```
johnschoeman, line 76: what do you think about, `Enable Exposure Notifications` here? in the case that the request succeeds, 'Tell me how' would end up not being very clear as no new dialog would pop up.

devinjameson: How could we be in a state where the request succeeds? I think the only other think that could happen is the native enable ENs alert pops up.

PR #724, opened 2020-12-02
Handle SMS long code verification links

tokumori:
Why:
Currently the code does not handle SMS links containing the verification long code (e.g., `appname://v?r=[region]&c=[longcode]`, `https://app.associateddomain.org/v?c=[longcode]`, etc.)

This commit:
- Opens the app when following a link from an associated domain.
- Navigates to the code input form if path is `/v`, and auto populates the code field if `c=[longcode]` parameter exists. The long code cannot be modified, so the user must continue or go back. Long code will not persist if the user goes back.
- Resets routes and navigates to `App` route if `Home` route does not exist. This prevents strange behavior and re-renders if `App` route is not the first route (index 0).
- Fixes back button when taken directly to code input form from link.
- Uses env variable `DEEP_LINK_PREFIXES` to make it customizable for different HAs.
- Adds basic deep link handler in `AndroidManifest.xml` and `BT-*.entitlements` files with `https://app.associateddomain.org/` as the customizable dummy field for each HA.

Additional requirements:
HAs using Universal Links/Android App Links must upload the proper files to `https://<domain>/.well-known/apple-app-site-association` and `https://<domain>/.well-known/assetlinks.json`.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -41,6 +41,18 @@
             android:launchMode="singleTask"
             android:screenOrientation="portrait"
             android:windowSoftInputMode="adjustResize">
+            <intent-filter >
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <!-- Accepts URIs that begin with "https://app.associateddomain.org/” -->
+                <!-- Change android:host field to a registered domain -->
+                <data 
+                    android:scheme="https"
+                    android:host="app.associateddomain.org"
+                    android:pathPrefix="/" 
+                  />
+            </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
 
--- a/example.env.bt
+++ b/example.env.bt
@@ -70,3 +70,6 @@ ANDROID_APPLICATION_ID='org.pathcheck.pc.bt'
 ANDROID_PLAY_STORE_URL=
 IOS_APP_STORE_URL=
 IOS_BUNDLE_ID='org.pathcheck.bt'
+
+// Deep Linking
+DEEP_LINK_PREFIXES=["pathcheck://", "https://app.associateddomain.org"]
--- a/ios/AppDelegate.m
+++ b/ios/AppDelegate.m
@@ -158,6 +158,13 @@ fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler
 {
   return [RCTLinkingManager application:application openURL:url options:options];
 }
+- (BOOL)application:(UIApplication *)application continueUserActivity:(nonnull NSUserActivity *)userActivity
+ restorationHandler:(nonnull void (^)(NSArray<id<UIUserActivityRestoring>> * _Nullable))restorationHandler
+{
+ return [RCTLinkingManager application:application
+                  continueUserActivity:userActivity
+                    restorationHandler:restorationHandler];
+}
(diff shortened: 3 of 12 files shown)


johnschoeman (commented):
Looking like a good start, thanks for opening this! We noticed some failing tests so @devinjameson and I are thinking of opening a branch against yours to fix them up. We might need to change some of the factoring in the routing logic with React Navigation to make this work. Any thoughts on this, @tokumori ?

tokumori:
Opening a branch against mine and fixing the errors sounds good. The `AffectedUserContext must be used with a provider` unit test will pass if the `toHome` function is moved from the `AffectedUserFlowContext.tsx` to both the `PublishConsentForm.tsx` and `Complete.tsx` files. I can tweak that and push it up.

The other unit tests look like issues with `route.params` and `navigation.dispatch` not being in the related `CodeInputForm.spec.tsx` and `Complete.spec.tsx` files, so there probably needs to be some test writing involved there.

Re: refactoring routing logic, the issue is that the `AffectedUserStack` is on the same level as `<Stack.Screen name={"App"} component={MainTabNavigator} />`. I decided to just reset the navigation state to what it would be if a user opened the app normally rather than nesting `AffectedUserStack`.

Normally on open (i.e., user presses the app icon), the app's initial route will be "App", all the names on the same level and rendered component names (e.g., AffectedUserStack, Home, etc.) will be created in the router/navigation object, and the "Home" route will exist so we can navigate to it.

Opening the app through a link with the long code goes directly to `AffectedUserStack` and bypasses the initial "App" route, so "Home" is never added to the navigation object. While `navigation.navigate("App")` is an option, the parent routes array looked something like, `[{"name": "AffectedUserStack"}, {"name": "App"}]` instead of `[{"name": "App"}, {"name": "AffectedUserStack"}]`. The reset makes it so the app starts from the "opened from app icon" state rather than "opened from URL" state.

johnschoeman:
> Opening a branch against mine and fixing the errors sounds good. The `AffectedUserContext must be used with a provider` unit test will pass if the `toHome` function is moved from the `AffectedUserFlowContext.tsx` to both the `PublishConsentForm.tsx` and `Complete.tsx` files. I can tweak that and push it up.
> 
> The other unit tests look like issues with `route.params` and `navigation.dispatch` not being in the related `CodeInputForm.spec.tsx` and `Complete.spec.tsx` files, so there probably needs to be some test writing involved there.
> 
> Re: refactoring routing logic, the issue is that the `AffectedUserStack` is on the same level as `<Stack.Screen name={"App"} component={MainTabNavigator} />`. I decided to just reset the navigation state to what it would be if a user opened the app normally rather than nesting `AffectedUserStack`.
> 
> Normally on open (i.e., user presses the app icon), the app's initial route will be "App", all the names on the same level and rendered component names (e.g., AffectedUserStack, Home, etc.) will be created in the router/navigation object, and the "Home" route will exist so we can navigate to it.
> 
> Opening the app through a link with the long code goes directly to `AffectedUserStack` and bypasses the initial "App" route, so "Home" is never added to the navigation object. While `navigation.navigate("App")` is an option, the parent routes array looked something like, `[{"name": "AffectedUserStack"}, {"name": "App"}]` instead of `[{"name": "App"}, {"name": "AffectedUserStack"}]`. The reset makes it so the app starts from the "opened from app icon" state rather than "opened from URL" state.

Ended up opening a new pr to avoid coordinating between different forks: https://github.com/Path-Check/gaen-mobile/pull/728

What do you think about this solution? We tend to try to avoid interacting with react-navigation routing directly as it's usually quite brittle, (and hard to understand imo). By updating the backButton header component to accept an onPress callback, we can keep all of the conditional logic in the component and by pass the need to manually manage the route state.

PR #725, opened 2020-12-02
Handle location services disabled when requesting ENs

devinjameson:
Why: some Android devices require location services to be enabled for exposure notifications to work. If someone is on such a device and tries to enable exposure notifications with location services disabled, would like to show an alert that tells them to enable location services.

This commit:
- Adds `LocationOffAndRequired` to the list of possible states for the `ENPermissionStatus`
- If `ENPermissionStatus` is `LocationOffAndRequired`, show an alert to the user telling them to enable location services

Screen Shot 2020-12-02 at 2 22 53 PM


src/locales/en.json
```diff
@@ -257,8 +257,8 @@
     "eula": "End User Legal Agreement",
     "eula_agree_terms_of_use": "I agree that I have read and accepted the terms of use",
     "i_am_over": "Yes, I am over {{minimumAge}}",
-    "location_alert_body": "To use Bluetooth scanning, the device location setting needs to be turned on.",
-    "location_alert_header": "Enable Location access from \"{{applicationName}}\"",
+    "location_alert_body": "To use Exposure Notifications, Location Services must be enabled",
+    "location_alert_header": "Enable Location Services",
```
devinjameson, line 243: @juancruzgs Can you confirm this copy is correct?

src/useRequestExposureNotifications.tsx
```diff
@@ -13,46 +13,34 @@ export const useRequestExposureNotifications = (): (() => void) => {
...
-      default:
-        showBaseExposureNotificationsAlert()
-    }
-  }
-
-  const showBaseExposureNotificationsAlert = () => {
-    if (Platform.OS === "ios") {
-      showShareExposureInformationAlert()
+    if (response.kind === "success") {
+      switch (response.status) {
```
devinjameson: @mattThousand I'm confused as to why `BluetoothOff` is under the "success" column here, but I figured `LocationOff` should go here as well. Should this be refactored so that anything but `Active` is considered a "failure"?

These types seem weird to me considering the overlap:

```
type RequestAuthorizationError =
  | "IOSUnknown"
  | "BadParameter"
  | "NotEntitled"
  | "NotAuthorized"
  | "Unsupported"
  | "Invalidated"
  | "BluetoothOff"
  | "InsufficientStorage"
  | "NotEnabled"
  | "APIMisuse"
  | "Internal"
  | "InsufficientMemory"
  | "RateLimited"
  | "Restricted"
  | "BadFormat"
  | "DataInaccessible"
  | "TravelStatusNotAvailable"
  | "Unknown"

type ENPermissionStatus =
  | "Unknown"
  | "Active"
  | "Disabled"
  | "BluetoothOff"
  | "LocationOff"
  | "Restricted"
  | "Paused"
  | "Unauthorized"
```

src/locales/en.json
juancruzgs: Looks good to me 👍

src/useRequestExposureNotifications.tsx
mattThousand: @devinjameson I agree the overlap is confusing; there are 2 separate concepts at play; one is the "status" of the `Exposure Notifications` framework, for which `BluetoothOff` is a valid state (neither a "success" nor a failure persay) and the other concept is the error that we might get back when attempting to activate ENs. 

For example, the `EN` framework could be in the `BluetoothOff` state, in which case if we try to activate it, we will get a `BluetoothOff` error response.

johnschoeman (approved):
Code looks good to me.

src/useRequestExposureNotifications.tsx
johnschoeman: perhaps it makes sense to update the api for `requestAuthorization` to fulfill with only a 'success' or reject with the error?

mattThousand (dismissed):
LGTM, let’s hold off on un-drafting until the android side is pushed

src/useRequestExposureNotifications.tsx
mattThousand: good points, I agree that the overlap is confusing. I think the intention in the current construction is to model the OS api; on the `iOS` side our JS method ultimately calls through to `setExposureNotificationEnabled` (https://developer.apple.com/documentation/exposurenotification/enmanager/3583729-setexposurenotificationenabled), which returns a `success` even in the case when `exposureNotificationStatus` is `bluetoothOff`; we _could_ move this special casing of the `bluetoothOff` scenario to the native layer, and vend a failure up to the JS layer in that case. Thoughts?

mattThousand: https://github.com/Path-Check/gaen-mobile/pull/732

src/Device/PermissionsContext.tsx
```diff
@@ -46,12 +43,14 @@ export type ENPermissionStatus =
   | "Active"
   | "Disabled"
   | "BluetoothOff"
+  | "LocationOffAndRequired"
```
devinjameson, line 53: @mattThousand, changed the name of this because it clarifies that `LocationOffAndNotRequired` is not a possible state for `ENPermissionStatus`. We may want to change this on the native layer in some places, too?

devinjameson: @juancruzgs This also may be relevant to the work you are doing tomorrow ^

PR #727, opened 2020-12-02
Adjust iOS scoring threshold.

mattThousand:
#### Why:
We'd like the scoring threshold check to reflect the original intention and be consistent across platforms.

#### This commit:
This commit brings the scoring threshold check for devices running `iOS 13.7+` in line with the Android implementation; the android implementation appropriately converts the the `weightedDurationSum` to minutes before making the comparison (https://github.com/Path-Check/gaen-mobile/blob/79e4d0091417fb71533edd3195d049b243be1fcd/android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/ExposureNotificationClientWrapper.java#L229), and this commit performs the identical check in reverse, converting the `triggerThresholdWeightedDuration` value to seconds to perform the check


johnschoeman (approved):

emilianojuang:
LG Matt, thanks for fixing this

PR #728, opened 2020-12-03
Introduce universal links for verification codes

johnschoeman:
Why:
We would like to eventually support universal links with verification codes in the query params that automatically open the app at the input verification code form and populate the form with the code. Currently, we support a url scheme of `pathcheck://`, but we also want to support universal links with a customizable domain in the future.

This commit:
- Adds the basic logic to support deep links navigating to the code input form and configures react-navigation to navigate to the correct screen. The app will navigate to the code input form given links with the `/v?c=[InputCode]` resource.

Note:
- A future PR should add the CI/CD logic for entitlements to include registering associated domains for universal links.

Co-Authored-By: Bryce Saito <bsaito88@gmail.com>
Co-Authored-By: John Schoemam <johnschoeman1617@gmail.com>

### DIFF
--- a/src/Activation/useActivationNavigation.ts
+++ b/src/Activation/useActivationNavigation.ts
@@ -61,7 +61,7 @@ export const useActivationNavigation = (): ActivationNavigation => {
 
   const activationSteps = determineActivationSteps(environment)
 
-  const goToNextScreenFrom = (currentStep: ActivationStep) => {
+  const goToNextScreenFrom = async (currentStep: ActivationStep) => {
     const currentStepIndex: number | undefined = activationSteps.findIndex(
       (step) => step === currentStep,
     )
@@ -71,9 +71,8 @@ export const useActivationNavigation = (): ActivationNavigation => {
       const nextScreen = toScreen(nextStepName)
       navigation.navigate(nextScreen)
     } else {
-      completeOnboarding().then(() => {
-        navigation.navigate("App", { screen: Stacks.Home })
-      })
+      await completeOnboarding()
+      navigation.navigate("App", { screen: Stacks.Home })
     }
   }
 
--- a/src/Activation/useActivationNavigation.ts
+++ b/src/Activation/useActivationNavigation.ts
@@ -4,7 +4,11 @@ import { useNavigation } from "@react-navigation/native"
 import { useConfigurationContext } from "../ConfigurationContext"
 import { usePermissionsContext } from "../Device/PermissionsContext"
 import { LocationPermissions } from "../Device/useLocationPermissions"
-import { ActivationStackScreen, ActivationStackScreens } from "../navigation"
+import {
+  ActivationStackScreen,
+  ActivationStackScreens,
+  Stacks,
+} from "../navigation"
 import { useOnboardingContext } from "../OnboardingContext"
 
 type ActivationStep =
@@ -67,7 +71,9 @@ export const useActivationNavigation = (): ActivationNavigation => {
       const nextScreen = toScreen(nextStepName)
       navigation.navigate(nextScreen)
     } else {
-      completeOnboarding()
+      completeOnboarding().then(() => {
+        navigation.navigate("App", { screen: Stacks.Home })
+      })
     }
   }
 
--- a/src/AffectedUserFlow/AffectedUserContext.tsx
+++ b/src/AffectedUserFlow/AffectedUserContext.tsx
@@ -48,21 +48,13 @@ export const AffectedUserProvider: FunctionComponent<AffectedUserProviderProps> 
   }
 
   const navigateOutOfStack = () => {
-    if (isOnboardingComplete) {
-      navigation.dispatch(
-        CommonActions.reset({
-          index: 0,
-          routes: [{ name: "App" }],
-        }),
-      )
-    } else {
-      navigation.dispatch(
-        CommonActions.reset({
-          index: 0,
-          routes: [{ name: WelcomeStackScreens.Welcome }],
-        }),
-      )
-    }
+    const route = isOnboardingComplete ? "App" : WelcomeStackScreens.Welcome
+    navigation.dispatch(
+      CommonActions.reset({
+        index: 0,
+        routes: [{ name: route }],
+      }),
+    )
   }
 
   return (
--- a/src/AffectedUserFlow/AffectedUserContext.tsx
+++ b/src/AffectedUserFlow/AffectedUserContext.tsx
@@ -4,8 +4,11 @@ import React, {
   useState,
   useContext,
 } from "react"
+import { useNavigation } from "@react-navigation/native"
 
 import { ExposureKey } from "../exposureKey"
+import { useOnboardingContext } from "../OnboardingContext"
+import { Stacks, WelcomeStackScreens } from "../navigation"
 
 type Token = string
 type Key = string
@@ -16,6 +19,7 @@ interface AffectedUserContextState {
   exposureKeys: ExposureKey[]
   setExposureKeys: (keys: ExposureKey[]) => void
   setExposureSubmissionCredentials: (certificate: Token, hmacKey: Key) => void
+  navigateOutOfStack: () => void
 }
 
 export const AffectedUserContext = createContext<
@@ -23,6 +27,9 @@ export const AffectedUserContext = createContext<
 >(undefined)
 
 export const AffectedUserProvider: FunctionComponent = ({ children }) => {
+  const { isOnboardingComplete } = useOnboardingContext()
+  const navigation = useNavigation()
+
   const [exposureKeys, setExposureKeys] = useState<ExposureKey[]>([])
   const [hmacKey, setHmacKey] = useState<Key | null>(null)
   const [certificate, setCertificate] = useState<Token | null>(null)
@@ -35,6 +42,14 @@ export const AffectedUserProvider: FunctionComponent = ({ children }) => {
     setHmacKey(hmacKey)
   }
 
+  const navigateOutOfStack = () => {
+    if (isOnboardingComplete) {
+      navigation.navigate("App", { screen: Stacks.Home })
+    } else {
+      navigation.navigate(WelcomeStackScreens.Welcome)
+    }
+  }
+
   return (
     <AffectedUserContext.Provider
       value={{
@@ -43,6 +58,7 @@ export const AffectedUserProvider: FunctionComponent = ({ children }) => {
         exposureKeys,
         setExposureKeys,
         setExposureSubmissionCredentials,
+        navigateOutOfStack,
       }}
     >
       {children}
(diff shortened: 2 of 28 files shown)


src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -54,16 +63,21 @@ const CodeInputForm: FunctionComponent = () => {
...
-  const codeLengthMin = 6
-  const codeLengthMax = 16
-  const codeIsInvalidLength =
-    code.length < codeLengthMin || code.length > codeLengthMax
-  const codeContainsNonAlphanumericChars = (code: string) =>
-    !code.match(/^[a-zA-Z0-9]*$/)
+  const linkCode: string | undefined = route?.params?.c || route?.params?.code
+  useEffect(() => {
+    if (linkCode) {
+      setCode(linkCode)
```
johnschoeman: @devinjameson , ended up moving to this as the `API.postCode` call uses the component state for the code and keeping a separate `linkCode` and `userInputedCode` made that somewhat awkward.  I think this is another vote for the make a bespoke screen/form for this use case, but this seems to work with the minimum amount of new logic.

src/navigation/MainNavigator.tsx
```diff
@@ -51,11 +52,19 @@ const settingsStackTransitionPreset = Platform.select({
   android: TransitionPresets.DefaultTransition,
 })
 
+const customPrefixes = env.DEEP_LINK_PREFIXES?.split(",") || []
+const allPrefixes = ["pathcheck://", "https://*.en.express/", ...customPrefixes]
```
johnschoeman, line 56: Per the notes from our contact at apple, it looks like the idea is to support universal links that integrate nicely with en express with this format: "https://appname.en.express".  Not 100% sure as we are not able to test locally without a real url to use, but we are hoping that 'react-navigation' will correctly handle this wildcard as that will save us the need to configure these hosts at build time.  Either way, I think it makes sense to take a pass at this in a future commit.

johnschoeman:
Also, @devinjameson , we should think through what happens if the GAEN permissions are not granted to the app as navigating to the code input screen in this state currently renders the `you need to enable gaen to submit a code` content and not the code input form. Which will likely cause some user confusion.

tokumori:
I like the changes, but the user is unable to navigate Home if they open the app from a Universal Link with this code.

Back button on `CodeInputForm`: 
Screenshot_20201202-182817_PathCheck BT  

Never mind on `PublishConsent`:
Screenshot_20201202-182922_PathCheck BT  

Done on `Complete`:
Screenshot_20201202-182956_PathCheck BT

Error message:
Screenshot_20201202-182847_PathCheck BT

This issue only occurs if the user goes through the code input from a Universal Link since the "Home" route name doesn't exist. Using `navigation.navigate("App")` would take the user back to the Home screen, but doing a one-time reset if route params exist (i.e., user opened the app with a URL and verification code) cleans up the routes and mimics what the app is like when opened regularly.

I would like to avoid directly interacting with react-navigation as well, so I'm open to any other ideas.

tokumori:
I like the changes, but the user is unable to navigate Home if they open the app from a Universal Link with this code.

#### Back button on `CodeInputForm`: 
  

#### Never mind on `PublishConsent`:


#### Done on `Complete`:


#### Error message:


This issue only occurs if the user goes through the code input from a Universal Link since the "Home" route name doesn't exist. Using `navigation.navigate("App")` would take the user back to the Home screen, but doing a one-time reset if route params exist (i.e., user opened the app with a URL and verification code) cleans up the routes and mimics what the app is like when opened regularly.

I would like to avoid directly interacting with react-navigation as well, so I'm open to any other ideas.

src/AffectedUserFlow/CodeInput/CodeInputForm.tsx
```diff
@@ -54,16 +63,21 @@ const CodeInputForm: FunctionComponent = () => {
   const [errorMessage, setErrorMessage] = useState(defaultErrorMessage)
   const [isFocused, setIsFocused] = useState(false)
 
-  const codeLengthMin = 6
-  const codeLengthMax = 16
-  const codeIsInvalidLength =
-    code.length < codeLengthMin || code.length > codeLengthMax
-  const codeContainsNonAlphanumericChars = (code: string) =>
-    !code.match(/^[a-zA-Z0-9]*$/)
+  const linkCode: string | undefined = route?.params?.c || route?.params?.code
```
devinjameson: Why `route?.params?.c`? Shouldn't the key be `code`?

devinjameson (commented):
> I like the changes, but the user is unable to navigate Home if they open the app from a Universal Link with this code.
> 
> This issue only occurs if the user goes through the code input from a Universal Link since the "Home" route name doesn't exist. Using `navigation.navigate("App")` would take the user back to the Home screen, but doing a one-time reset if route params exist (i.e., user opened the app with a URL and verification code) cleans up the routes and mimics what the app is like when opened regularly.
> 
> I would like to avoid directly interacting with react-navigation as well, so I'm open to any other ideas.

@johnschoeman and @tokumori What do you think of the current solution?

src/OnboardingContext.tsx
```diff
@@ -36,6 +36,7 @@ export const OnboardingProvider: FunctionComponent<OnboardingProviderProps> = ({
   const completeOnboarding = () => {
     StorageUtils.setIsOnboardingComplete()
     setIsOnboardingComplete(true)
+    return Promise.resolve()
```
johnschoeman: What do you think about using async await here? Not sure if this will guarantee that StorageUtils.setIsOnboardingComplete() will finish before the chained then will resolve with this set up, if thats the intent here.

src/OnboardingContext.tsx
```diff
@@ -33,8 +33,8 @@ export const OnboardingProvider: FunctionComponent<OnboardingProviderProps> = ({
     userHasCompletedOnboarding,
   )
 
-  const completeOnboarding = () => {
-    StorageUtils.setIsOnboardingComplete()
+  const completeOnboarding = async () => {
```
devinjameson, line 36: This look right @johnschoeman?

src/Activation/useActivationNavigation.ts
```diff
@@ -67,7 +71,8 @@ export const useActivationNavigation = (): ActivationNavigation => {
       const nextScreen = toScreen(nextStepName)
       navigation.navigate(nextScreen)
     } else {
-      completeOnboarding()
+      await completeOnboarding()
```
devinjameson, line 74: Also this @johnschoeman?

src/OnboardingContext.tsx
devinjameson: Makes sense!

tokumori:
> @johnschoeman and @tokumori What do you think of the current solution?

Thanks for working on this @devinjameson !

Two issues from a UI/UX perspective. One is if the URL has something other than `v?c=[longcode]` or `v?code=[longcode]` for its verification-code key, the user won't be able to go back. I doubt this is going to be an issue since we were told to use this format, and would think other HAs would be told this as well.

The second is if the user presses the back button and then goes back into the `AffectedUserFlowStack`, pressing the back button again in the `AffectedUserStart` screen takes them "forward" to `CodeInput` instead of `Home`. Below is some console logs of the navigation.


Something that's not visible to the user is the same console.logs will run in the `Exposures` tab. This will happen if the user backs out or completes the code submission. There's probably an underlying issue with the `ExposureHistoryStack`, but this could be happening because the route params are not cleared.

devinjameson:
> > @johnschoeman and @tokumori What do you think of the current solution?
> 
> Thanks for working on this @devinjameson !
> 
> Two issues from a UI/UX perspective. One is if the URL has something other than `v?c=[longcode]` or `v?code=[longcode]` for its verification-code key, the user won't be able to go back. I doubt this is going to be an issue since we were told to use this format, and would think other HAs would be told this as well.
> 
> The second is if the user presses the back button and then goes back into the `AffectedUserFlowStack`, pressing the back button again in the `AffectedUserStart` screen takes them "forward" to `CodeInput` instead of `Home`. Below is some console logs of the navigation.
> 
> 
> Something that's not visible to the user is the same console.logs will run in the `Exposures` tab. This will happen if the user backs out or completes the code submission. There's probably an underlying issue with the `ExposureHistoryStack`, but this could be happening because the route params are not cleared.
> 

Thanks @tokumori! Will take a deeper look and continue to work through. I actually just came across these issues when running on my device. Appreciate the thorough explanation and insight. :)

src/AffectedUserFlow/AffectedUserContext.tsx
```diff
@@ -49,9 +49,19 @@ export const AffectedUserProvider: FunctionComponent<AffectedUserProviderProps>
 
   const navigateOutOfStack = () => {
     if (isOnboardingComplete) {
-      navigation.navigate("App", { screen: Stacks.Home })
+      navigation.dispatch(
```
devinjameson, line 56: @tokumori Does this solve the issue you noticed?

devinjameson: @johnschoeman Also want to make you aware of this update. The intention is to reset the navigation state to avoid any funky behavior (instead of just navigating), which I believe was @tokumori's original intention here.

tokumori: @devinjameson Yes, this solves those issues! Avoiding the funky behavior was exactly my intention with the reset.

johnschoeman: LGTM! nice work y'all

johnschoeman:
also, @tokumori , wanted to note that we ended up needing to pull out the changes to the Info.plist for the associated domains as that needs to be coordinated with updating some provisioning profiles in our CI/CD process. We plan to add this back in a future commit.

Thanks again for getting this started and helping getting it over the finish line! Do you have any other thoughts on the PR?

tokumori:
> also, @tokumori , wanted to note that we ended up needing to pull out the changes to the Info.plist for the associated domains as that needs to be coordinated with updating some provisioning profiles in our CI/CD process. We plan to add this back in a future commit.
> 
> Thanks again for getting this started and helping getting it over the finish line! Do you have any other thoughts on the PR?

No problem, @johnschoeman . That was more to help other HAs know where/what to edit more than anything else since I had to figure that out for the Hawaii app.

Only other thing I can think of is navigation always resets when `navigateOutOfStack` is called versus just when there are route params, but I can't think of that being an issue.

tokumori (approved):
- [x] Opens the app when following a link from an associated domain.

- [x] Navigates to the code input form if path is `/v`, and auto populates the code field if `c=[longcode]` parameter exists.

- [x] The long code cannot be modified, so the user must continue or go back.

- [x] Long code will not persist if the user goes back/exits the stack.

- [x] Resets routes and navigates to "App" route when trying to leave the stack. This prevents strange behavior.

- [x] `navigateOutOfStack` is accessible to all screens in the stack.

- [x] Uses env variable DEEP_LINK_PREFIXES to make it customizable for different HAs.

devinjameson:
> Only other thing I can think of is navigation always resets when `navigateOutOfStack` is called versus just when there are route params, but I can't think of that being an issue.

Just want to note that we addressed this so we shouldn't be resetting the route anymore unless we navigate in from a link.

devinjameson (approved):

PR #729, opened 2020-12-03
Add tr locale

stefanoschrs:
#### Why:
Because we are fetching the language file from lokalise but it's not processed


johnschoeman (approved):
LGTM, ~should there be a tr.json file included?~ nope, as your comment mentioned we already have this in the repo!

PR #732, opened 2020-12-04
Treat `BluetoothOff` as Error when requesting to enable Exposure Notifications (iOS)

mattThousand:
#### Why:
When requesting to enable exposure notifications from the JS layer, we'd like to treat responses other than an `Active` status as errors.

#### This commit:
This commit updates the iOS exposure enablement call to vend a `success` to the JS layer only if the `exposureNotificationStatus` is `Active` after the request, and vends and error otherwise


devinjameson (approved):
Nice nice looks good!

PR #733, opened 2020-12-07
Introduce config for different verification flows

johnschoeman:
Why:
We would like to support multiple strategies for users submitting
verification codes.

This commit:
Introduces the environment variables and basic config logic to
conditionally render the start verification flow buttons on the
dashboard. Note that a future commit will start to introduce the new
verification flow logic.

|SimpleVerification|EscrowVerification|
|---|---|
|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-07 at 06 26 31|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-07 at 06 28 14|


mattThousand (approved):

devinjameson (approved):
Looks good! Would it make any sense to have something like VERIFICATION_FLOW in the env that is either “simple” or “escrow”?

johnschoeman:
> Looks good! Would it make any sense to have something like VERIFICATION_FLOW in the env that is either “simple” or “escrow”?

I think we could, I went with this way to keep things slightly simpler and faster to implement as we get to keep just true/false. of course this gives the possibility of having both or neither enabled, so a future commit might want to introduce the union type approach.

Though now that you mention it, I'm thinking this might be worth getting done correctly the first time, will make the update.

PR #736, opened 2020-12-07
Phone Number/Diagnosis Key Submission for Escrow Server (iOS)

mattThousand:
#### Why:
We'd like to submit phone numbers and keys to an escrow verification server if a given HA's server architecture dictates 

#### This commit:
This commit implements the network calls necessary to submit a user's phone number and a positively diagnosed user's keys to an escrow server.


johnschoeman (approved):
LGTM

PR #744, opened 2020-12-09
Update UserDetailsFrom style for escrow verification flow

devinjameson:
Why: we would like for the user details form to provide a good UX on both iOS and Android.


src/EscrowVerification/UserDetailsForm.tsx
```diff
@@ -45,20 +47,26 @@ export const phoneToFormattedString = (phonenumber: string): string => {
...
+  const [isPhoneInputFocused, setIsPhoneInputFocused] = useState(false)
   const [errorMessage, setErrorMessage] = useState(defaultErrorMessage)
+  const [showDatePickerAndroid, setShowDatePickerAndroid] = useState(false)
 
-  const handleOnChangeTestDate = (testDate: Date | undefined) => {
+  const handleOnChangeTestDate = (
+    _event: Event,
+    testDate: Date | undefined,
+  ) => {
+    setShowDatePickerAndroid(false)
```
devinjameson, line 69: This line seems weird, but there's a really UI bug that occurs if it's not included. Found in the issues for the date picker library.

johnschoeman: Could we make sure to mention this in the commit message?

johnschoeman (approved):
LGTM!

src/EscrowVerification/UserDetailsForm.tsx
johnschoeman: and add the issue link here as well?

johnschoeman: https://github.com/react-native-datetimepicker/datetimepicker/issues/263
https://github.com/react-native-datetimepicker/datetimepicker/issues/54

PR #746, opened 2020-12-09
Show notification status on home screen if off

devinjameson:
Why: we would like to tell the user if their notifications are off.

This commit:
Adds a card to the home screen that informs the user if their
notifications are off. If the card is pressed and notification
permissions have not been previously denied, the app will request
notification permissions from the OS. If notifications have been
previously denied, the app will open the app settings.

Note:
`notificationPermissionsStatusFromString` was updated to
replace `unknown` with `unavailable` to match what the call to
`checkNotifications` can return.

gif


devinjameson:
@mattThousand Running into some funky behavior here. The notification status seems to start off as `Denied`, and to activate notifications, we need to request permissions. However, if the user previously had notifications on and then turned them off, requesting notification permissions will do nothing. In that case, we need to send the user to the app settings to manually turn on notifications. However, I don't know how we can tell on the JS side which route to take (request notification permissions or send the user to app settings) because in either case the notification status is `Denied`. Thoughts?

mattThousand:
> @mattThousand Running into some funky behavior here. The notification status seems to start off as `Denied`, and to activate notifications, we need to request permissions. However, if the user previously had notifications on and then turned them off, requesting notification permissions will do nothing. In that case, we need to send the user to the app settings to manually turn on notifications. However, I don't know how we can tell on the JS side which route to take (request notification permissions or send the user to app settings) because in either case the notification status is `Denied`. Thoughts?

Check out the flow chart below, which describes how permissions behave on iOS - my reading is that you would get `Denied` both in the case in which a user has not requested notification permissions previously _and_ in the case where they have, but they've also gone into settings and disabled notifications. Are you getting the same status in both cases?
  
Screen Shot 2020-12-10 at 12 04 33 PM

src/Home/index.tsx
```diff
@@ -79,6 +82,42 @@ const Home: FunctionComponent = () => {
   )
 }
 
+const NotificationsOff = () => {
+  const { t } = useTranslation()
+  const { notification } = usePermissionsContext()
+
+  if (notification.status === "Granted") {
+    return null
```
johnschoeman: What do you think about having a single return?

johnschoeman (approved):
LGTM, one minor comment

PR #749, opened 2020-12-10
fix typos en copy

artgibson:
#### Why:

en typos and copy updates

#### This commit:

changes en local file


mattThousand (approved):
LGTM!

PR #753, opened 2020-12-14
Update bundler version and gems

johnschoeman:
Why:
Currently our builds our failing due to bundler versions being out of
sync with the CI box action `Run ruby/setup-ruby@v1` default bundler.

This commit:
Updates bundler to 2.2.1
Updates fastlane version to the most recent of 2.170.0
Updates the rest of the gem dependencies.


mattThousand (approved):
Thanks for fixing this!

PR #760, opened 2020-12-14
Refactor Exposure History screen

devinjameson:
Why: we would like to better inform users about their exposure information.

This commit:
Introduces two major changes. 
Firstly, this commit moves the most relevant information from the exposure details screen to the exposure history screen. We decided to do this because the exposure detail screen currently does not provide any additional info on the exposure and always has the same content regarding next steps and health recommendations, regardless of the exposure. For this reason, it makes more sense to include this information on the main exposure history screen if the user has had any exposures within the past 14 days.
Secondly, this commit introduces an exposure summary at the top of the exposure history screen, which informs the user how many days remain in their recommended quarantine period if they have a recent exposure.
Fixes an issue that caused the Exposure History screen to render more often than necessary, by applying `useCallback` to the `refreshExposureInfo` function in the `ExposureContext`.

|Before|After|
|----|----|
|Simulator Screen Shot - iPhone 12 Pro - 2020-12-14 at 18 09 28|Simulator Screen Shot - iPhone 12 Pro - 2020-12-14 at 18 09 07|


johnschoeman (approved):
LGTM

PR #761, opened 2020-12-15
Introduce SymptomOnsetDate screen

devinjameson:
Why: we would like to be have the option to include the symptom onset date in the positive test result submission.

This commit:
Adds a `SymptomOnsetDate` screen to the `AffectedUserFlow` that can be activated with a build flag
The `SymptomOnsetDate` either collects a date or null if the user was asymptomatic. This value is saved to the context. Nothing else is done with this value. A future PR will add the symptom onset date to the network request if it exists.
Adds the `react-native-static-safe-area-insets` package, which enables us to get the device's insets without using a hook.

Note: the yellow color on the iPhone simulator isn't in the actual app, there's just something off with my simulator.

gif

### DIFF
--- a/src/AffectedUserFlow/AffectedUserContext.tsx
+++ b/src/AffectedUserFlow/AffectedUserContext.tsx
@@ -11,6 +11,7 @@ import { Stacks, WelcomeStackScreens } from "../navigation"
 
 type Token = string
 type Key = string
+type Posix = number
 
 export interface AffectedUserContextState {
   certificate: Token | null
@@ -21,6 +22,8 @@ export interface AffectedUserContextState {
   navigateOutOfStack: () => void
   linkCode: string | undefined
   setLinkCode: (linkCode: string | undefined) => void
+  symptomOnsetDate: Posix | null
+  setSymptomOnsetDate: (symptomOnsetDate: Posix | null) => void
 }
 
 interface AffectedUserProviderProps {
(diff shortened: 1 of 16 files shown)


src/AffectedUserFlow/AffectedUserContext.tsx
```diff
@@ -41,6 +44,9 @@ export const AffectedUserProvider: FunctionComponent<AffectedUserProviderProps>
   const [hmacKey, setHmacKey] = useState<Key | null>(null)
   const [certificate, setCertificate] = useState<Token | null>(null)
   const [linkCode, setLinkCode] = useState<string | undefined>(undefined)
+  const [symptomOnsetDate, setSymptomOnsetDate] = useState<Posix | null>(
+    Date.now(),
```
johnschoeman, line 48: should this default to null? (probably okay either way as we'll figure out what makes sense for the data model when we wire up the networking side)

src/AffectedUserFlow/SymptomOnsetDate.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const handleOnPressDateInput = () => {
+    setShowDatePickerAndroid(true)
+  }
+
+  const handleOnChangeTestDate = (
+    _event: Event,
+    symptomOnsetDate: Date | undefined,
+  ) => {
+    setShowDatePickerAndroid(false)
+    symptomOnsetDate && setSymptomOnsetDate(dayjs(symptomOnsetDate).valueOf())
```
johnschoeman:
```suggestion
    if (symptomOnsetDate) {
       const posix = dayjs(symptomOnsetDate).valueOf()
       setSymptomOnsetDate(posix)
     }
```

johnschoeman (approved):
LGTM, some minor questions

src/AffectedUserFlow/AffectedUserContext.tsx
devinjameson: It depends. In that case we'd probably want to have some local state on the component and then update this when the user continues. I'll probably leave for now and we can cross that bridge in the next bit of work.

src/AffectedUserFlow/SymptomOnsetDate.tsx
devinjameson: 👍

PR #763, opened 2020-12-16
Fix typo in en deep link domain

johnschoeman:
Why:
We currently have an extra `/` at the end of a the wildcard en link for
universal links to the app.

This commit:
Removes the extra '/'


mattThousand (approved):
LGTM!

PR #772, opened 2020-12-18
Update EN configuration defaults

mattThousand:
Updating our default EN configurations, relying on https://github.com/lfph/gaen-risk-scoring/blob/d06da88fe1912529f0e6c6e6c0c3ca91bd5d93fc/risk-scoring.md as reference


artgibson (approved):
lgtm. matches model

Issue #711, opened 2020-11-27
Translations pull script issue on Linux


stefanoschrs:
## Issue type
Localization

## Steps to reproduce
Run `yarn i18n:pull` without lokalise2 installed

## Expected Behavior
Expected to work but instead you get `./src/locales/pull.sh: line 38: lokalise2: command not found` because the lokalise2 binary is installed on the relative `./bin` which is not in the PATH.

If you check the [install.sh](https://github.com/lokalise/lokalise-cli-2-go/blob/master/install.sh#L26) it's an intended behavior, so maybe add a message or something to the pull script to tell the user to move the binary somewhere in PATH or do it automatically.

johnschoeman:
Thanks for opening! Doesn't look like anyone has been available to automate this, but adding a note for future developers here: https://github.com/Path-Check/gaen-mobile/pull/776

PR #776, opened 2020-12-21
Add note to src/locales/pull.sh

johnschoeman:
Why:
If a user of the i18n:pull script doesn't have lokalise2 available the
script will fail. We would like to include a note to explain this to
readers of the script.

Reference: https://github.com/Path-Check/gaen-mobile/issues/711


mattThousand (approved):
👍

PR #779, opened 2020-12-21
Use explict applinksdomain and update build script

johnschoeman:
Why:
Currently the universal link for enx is failing to route to the correct
app.

This commit:
Updates the applink domain in the xcodeproj capabilities to not use the
wildcard character to register the domain. That is, we switched
`applinks:*.en.express` for `applinks:xx-xx.en.express`. As a
consequence we needed to update the build scripts to set the correct
domain.

We also update the logging for the build scripts to be more readable.

|before|after|
|---|---|
|Screen Shot 2020-12-21 at 12 39 50 PM|Screen Shot 2020-12-21 at 11 58 27 AM|


mattThousand (approved):
LGTM

PR #781, opened 2020-12-21
Add next steps tap to analytics events

johnschoeman:
Why:
When a user that has opted into product analytics we would like to
record the tap event of going to 'NextSteps' from the exposure history
screen.

This commit:
Adds a call to trackEvent for the NextSteps button tap.


mattThousand (approved):
LGTM!

PR #786, opened 2020-12-23
Verify app links (Android)

juancruzgs:
This commit reads the EN Express domain from the env file and adds a new `intent-filter` to open the app once an EN Express link is clicked.

**Important note:** The app link will only work on release builds since the signature has to match the SHA256 present in `assetlinks.json`. In debug builds it will show a chooser to let the user select which app should open the link

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -162,6 +162,9 @@ android {
 
         //Fix problem in bt where .env.bt is not recognized
         resValue "string", "build_config_package", "org.pathcheck.covidsafepaths"
+
+        // Handle EN Express deep links
+        manifestPlaceholders = [enxDomain: project.env.get("ENX_APPLINKS_DOMAIN") ?: "*.en.express"]
     }
     splits {
         abi {
--- a/bin/configure_builds.sh
+++ b/bin/configure_builds.sh
@@ -81,8 +81,8 @@ end
 
 def replace_string_in_file(file_path:, regex:, value:)
   file_contents = File.read(file_path)
-  new_contents = file_contents.gsub(regex, value)
-  if new_contents.match(value)
+  if file_contents.match(regex)
+    new_contents = file_contents.gsub(regex, value)
     File.open(file_path, 'w') { |f| f.write new_contents }
   else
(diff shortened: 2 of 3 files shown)


juancruzgs:
@johnschoeman Should it redirect to a specific screen?
It just opens the app currently

android/app/build.gradle
```diff
@@ -162,6 +162,9 @@ android {
 
         //Fix problem in bt where .env.bt is not recognized
         resValue "string", "build_config_package", "org.pathcheck.covidsafepaths"
+
+        // Handle EN Express deep links
+        manifestPlaceholders = [enxDomain: project.env.get("ENX_APPLINKS_DOMAIN") ?: "*.en.express"]
```
artgibson, line 167: any issues with this wildcard being the default? i.e if an app doesnt need deeplinks

mattThousand (approved):

android/app/build.gradle
juancruzgs: It will show a chooser dialog and the user can select to open the link in our app or Google Chrome (for example).
Since it cannot verify the app is linked to the website it won't be opened automatically.

That's the only thing that will happen when using a wildcard

johnschoeman:
> @johnschoeman Should it redirect to a specific screen?
> It just opens the app currently

Are you using a link with params? e.g. `xx-xx.en.express/v?c=12341234` ? that should direct directly to the code input form.

bin/configure_builds.sh
```diff
@@ -178,6 +178,26 @@ def update_bundle_identifiers
...
+  puts "Adding android app link with domain #{domain}"
+  replace_string_in_file(
+    file_path: './android/app/build.gradle',
+    regex: /manifestPlaceholders (.+)/,
+    value: "manifestPlaceholders = [enxDomain: \"#{domain}\"]"
+  )
+end
+
+APP_LINKS_DOMAIN = "ENX_APPLINKS_DOMAIN"
+def update_app_links
```
johnschoeman, line 191: 💯

johnschoeman (approved):
LGTM

PR #808, opened 2020-12-31
Add Realm Migration to Account for New Stored Property

mattThousand:
#### Why:
Without migrating realm to account for the new `weightedDurationSum` property, the local database will be reset when users update the the new version

#### This commit:
This commit updates the `schemaVersion` to trigger the necessary realm migration


artgibson (approved):
lgtm

PR #809, opened 2020-12-31
Improve Error Logging for Escrow Verification Server Failures

mattThousand:
#### Why:
We'd like to have better insight into errors from requests to the escrow verification server during development.

** Note that logging is only present in staging and debug builds, and is not present in release builds

Before:
`failed to submit phone number:  [Error]`

After:
`failed to submit phone number:  {"code": "403", "message": "", "nativeStackAndroid": [], "userInfo": null}`


artgibson (approved):
lgtm

PR #810, opened 2021-01-04
copy updates

artgibson:
#### Why:

copy updates based on feedback


johnschoeman (approved):

src/locales/en.json
```diff
@@ -219,10 +219,11 @@
     "not_feeling_well": "Not feeling well?",
     "notifications_off": "Turn on notifications to get notified if you receive an exposure",
     "open_healthcheck": "Open Healthcheck",
```
johnschoeman: You might be able to run `yarn i18n:extract` to remove unused entries. (or perhaps at least manually remove this one as its no longer being used.)

PR #811, opened 2021-01-05
Update callback api to use refresh token

johnschoeman:
Why:
Currently the callback form uses a username password based oauth
strategy. We would like to use different strategy using a refresh token.

This commit:
Updates the callbackAPI.ts file to use the new oauth strategy.


src/Callback/callbackAPI.tsx
```diff
@@ -1,78 +1,107 @@
...
     LA_Last_Name__c: lastname,
     LA_Mobile_Phone__c: phoneNumber,
   }
 
+  const endpoint = formUrl
+
   try {
-    const oauthResponse = await postOAuth()
-    const data = await oauthResponse.json()
+    const oauthResonse = await postGetNewAccessToken()
```
artgibson: small typo here oauthResonse -> oauthResponse, this carries down

artgibson (approved):
lgtm

PR #813, opened 2021-01-06
Downgrade okhttp version to 3.14.9

juancruzgs:
#### Why:

There are some conflicts between the okhttp version that RN is using and the one that the native side uses.
That is causing this crash https://app.bugsnag.com/pathcheck/gaen-mobile/errors/5fe50b4618ec5000177f2861?filters[event.since][0]=30d&filters[error.status][0]=open&event_id=5ff4f7d20065958e9a020000&pivot_tab=event

#### This commit:

Downgrades the okhttp version to avoid having conflicts with RN


artgibson:
i was able to submit the contact form and submission of a verification code

artgibson (approved):
worked on test builds, lgtm.

PR #815, opened 2021-01-07
Update icon used for the share link

johnschoeman:
Why:
Currently the icon used for the share link is named based off the
content. We would like for this image to be configurable, so the name
should be changed to be based on the usage rather than the content.

This commit:
Replaces `HugEmoji.png` with `ShareImage.png`, which is the standard
material UI 'share' graphic. We also remove some unused iconography
styles.

|before|after|
|---|---|
|Simulator Screen Shot - iPhone 12 - 2021-01-07 at 11 47 30|Simulator Screen Shot - iPhone 12 - 2021-01-07 at 11 47 40|


artgibson (approved):
lgtm

PR #819, opened 2021-01-12
Use custom copy for callback form instruction

johnschoeman:
Why:
We would like to provide the ability to set custom copy for the callback
form instructions.


artgibson (approved):
lgtm

PR #820, opened 2021-01-12
Introduce support number

johnschoeman:
Why:
We would like to provide the option render a support contact number on
the callback form and in the 'How do I get a verification code?' section
of the verification code info screen. When a user taps this number the
device should open the call app with this number loaded.

This commit:
Introduces a new env variable, SUPPORT_PHONE_NUMBER, which when provided
will render a link for that number on the callback form and the more
info screen

|CallbackForm|MoreInfo|
|---|---|
|Simulator Screen Shot - iPhone 12 Pro - 2021-01-12 at 12 28 09|Simulator Screen Shot - iPhone 12 Pro - 2021-01-12 at 12 38 23|


artgibson (approved):
lgtm.

PR #821, opened 2021-01-13
Update covid recommendation text

johnschoeman:
Why:
We would like to update the content of the covid recommendation screen.

This commit:
Updates the content on the covid recommendation screen and moves the
logic for parsing the quarantine length environment variable to the
conficonfigurationContext.

|Before|After|
|---|---|
|Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-13 at 10 41 21|Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-13 at 10 43 39|


artgibson (approved):
Approving with the change request to create a new environment variable that handles symptom quarantine length or you can remove the environment variable usage.

SYMPTOM_QUARANTINE_LENGTH, which is 10 days at default and different from Exposure quarantine length which is 14 days.

From HA:
If a user says they have symptoms; they should isolate -> leading to a 10 day duration.    
If a user gets an exposure notification, they should be advised to do a 14 day quarantine.

johnschoeman:
> Approving with the change request to create a new environment variable that handles symptom quarantine length or you can remove the environment variable usage.
> 
> SYMPTOM_QUARANTINE_LENGTH, which is 10 days at default and different from Exposure quarantine length which is 14 days.
> 
> From HA:
> If a user says they have symptoms; they should isolate -> leading to a 10 day duration.
> If a user gets an exposure notification, they should be advised to do a 14 day quarantine.

~~Got it, thanks for the clarification, I'll make the updates.~~

Edit: discussed IRL, we are going to have the addition of a symptom quarantine length introduced in a future PR.

PR #822, opened 2021-01-13
Fix bug with quarantine end date

johnschoeman:
Why:
Currently we have a bug where we incorrectly calculate the end date of
a recommended quarantine for users that have received an exposure
notification.

This commit:
Determines the last day of quarantine by adding quarantine days
remaining to the current day instead of the date of the exposure itself.

### DIFF
--- a/src/ExposureHistory/History/ExposureSummary.tsx
+++ b/src/ExposureHistory/History/ExposureSummary.tsx
@@ -50,10 +50,10 @@ const ExposureSummary: FunctionComponent<ExposureSummaryProps> = ({
     Date.now(),
     exposure.date,
   )
-  const quarantineEndDate = dayjs(exposureEndDate)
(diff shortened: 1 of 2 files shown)


artgibson (approved):
The calculation looks accurate , but it would continue to set the quarantineEndDate=Today when it reaches 0. This would likely require some UI modifications.

If the Quarantine period is over, I'd expect the UI to either show the the last day I needed to quarantine (exposure + days of quarantine) or hide the date.

johnschoeman:
> The calculation looks accurate , but it would continue to set the quarantineEndDate=Today when it reaches 0. This would likely require some UI modifications.
> 
> If the Quarantine period is over, I'd expect the UI to either show the the last day I needed to quarantine (exposure + days of quarantine) or hide the date.

good catch! It probably makes sense to not render the quarantine recommendation at all once the quarantine is over.

PR #823, opened 2021-01-13
Update callback form styling

johnschoeman:
Why:
We would like to use a better phone input styling for the callback form

|Before|After|
|---|---|
|Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-13 at 13 34 42|Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-13 at 13 30 19|


artgibson (approved):
lgtm

PR #824, opened 2021-01-14
Check for Existing Keys on Device Before Escrow Submission

mattThousand:
#### Why:
We'd like to avoid making requests to the escrow verification server containing empty key payloads (the server will return a 4xx).

#### This commit:
This commit short circuits POSTs to the escrow verification server if there are no existing keys on the device, and provides instructions for the user to wait until their device has had time to generate its first keys.
IMG_2496
IMG_2495

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Set XCode Version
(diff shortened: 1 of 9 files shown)


artgibson (approved):
lgtm

johnschoeman (approved):
LGTM

PR #829, opened 2021-01-15
Jts update quarantine summary

johnschoeman:
Why:
We would like to update the recommended quarantine section on the
exposure history screen to be more direct and to remove some ambiguity
in what 'stay quarantined through means'

This commit:
Updates the language of the recommended quarantine section to use fewer
words. We also change the notion of `quarantine through:` to `last day:`

Note:
We would like for days remaining to say 1 when the current day
is the last day shown on the card.
We would like the first day of quarantine to be the day after the
exposure range end date.
So the time line for a 14 quarantine is determined as follows:

recorded exposure date : 1/2/2021
expected exposure range : 1/1/2021 - 1/3/2021
first day of quarantine: 1/4/2021 (14 days remaining)
last day of quarantine: 1/17/2021 (1 day remaining)

|Before|After|
|---|---|
|Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-15 at 12 57 56|Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-15 at 12 00 40|

### DIFF
--- a/src/ExposureHistory/History/ExposureSummary.spec.tsx
+++ b/src/ExposureHistory/History/ExposureSummary.spec.tsx
@@ -46,9 +46,9 @@ describe("determineRemainingQuarantine", () => {
 
       expect(result1).toEqual(0)
       expect(result2).toEqual(0)
-      expect(result3).toEqual(1)
-      expect(result4).toEqual(4)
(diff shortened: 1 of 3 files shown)


artgibson (approved):
lgtm. thanks!

PR #832, opened 2021-01-19
Update image size on activation summary.

johnschoeman:
Why:
We would like to better support custom images on the activation summary
screen.

This commit:
Updates the image styling to have a 1:1 aspect ratio such that custom
images for this screen dont have to match the exact same image
dimensions, but just be about 1:1.

We also update the mobile resources commit.


artgibson (approved):
lgtm

PR #833, opened 2021-01-21
Handle dont share action in escrow verification

johnschoeman:
Why:
During the submit your phone number step of the Escrow Verification
flow, an OS level prompt asks if the user would like to share their
devices Random IDs. If the user chooses 'Dont Share' the os will raise
an error which we are not handling.

This commit:
- Updates the error handling for the submitPhoneNumber iOS native method
to return ENDomainErrors when the os returns them.

- Updates the javascript side error handling for the submitPhoneNumber
native method. Note that the api for the native function can return both
network error and ENDomainErrors which do not follow the same error
format. We introduced a `handleNetworkErrror()` and
`handleENDomainError()` to handle these two cases. A future refactor
might consider updating the native method to better separate the error
conditions.

|GIF|
|---|
|RPReplay_Final1611251740|

### DIFF
--- a/ios/EscrowVerification/Manager/EscrowVerificationManager.swift
+++ b/ios/EscrowVerification/Manager/EscrowVerificationManager.swift
@@ -25,8 +25,8 @@ class EscrowVerificationManager: NSObject {
             completionHandler(nil)
           }
         }
-      default:
-        break
+      case let .failure(error):
+        completionHandler(error)
       }
     }
   }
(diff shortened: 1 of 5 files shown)


artgibson (approved):
could this error check also be added to the verification code screen?

johnschoeman:
> could this error check also be added to the verification code screen?

Yes, good catch, the next screen has the same issue.

PR #835, opened 2021-01-22
Address some accessibility issues

johnschoeman:
Why:
Currently the app has a few non WCAG 2.0 AA compliant issues.

This commit:
- Adds accessibility labels to several images and buttons
- Increases the contrast on the Delete My Data Button.
- Moves the tappable target for the Dashboard CTA cards from being the
entire card to just the graphical indicator on the card so that the
'MoreInfo' button on the card is accessible to screen readers.


artgibson (approved):
lgtm.

PR #836, opened 2021-01-22
GAEN-685 Return error when there are no keys to submit

juancruzgs:
#### Why:

We user was seeing a Success message when he doesn't have keys to submit

#### This commit:

Returns a Failure instead to let the user know we didn't have anything to send to the backend

### DIFF
--- a/android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/network/escrowserver/EscrowVerificationClient.kt
+++ b/android/app/src/main/java/org/pathcheck/covidsafepaths/exposurenotifications/network/escrowserver/EscrowVerificationClient.kt
@@ -16,6 +16,7 @@ import kotlinx.coroutines.withContext
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
 import org.pathcheck.covidsafepaths.BuildConfig
+import org.pathcheck.covidsafepaths.exposurenotifications.reactmodules.EscrowVerificationKeySubmissionModule.Companion.NO_KEYS_ERROR_CODE
 import org.pathcheck.covidsafepaths.exposurenotifications.storage.ExposureNotificationSharedPreferences
 import org.pathcheck.covidsafepaths.exposurenotifications.utils.Error
 import org.pathcheck.covidsafepaths.exposurenotifications.utils.Result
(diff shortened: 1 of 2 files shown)


juancruzgs:
@artgibson Could you help me test this branch?

johnschoeman (approved):
LGTM

juancruzgs:
I haven't tested it because I'll need a VPN and a phone number
If someone could test it quickly please let me know if it works okay

artgibson (commented):
@juancruzgs  I tested a build with this, but was unable to get a "correct" error message that matched No Keys on Device. Only received "Something Went Wrong, Try Again".

My test included:

1. Deleting app and removing random ids from device
2. Re-installing app fresh and then submitting a phone number.


@johnschoeman do we know if this error from android will get bubbled properly to react native layer to show these messages: https://github.com/Path-Check/gaen-mobile/blob/develop/src/locales/en.json#L530

juancruzgs:
We are sending the error code "999" to the JS layer
It may be worth debugging this method in JS to check what's happening

```
export const submitDiagnosisKeys = async (
  verificationCode: string,
  date: Posix,
): Promise<SubmitKeysResponse> => {
  try {
    await escrowVerificationKeySubmissionModule.submitDiagnosisKeys(
      verificationCode,
      date,
    )
    return { kind: "success" }
  } catch (e) {
    Logger.error(`failed to submit verification code: `, { ...e })
    switch (e.code) {
      case "999":
        return { kind: "failure", error: "NoKeysOnDevice" }
      default:
        return { kind: "failure", error: "Unknown" }
    }
  }
}
```

johnschoeman:
@artgibson , @juancruzgs , we are handling this in the view layer: https://github.com/Path-Check/gaen-mobile/blob/develop/src/EscrowVerification/VerificationCodeForm.tsx#L101.

I would agree with Juan in that the issue is probably with `submitDiagnosisKeys`, likely something is off with the error shape and its defaulting to 'Unknown' rather than 'NoKeysOnDevice'

Im happy to put a commit onto this branch if you'd like Juan, but I think this pr could be shipped as is and i can take a look at fixing the JS side in a future commit.

artgibson (approved):
lgtm. we can review react-native layer for error message. cc @johnschoeman

juancruzgs:
Sounds good to me

PR #837, opened 2021-01-22
Introduce covid data webview

johnschoeman:
Why:
We would like to support webviews for covid data dashboards.

This commit:
Introduces a new optional feature, CovidDataWebView, when enabled an
provided a webivew url it will render a 'CovidData' button which will
navigate the user to a screen with the webview.

Kapture 2021-01-22 at 13 00 30


artgibson (approved):
lgtm

PR #838, opened 2021-01-27
(iOS) Remove wildcard domain from app domains

mattThousand:
#### Why:

The wildcard app domain doesn’t add anything, and could cause unintended consequences

#### This commit:

Removes wildcard domain from app domains


artgibson (approved):
lgtm. ran 2 test builds with sucess.

johnschoeman (approved):
TY!

PR #840, opened 2021-01-28
Url building

stefanoschrs:
#### Why:
- There is a missing space after the `Terms of Use` description, resulting the link to be stuck on it.
- The url building in `fetchCustomCopy` (and I believe this should be applied in all the places) needs some improvement to ignore/ add trailing slash. In the `example.env.bt` the `REMOTE_CONTENT_URL` is without trailing slash so when you append `content/v1/copy.json` on it, it results in a bad url.


johnschoeman (approved):
Nice catch, looks good to me.

PR #842, opened 2021-01-28
GAEN-656 Create notification channel for background notifications

juancruzgs:
#### Why:

We have multiple user requests for a way to silence the periodic “checking for exposures” message. The current configuration respects the normal notification silencing setting on phones, but people want a separate way to silence just the COVIDaware MN “checking for exposures” notification (independent from an actual “exposure notification” and the overall phone settings).

#### This commit:

Creates a new notification channel (Exposure Check) that gives the user the option to disable "Checking exposures..." notifications when the background job is running.

#### Screenshots:

device-2021-01-28-123812

#### How to test:

- Disable "Exposure Check" notifications from `Settings > Apps & notifications > App > App Info > Notifications `
- Tap on Check for exposures in the app
- Verify the notification is not shown

#### Linked issues:

https://pathcheck.atlassian.net/browse/GAEN-656


johnschoeman (approved):
LGTM

PR #843, opened 2021-01-28
Allow custom copy for how to get a code section

johnschoeman:
Why:
Currently the Verification Code More Info screens have three sections:
What is a verification code, How do I get a verification code, and what
happens when I submit a verification code. The first two sections allow
for customizable copy, we would like for the third section to also allow
for custom copy.

This commit:
Adds the copy for the VerificationCodeOnSubmitInfo to the custom copy
object and adds it to the More Info Screens.


artgibson (approved):
lgtm

PR #844, opened 2021-01-29
Handle 429 rate limit code

johnschoeman:
Why:
During the escrow verification code flow if the escrow server returns a
429 rate limit, we are only informing the user that 'something went
wrong'. We would like to explicitly let the user know they hit the rate
limit set by the server. Additionally we are treating 403's as rate
limits rather than 'Forbidden' as defined by the http status code spec.

This commit:
Adds the 429 code to the error handling flow and remaps a 403 to the
correct status.


artgibson (approved):
lgtm.

PR #847, opened 2021-02-02
Accessibility updates

mattThousand:
#### This commit: 
This commit adds a high contrast border around the `EN` card. After the update, the contrast between the white background and the border is > 4.5, as is the contrast between the black text and the card's background.
Screen Shot 2021-02-02 at 1 25 13 PM
Screen Shot 2021-02-02 at 1 24 24 PM


artgibson (approved):
lgtm

PR #849, opened 2021-02-09
Update code input form for accessibility.

mattThousand:
#### Why:
Contrast of the placeholder does not meet 4.5 to 1 minium

#### This commit:
This commit uses higher contrast colors for the `textinput` border and placeholder to bring it in line with `WCAG`

### Before
Simulator Screen Shot - iPhone 12 - 2021-02-09 at 15 14 03

### After
Simulator Screen Shot - iPhone 12 - 2021-02-09 at 15 15 40


artgibson (approved):
lgtm

PR #850, opened 2021-02-09
Focus verification code input when user arrives on screen (accessibility)

mattThousand:
#### Why:
Screen reader does not know there is a Code textbox because the initial page focus is on the Next Button (WCAG 2.4.3)

#### This commit:
This commit shifts focus to the verification code input field when the user first arrives on screen. The timeout is necessary to avoid announcing the back button first each time the user arrives on this screen.


artgibson (approved):
lgtm

PR #851, opened 2021-02-11
Use System Dialog for Manual Exposure Detection Completion Event

mattThousand:
#### Why:
We'd like to announce the completion of manual exposure detection to visually impaired users.

#### This commit:
This commit replaces the existing toast with a system dialog that reliably announces the completion event.

IMG_2595
IMG_2596


artgibson (approved):
lgtm

PR #852, opened 2021-02-11
Only send exposure notification received analytics event if new exposure is found

mattThousand:
### This commit:
This commit fixes a bug whereby the `en_notification_received` event was being sent on every exposure detection cycle.


artgibson (approved):
lgtm

PR #853, opened 2021-02-17
Add titles to untitled screens for accessibility 

mattThousand:
## Why:
We'd like the app to be accessible for visually impaired users

### This commit:
This commit adds titles to previously untitled screens to bring the UI in line with `WCAG` guidelines.

IMG_2635
IMG_2634
IMG_2633
IMG_2632


artgibson (approved):
lgtm

PR #856, opened 2021-02-23
Remove --poll flag from i18n script

mattThousand:
### Why:
Running `yarn i18n:push` results in an "unknown flag" error on the command line due to the `--poll` flag.

### This Commit:

This commit removes the `--poll` flag from the script.


artgibson (approved):
lgtm.

PR #861, opened 2021-02-26
Store crash reporting API key in private repository

mattThousand:
### This Commit:
This commit removes the hardcoded api key and instead pulls it in from a private repository.


artgibson (approved):
lgtm

PR #864, opened 2021-03-02
UI Bugfix

mattThousand:
### This Commit:
This commit fixes a mistake from a previous commit where a string key was left in place of the intended localized string.

Before:
Simulator Screen Shot - iPhone 11 - 2021-03-01 at 20 53 46

After: 
Simulator Screen Shot - iPhone 11 - 2021-03-01 at 20 53 53


artgibson (approved):
lgtm

PR #866, opened 2021-03-03
Gaen 686 - Prompt for airplane mode events

lundjrl:
Why:
GAEN-686

This commit:
If we cannot complete an exposure check due to network issues, I throw a DataInaccessible error and alert the user to turn off airplane mode and turn on wifi or data.

Screenshots:
How to test:
Open the app, navigate to "exposures" in the bottom tab bar, turn on airplane mode, and press "check for exposures" you should get a modal that prompts your to turn off airplane mode.

Linked issues:
GAEN-686


artgibson (approved):
lgtm

PR #867, opened 2021-03-04
Tuik verification code screen link

lundjrl:
#### Why:

If a user closes out the app, gets their code, and reopens the app, they have to enter their phone number again.

#### This commit:

Adds button to screen before phone number verification screen.

#### Linked issues:

---


artgibson (approved):
lgtm

PR #868, opened 2021-03-05
Ag en network copy

artgibson:
#### Why:

update network copy to be a broader message to check the user has internet


lundjrl:
> #### Why:
> update network copy to be a broader message to check the user has internet

Looks good @artgibson

lundjrl (approved):
Looks good

PR #869, opened 2021-03-05
Filter Exposures to Last 14 Days

mattThousand:
### Why:
We'd like to display only exposures from the last 14 days on the exposure history screen

### This commit:
This commit filters exposures fetched from the local database for only those that occurred within the last 2 weeks

### Before:

Simulator Screen Shot - iPhone 11 - 2021-03-05 at 10 14 34

### After:
Simulator Screen Shot - iPhone 11 - 2021-03-05 at 10 14 25


artgibson (commented):
suggestion to move filter code into the getCurrentExposures method instead of toExposureInfo, which seems to be used in a few places where filtering may be unintended.

see: https://github.com/Path-Check/gaen-mobile/blob/9a9e46a1c603215514fcf9ef13de7bfac3e1c55a/src/gaen/nativeModule.ts#L152

mattThousand:
> suggestion to move filter code into the getCurrentExposures method instead of toExposureInfo, which seems to be used in a few places where filtering may be unintended.
> 
> see:
> 
> https://github.com/Path-Check/gaen-mobile/blob/9a9e46a1c603215514fcf9ef13de7bfac3e1c55a/src/gaen/nativeModule.ts#L152



> suggestion to move filter code into the getCurrentExposures method instead of toExposureInfo, which seems to be used in a few places where filtering may be unintended.
> 
> see:
> 
> https://github.com/Path-Check/gaen-mobile/blob/9a9e46a1c603215514fcf9ef13de7bfac3e1c55a/src/gaen/nativeModule.ts#L152

Filtering is _always_  intended 😃, which is why the initial filter implementation was done in the native layer. Ideally the JS layer will only ever see exposures from the filtered timeframe.

artgibson (approved):
lgtm

PR #875, opened 2021-03-11
Changed nextsteps timestamp to be the number of days since the exposure.

lundjrl:
#### Why:

We need to pass the number of days since an exposure rather than the timestamp to our dashboard. This is just a cheap way to do so on the client. 

I just compare the timestamp to the current timestamp, get the difference of days, then pass that like normal.

#### Linked issues:

---


lundjrl:
@artgibson  is this what you were looking for? I just parsed the exposure date created to be a difference in the days since the exposure. Lmk if changes need to be made and how to communicate that to the server.

artgibson (approved):
lgtm

PR #876, opened 2021-03-12
Onboarding UX Fix

mattThousand:
### Why:
When onboarding to the app on an iPhone 7 form factor after hitting "enable exposure notifications" the user needs to scroll down in order to "continue".

### This Commit
This commit programmatically scrolls to the bottom of the screen so that the "continue" button is visible when the user first arrives.

Mar-11-2021 21-07-37


artgibson (approved):
lgtm.

PR #878, opened 2021-03-16
Gaen 319

lundjrl:
#### Why:

This just makes the text of our empty-key error more user-friendly.

#### Screenshots:

I couldn't actually reproduce this on a physical android. No error showed up when following the ticket steps.
But essentially:
This:  "unable to validate diagnosis verification: calculating expected HMAC: cannot calculate hmac on empty exposure keys."
Is now this: "Looks like you have no exposure keys, this means that your device hasn't generated any yet. Please wait an hour for them to be generated."

#### Linked issues:

---


artgibson (approved):
lgtm. ideally a future commit to localize this string.

PR #879, opened 2021-03-16
GAEN-645, checking for phone numbers used every 5 minutes in callbac…

lundjrl:
…k form

#### Why:

Users can repeatedly spam the callback form with their phone number, this solution makes a timestamp on the first submit and saves the phone number. Only after 15 minutes can that phone number be reused on the same device.

#### Linked issues:

---

### DIFF
--- a/src/Callback/Form.tsx
+++ b/src/Callback/Form.tsx
@@ -14,6 +14,7 @@ import {
   TouchableOpacity,
 } from "react-native"
 import { useNavigation } from "@react-navigation/native"
+import AsyncStorage from "@react-native-community/async-storage"
 
 import { useStatusBarEffect, CallbackStackScreens } from "../navigation"
 import { useCustomCopy } from "../configuration/useCustomCopy"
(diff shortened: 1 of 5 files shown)


lundjrl:
Note: This build fails one of the "describe" jest tests in Form.spec.tsx

src/Callback/Form.tsx
```diff
@@ -74,33 +75,84 @@ const CallbackForm: FunctionComponent = () => {
...
       Alert.alert(t("errors.something_went_wrong"), e.message)
       setIsLoading(false)
     }
   }
 
+  /* Wait for 15 minutes to pass before submitting the same phonenumber. */
+  const initWaitForResubmit = async (phoneNumber: string) => {
+    const now = new Date()
+    const nowStr = now.getTime().toString()
+    await AsyncStorage.setItem("LastPhoneNumber", phoneNumber)
```
artgibson: @lundjrl let's not store the phone number, cap the number of successful attempts to 2 per 15 minutes. Update error messages to handle that also.

lundjrl: @artgibson  This should work without storing the phone number in local storage. We just check to see if the user has completed two submissions in the past 15 minutes and throw an error if so.

lundjrl: @artgibson  can you approve so we can merge into develop?

src/Callback/Form.tsx
```diff
@@ -74,33 +75,78 @@ const CallbackForm: FunctionComponent = () => {
...
       }
-      setIsLoading(false)
     } catch (e) {
       Logger.error(`FailureToRequestCallback.exception.${e.message}`)
       Alert.alert(t("errors.something_went_wrong"), e.message)
       setIsLoading(false)
     }
   }
 
+  /* Wait for 15 minutes to pass before submitting another phonenumber after two have been submitted. */
```
artgibson: @lundjrl clean up this doc

src/locales/en.json
```diff
@@ -72,6 +72,7 @@
   },
   "errors": {
     "description": "An error has occurred. Please reload the app.",
+    "maxed_submission_requests": "You have reached the max number of submissions, please wait 15 minutes to resubmit.",
```
artgibson: 15 to 5 minutes

artgibson (changes requested):
@lundjrl a unit test is currently failing.

lundjrl:
@artgibson I added a test case for the Callback form changes.

artgibson:
lgtm

artgibson (approved):

PR #880, opened 2021-03-18
Prevent long navigation titles from colliding with back button

mattThousand:
#### Why:
We'd like long navigation titles to be handled by an ellipsis rather than having them overlap with the back button text on small devices

#### This commit:
This commit caps the title area width at 60% of the total screen width, ensuring that no collisions will occur

#### Screenshots:

Before:
LA 1 11 1 - Overlapping - Do you have your verification code 3

After:
Simulator Screen Shot - iPhone 8 - 2021-03-18 at 09 16 45


artgibson (approved):
lgtm.

PR #886, opened 2021-03-29
Accessibility Enhancements

mattThousand:
### Why:
We'd like the app to be accessible for visually impaired users.

### This commit:
This commit addresses accessibility concerns related to increasing font size for easier readability.


artgibson (approved):
lgtm

PR #889, opened 2021-03-30
Analytics Events Bugfix

mattThousand:
### Why:
We'd like analytics events to be logged as expected for apps making use of escrow verification

### This commit:

This commit fixes a bug which caused `key_submission_consented_to` and `ens_preceding_positive_diagnosis_count` to not be logged


artgibson (approved):
lgtm

PR #892, opened 2021-04-13
Work/gaen 780 dont share keys error handling

lundjrl:
#### Why:

A generic error is passed when user's don't share their keys via escrow. This PR aims to show them what will happen if they don't share their keys.

#### Screenshots:

Screenshot_20210413-073528_GuideSafe

#### How to test:

Submit a phonenumber via Escrow then don't share your keys.

#### Linked issues:

---


artgibson (approved):
lgtm

PR #893, opened 2021-04-20
Prune Exposures > 14 days old from the local device DB

mattThousand:
### Why:
Currently, exposures > 14 days old remain in the local on-device database (encrypted).

### This Commit:
This commit implements a background task that deletes exposures > 14 days old from the local database


artgibson (approved):
lgtm, fix merge conflicts

PR #894, opened 2021-04-23
GAEN-743 Analytic Event For Sending Notification

KyletheWolff:
#### Why:
The requirement here was to have Android call up to the react layer for the `en_notification_received` analytic event to be triggered. 

#### This commit:
The solution was to extract the `ReactContext`from the `getApplicationContext` inside the `StateUpdatedWorker`. This allowed us to then fire off `EventSender.INSTANCE.sendExposureRecordUpdatedChangedEvent(reactContext, exposureResult.getExposures())`

In order to simulate the exact solution without basically mocking it. I updated the DebugMenuModule to send the broadcast. It also send a boolean only for debug mode. If QA still wants to validate with simulation this can still be done however they update the DailySummary. 

#### Screenshots:
N / A

#### How to test:
You can test this through the DebugMenu and simulate it.

---


artgibson (approved):
lgtm

PR #895, opened 2021-04-23
Upgraded react-i18next package locally. Also needed to pass null to a…

lundjrl:
… resolve call to satisfy typescript compiler.

#### Why:

The last version of react-i18next has a vulnerability. This version is way safer.
I also passed "null" to a resolve call since the typescript compiler was arguing about no value being passed at all.

#### Linked issues:

---


artgibson:
lgtm

artgibson (approved):
lgtm

PR #900, opened 2021-05-05
Bitrise Error For Release

KyletheWolff:
#### Why:
Bitrise was failing because of a release compliation error. 

#### This commit:
Add the missing import to the StatusUpdatedWorker in the release package.
Also removes the static import that should not have been there because of a styling issue


artgibson (approved):
lgtm

PR #904, opened 2021-05-07
Fixing Stage Variant Error

KyletheWolff:
#### Why:
To fix bitrise

#### This commit:
We needed to add the Staging package with the worker class.

---


artgibson (approved):
lgtm, bitrise builds for staging and debug were successful

PR #907, opened 2021-05-19
Added await to tracking events to ensure Matomo logs correctly.

lundjrl:
#### Why:

Matomo tracking events (specifically ens_preceding_positive_diagnosis_count) aren't firing off correctly. By adding some async/await logic. We can ensure that the track event has time to fire off.

#### Linked issues:

---


artgibson (approved):
lgtm

PR #915, opened 2021-06-21
Gaen 734

lundjrl:
#### Why:

Text in the header overlaps the back button text. This PR just shrinks the size of the header text in headerMinimalOptions.

#### Screenshots:

IMG_0021
IMG_0020

Taken on an iphone7

#### Linked issues:

---


si1k (approved):
tyvm

PR #932, opened 2021-08-30
Added new .env and text changes for LA.

lundjrl:
#### Why:

LA requires some new text and UI changes to the app to navigate users to their salesforce platform.

#### Linked issues:

---


si1k (approved):
tyvm

PR #934, opened 2021-08-31
Handle Null Exposure Keys

KyletheWolff:
#### Why:
There was a crash occurring on the Android side where the Util class was trying to do a for loop with a null list. 

#### This commit:
Added  logic to handle null list in the ChaffManager and return an empty array if the function returns null. 

#### Linked issues:

---


si1k (approved):
tyvm

PR #935, opened 2021-09-02
GAEN-CheckForExposureError

RJ-Hamilton:
#### Why:

Upon clicking the "Check for Exposures" button on Android, the following message was being displayed to the user...

Screen Shot 2021-09-02 at 2 00 08 AM

#### This commit:

It was discovered that there was a race condition happening when setting up the `ProvideDiagnosisKeyService`. To allow a buffer window for the service to be setup properly, a short delay was added. 

Additionally, an old comment was removed because it was no longer relevant.

#### How to test:

1. Start up application
2. Navigate to "Exposures"
3. Click "Check for Exposures"
    - Expected result: There should be a message that prompts the user by saying "Exposure check complete".


si1k (approved):
tyvm

PR #937, opened 2021-09-16
Android Fixes

KyletheWolff:
#### Why:
There are a few issues regarding Chaff and UAB testing. 

#### This commit:
Reschedule the Chaff worker when the config is changed
Use Observable create for the ProvideDiagnosisKeyService which will handle the task correctly
Remove the Thread sleep function

#### Linked issues:

---


RJ-Hamilton (approved):
LGTM 👍

PR #941, opened 2021-10-29
Add symptoms list to symptoms onset date

mxMarkowitz:
#### Why:
Add a symptoms list translated too the symptoms onset date page.


si1k:
Thank you!

PR #946, opened 2021-11-30
Improvement/updates ruby to 2.6.8

mxMarkowitz:
#### Why:

Updates ruby to more recent version within same main version


TuikBot (approved):
lgty

PR #952, opened 2021-12-08
Update icons external links

mxMarkowitz:
#### Why:

Icons needed to match and be different


si1k (approved):
looks good, thank you

